PR #8, opened 2018-10-12
feature: ZENKO-1224 cleanup bucket

rahulreddy:
This script cleans up a versioned bucket by deleting all versions,
delete markers and abort any ongoing multipart uploads. The script
uses async and batch delete for efficiency.

### DIFF
--- a/cleanupBuckets.js
+++ b/cleanupBuckets.js
@@ -1,0 +1,152 @@
+const http = require('http');
+const async = require('async');
+const AWS = require('aws-sdk');
+const { Logger } = require('werelogs');
+
+const log = new Logger('s3utils::emptyBucket');
+// configurable params
+const BUCKETS = process.argv[2] ? process.argv[2].split(',') : null;
+const ACCESS_KEY = process.env.ACCESS_KEY;
+const SECRET_KEY = process.env.SECRET_KEY;
+const ENDPOINT = process.env.ENDPOINT;
+if (!BUCKETS || BUCKETS.length === 0) {
+    log.fatal('No buckets given as input! Please provide ' +
+        'a comma-separated list of buckets');
+    process.exit(1);
+}
+if (!ENDPOINT) {
+    log.fatal('ENDPOINT not defined!');
+    process.exit(1);
+}
+if (!ACCESS_KEY) {
+    log.fatal('ACCESS_KEY not defined');
+    process.exit(1);
+}
+if (!SECRET_KEY) {
+    log.fatal('SECRET_KEY not defined');
+    process.exit(1);
+}
+const LISTING_LIMIT = 1000;
+
+AWS.config.update({
+    accessKeyId: ACCESS_KEY,
+    secretAccessKey: SECRET_KEY,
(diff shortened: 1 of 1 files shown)


cleanupBuckets.js
```diff
@@ -0,0 +1,160 @@
...
+const async = require('async');
+const AWS = require('aws-sdk');
+const { Logger } = require('werelogs');
+
+const log = new Logger('s3utils::emptyBucket');
+// configurable params
+const BUCKETS = process.argv[2] ? process.argv[2].split(',') : null;
+const ACCESS_KEY = process.env.ACCESS_KEY;
+const SECRET_KEY = process.env.SECRET_KEY;
+const ENDPOINT = process.env.endpoint;
```
philipyoo: capitalized `endpoint`? following what is used in `stalled.js`

cleanupBuckets.js
```diff
@@ -0,0 +1,160 @@
...
+}
+if (!ACCESS_KEY) {
+    log.fatal('ACCESS_KEY not defined');
+    process.exit(1);
+}
+if (!SECRET_KEY) {
+    log.fatal('SECRET_KEY not defined');
+    process.exit(1);
+}
+const LISTING_LIMIT = 300;
```
philipyoo: why not just rely on the default 1000?

cleanupBuckets.js
```diff
@@ -0,0 +1,160 @@
...
+            if (err) {
+                log.error('batch delete err', err);
+                return done(err);
+            }
+            Objects.forEach(v => log.info(`deleted key: ${v.Key}`));
+            return done();
+        });
+    }
+    const fns = [];
+    while (Object.keys(objectsToDelete).length > 0) {
```
philipyoo: I'm not sure if this is needed because I don't see `objectsToDelete` ever being greater than 1000

cleanupBuckets.js
rahulreddy: Yeah it can be 1000, originally I put it lower because when there are a billion objects to delete it can overwhelm the cluster.

cleanupBuckets.js
rahulreddy: good catch

cleanupBuckets.js
rahulreddy: Correct. It's no longer the case that listing returns results more than 1000.

rahulreddy:
@bert-e reset

cleanupBuckets.js
```diff
@@ -0,0 +1,155 @@
...
+    return keys.map(v => ({
+        Key: v.Key,
+        VersionId: v.VersionId,
+    }));
+}
+
+// delete all versions of an object
+function _deleteVersions(bucket, objectsToDelete, cb) {
+    // multi object delete can delete max 1000 objects
+    function _batchDelete(Objects, done) {
```
philipyoo: just cosmetic so no biggie, but no longer need this inner function

philipyoo (approved):

rahulreddy:
@bert-e reset

PR #18, opened 2018-11-08
feature: S3C-1607 script to trigger crr on existing objects

rahulreddy:
feature: S3C-1607 script to trigger crr on existing objects

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,15 @@
 # s3utils
 S3 Connector and Zenko Utilities
+
+Run the docker container as
+```
+docker run -d -e 'ACCESS_KEY=accessKey' -e 'SECRET_KEY=secretkey' -e 'ENDPOINT=http://127.0.0.1:8000'
+```
+
+## Trigger CRR on objects that were put before replication was enabled on the bucket
+
+1. Enable versioning and setup replication on the bucket
+2. Run script as
+```
+node crrExistingObjects.js testbucket1, testbucket2
+```
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ S3 Connector and Zenko Utilities
 
 Run the docker container as
 ```
-docker run -d -e 'ACCESS_KEY=accessKey' -e 'SECRET_KEY=secretkey' -e 'ENDPOINT=http://127.0.0.1:8000'
+docker run -d -e 'ACCESS_KEY=accessKey' -e 'SECRET_KEY=secretkey' -e 'ENDPOINT=http://127.0.0.1:8000' zenko/s3utils
 ```
 
 ## Trigger CRR on objects that were put before replication was enabled on the bucket
@@ -11,5 +11,15 @@ docker run -d -e 'ACCESS_KEY=accessKey' -e 'SECRET_KEY=secretkey' -e 'ENDPOINT=h
 1. Enable versioning and setup replication on the bucket
 2. Run script as
 ```
-node crrExistingObjects.js testbucket1, testbucket2
+node crrExistingObjects.js testbucket1,testbucket2
+```
+
+# Empty a versioned bucket
+
+This script deletes all versions of objects in the bucket including delete markers,
+and aborts any ongoing multipart uploads to prepare the bucket for deletion.
+
+**Note: This will delete data associated with the objects and it's not recoverable**
+```
+node cleanupBuckets.js testbucket1,testbucket2
 ```
--- a/crrExistingObjects.js
+++ b/crrExistingObjects.js
@@ -1,0 +1,166 @@
+const http = require('http');
+
+const AWS = require('aws-sdk');
+const { doWhilst, mapLimit, waterfall } = require('async');
+
+const { Logger } = require('werelogs');
+
+const BackbeatClient = require('./BackbeatClient');
+
+const log = new Logger('s3utils::emptyBucket');
+const BUCKETS = process.argv[2] ? process.argv[2].split(',') : null;
+const ACCESS_KEY = process.env.ACCESS_KEY;
+const SECRET_KEY = process.env.SECRET_KEY;
+const ENDPOINT = process.env.ENDPOINT;
+const LISTING_LIMIT = 1000;
+
+if (!BUCKETS || BUCKETS.length === 0) {
+    log.fatal('No buckets given as input! Please provide ' +
+        'a comma-separated list of buckets');
+    process.exit(1);
+}
+if (!ENDPOINT) {
+    log.fatal('ENDPOINT not defined!');
+    process.exit(1);
+}
+if (!ACCESS_KEY) {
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -1,2 +1,15 @@
...
+```
+docker run -d -e 'ACCESS_KEY=accessKey' -e 'SECRET_KEY=secretkey' -e 'ENDPOINT=http://127.0.0.1:8000'
+```
+
+## Trigger CRR on objects that were put before replication was enabled on the bucket
+
+1. Enable versioning and setup replication on the bucket
+2. Run script as
+```
+node crrExistingObjects.js testbucket1, testbucket2
```
jonathan-gramain: The current version only works on one bucket right?

crrExistingObjects.js
```diff
@@ -0,0 +1,123 @@
...
+                Body: mdBlob,
+            }, next);
+        },
+    ], cb);
+}
+
+
+// Here's the logic that triggers everything
+waterfall([
+    next => s3.listObjectVersions({ Bucket: BUCKET }, next),
```
jonathan-gramain: Don't we need a loop of some sort to iterate when the listing is truncated?

crrExistingObjects.js
```diff
@@ -0,0 +1,123 @@
...
+        }
+        return next(null, data, res.ReplicationConfiguration);
+    }),
+    (data, repConfig, next) => mapLimit(data.Versions, 10, (i, apply) => {
+        const { Key, VersionId } = i;
+        _markObjectPending(BUCKET, Key, VersionId, repConfig, apply);
+    }, next),
+], err => {
+    if (err) {
+        return log.error('error occured', err);
```
jonathan-gramain: Could be good to know which object(s) caused errors, hence logging in the inner code instead could be better

crrExistingObjects.js
rahulreddy: Oh yeah I forgot to swap out the method with the recursive listing one.

README.md
rahulreddy: Right, I will update that.

rahulreddy:
@bert-e reset

rahulreddy:
@bert-e reset

rahulreddy:
@bert-e reset

rahulreddy:
@bert-e reset

rahulreddy:
@bert-e reset

crrExistingObjects.js
```diff
@@ -0,0 +1,166 @@
...
+
+const AWS = require('aws-sdk');
+const { doWhilst, mapLimit, waterfall } = require('async');
+
+const { Logger } = require('werelogs');
+
+const BackbeatClient = require('./BackbeatClient');
+
+const log = new Logger('s3utils::emptyBucket');
+const BUCKETS = process.argv[2] ? process.argv[2].split(',') : null;
```
jonathan-gramain, line 11: I think it's fine for now, but I'm thinking the script will likely evolve a little bit, e.g. to support DR and reprocess REPLICA entries, and then it could be worth switching to a proper argument parsing module like "commander" for the extra options, actually even now it could be used e.g. to specify a different number of workers, as the default may not be a good fit for everyone. Using an argument-parsing module would also have the benefit of providing consistent help information for usage. Another way could be to pass everything as environment variables, but I don't think this would be the most usable approach.

Also it looks to me more natural to allow passing the buckets as a list of arguments instead of custom `split(',')` call, because most command-line tools already use this scheme to parse extra args (so to comply with principle of least surprise).

rahulreddy:
@bert-e help

rahulreddy:
@bert-e clear

crrExistingObjects.js
```diff
@@ -0,0 +1,166 @@
...
+    const options = { Bucket: bucket };
+    waterfall([
+        next => s3.getBucketReplication(options, (err, res) => {
+            if (err) {
+                log.error('error getting bucket replication', { error: err });
+                return next(err);
+            }
+            return next(null, res.ReplicationConfiguration);
+        }),
+        (repConfig, next) => mapLimit(versions, 10, (i, apply) => {
```
jonathan-gramain, line 122: 10 may not be a good fit for everyone, so should be made configurable or a high-level exposed constant (also see my earlier comment on argument parsing).

crrExistingObjects.js
```diff
@@ -0,0 +1,166 @@
...
+    signatureVersion: 'v4',
+    signatureCache: false,
+    httpOptions: {
+        maxRetries: 0,
+        timeout: 0,
+        agent: new http.Agent({ keepAlive: true }),
+    },
+};
+const s3 = new AWS.S3(options);
+
```
jonathan-gramain, line 51: Could remove the spacing here

crrExistingObjects.js
```diff
@@ -0,0 +1,166 @@
...
+                log.error('error marking objects for crr', { bucket });
+                return cb(err);
+            }
+            log.info(`completed task for bucket: ${bucket}`);
+            return cb();
+        });
+}
+
+// trigger the calls to list objects and mark them for crr
+mapLimit(BUCKETS, 1, triggerCRR, err => {
```
jonathan-gramain, line 161: Since you don't use the result, you may use `eachSeries`

crrExistingObjects.js
```diff
@@ -0,0 +1,166 @@
...
+                return true;
+            }
+            return false;
+        },
+        err => {
+            if (err) {
+                log.error('error marking objects for crr', { bucket });
+                return cb(err);
+            }
+            log.info(`completed task for bucket: ${bucket}`);
```
jonathan-gramain, line 155: minor: this log could be made more consistent by logging the bucket name as an extra param `{ bucket }`

crrExistingObjects.js
```diff
@@ -0,0 +1,166 @@
...
+
+function triggerCRR(bucketName, cb) {
+    const bucket = bucketName.trim();
+    let VersionIdMarker = null;
+    let KeyMarker = null;
+    doWhilst(
+        done => _listObjectVersions(bucket, VersionIdMarker, KeyMarker,
+            (err, data) => {
+                if (err) {
+                    log.error('error listing object versions', { error: err });
```
jonathan-gramain, line 137: The bucket name could be logged

crrExistingObjects.js
```diff
@@ -0,0 +1,166 @@
...
+        KeyMarker,
+    }, cb);
+}
+
+function _markPending(bucket, versions, cb) {
+    const options = { Bucket: bucket };
+    waterfall([
+        next => s3.getBucketReplication(options, (err, res) => {
+            if (err) {
+                log.error('error getting bucket replication', { error: err });
```
jonathan-gramain, line 117: The bucket name could be logged

jonathan-gramain (approved):

PR #51, opened 2018-11-30
S3C-1850 fixes for CRR on existing objects

jonathan-gramain:
* S3C-1850 require SITE_NAME environment variable

When StorageClass is not defined in the replication configuration,
which is the case on S3C, require it to be passed via the SITE_NAME
environment variable, so that it can be set correctly in object
metadata.

* S3C-1850 trigger CRR on nonversioned keys

Add an extra metadata update for keys that were originally put on a
non-versioned bucket (see code comments for technical details on why
this is needed).

Note: in w/8.0 and w/8.1, I updated the merge to remove the update on README.md that shows how to pass the SITE_NAME environment variable: for zenko, it should not be needed because we always should have a storage class.

Tested with:

- objects put originally on nonversioned bucket: script adds an internal version and triggers replication
- objects put originally on versioning-suspended bucket: script triggers replication
- objects put originally on versioned bucket: script triggers replication on all versions
- objects already replicated: script skips those entries

### DIFF
--- a/crrExistingObjects.js
+++ b/crrExistingObjects.js
@@ -7,11 +7,12 @@ const { Logger } = require('werelogs');
 
 const BackbeatClient = require('./BackbeatClient');
 
-const log = new Logger('s3utils::emptyBucket');
+const log = new Logger('s3utils::crrExistingObjects');
 const BUCKETS = process.argv[2] ? process.argv[2].split(',') : null;
 const ACCESS_KEY = process.env.ACCESS_KEY;
 const SECRET_KEY = process.env.SECRET_KEY;
 const ENDPOINT = process.env.ENDPOINT;
+const SITE_NAME = process.env.SITE_NAME;
 const LISTING_LIMIT = 1000;
 
 if (!BUCKETS || BUCKETS.length === 0) {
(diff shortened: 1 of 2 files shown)


crrExistingObjects.js
```diff
@@ -62,14 +63,50 @@ function _markObjectPending(bucket, key, versionId, repConfig, cb) {
         }, next),
         // update replication info and put back object blob
         (mdRes, next) => {
-            const objMD = JSON.parse(mdRes.Body);
+            objMD = JSON.parse(mdRes.Body);
+            if (objMD.replicationInfo && objMD.replicationInfo.status !== '') {
```
bennettbuchanan, line 67: Maybe it's overkill to create a `ObjectMD` instance to check the values (e.g. using the methods `getReplicationInfo` and `getReplicationStatus`), but it would have the advantage of referring to a source of truth for the metadata schema. However, it could require extra work in Arsenal to handle backward compatibility with older object metadata versions. Just thinking out loud here. 😸

bennettbuchanan (approved):

rahulreddy (approved):

rahulreddy:
@jonathan-gramain New image can be generated using Dockerfile in this repo. Make sure you push twice
- once for the tag push 0.x
- once for tagging 0.x as latest

crrExistingObjects.js
jonathan-gramain: Actually, sounds like a good idea, but when I tried using an arsenal dependency to use some versioning helper I ran into some weird compatibility error between the node.js version used by "diskusage" module and this s3utils script, where I assume "diskusage" is pulled by arsenal in a dependency chain, so I will keep as is. It would be good to understand if some time we have to pull this dependency for another reason.

jonathan-gramain:
@bert-e approve

PR #57, opened 2018-12-13
ft: ZENKO-1379 update for disaster recovery use case

jonathan-gramain:
* allow processing all objects for CRR on the set of provided buckets
  (not just the ones without a replication configuration), by setting
  the PROCESS_ALL environment variable to a truish value

* provide ability to specify the number of workers with the WORKERS
  environment variable (default is 10, may not be enough for very
  large number of objects)

* add "skipped" count to periodic progress log messages

* handle SIGINT/SIGHUP/SIGTERM/SIGQUIT to terminate the script

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,10 @@ Run the docker container as
 docker run --net=host -e 'ACCESS_KEY=accessKey' -e 'SECRET_KEY=secretkey' -e 'ENDPOINT=http://127.0.0.1:8000' -e 'SITE_NAME=crrSiteName' zenko/s3utils node scriptName bucket1[,bucket2...]
 ```
 
+Optionally, the environment variable "WORKERS" may be set to specify
+how many parallel workers should run, otherwise a default of 10
+workers will be used.
+
 ## Trigger CRR on objects that were put before replication was enabled on the bucket
 
 1. Enable versioning and setup replication on the bucket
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,6 +18,18 @@ docker run --net=host -e 'ACCESS_KEY=accessKey' -e 'SECRET_KEY=secretkey' -e 'EN
 node crrExistingObjects.js testbucket1,testbucket2
 ```
 
+## Trigger CRR on *all* objects of a bucket, including those that have
```
rahulreddy: Not sure if it's intentional but the heading appears to be either cut off or included in header by accident.
`Trigger CRR on *all* objects of a bucket` makes sense as header.

crrExistingObjects.js
```diff
@@ -13,6 +13,10 @@ const ACCESS_KEY = process.env.ACCESS_KEY;
 const SECRET_KEY = process.env.SECRET_KEY;
 const ENDPOINT = process.env.ENDPOINT;
 const SITE_NAME = process.env.SITE_NAME;
+const PROCESS_ALL = ['1', 'true', 'yes', 'on'].includes(
```
rahulreddy: I think restricting the input is important, otherwise each person can have a different value used in their scripts. It will help validation as well. For example, if `-e "PROCESS_ALL=true"`, all you have to check for is
`const PROCESS_ALL = process.env.PROCESS_ALL === 'true;` You can throw an assertion if the user enters anything else.

rahulreddy (changes requested):

rahulreddy (approved):

jonathan-gramain:
@bert-e reset

jonathan-gramain:
@bert-e approve

PR #64, opened 2019-03-20
ft: Add mongo to dockerfile

anurag4DSB:
**What does this PR do, and why do we need it?**
Add a step in dockerfile to install MongoDB

**Which issue does this PR fix?**
Zenko-1594

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ COPY ./package.json .
 
 RUN apt-get update \
     && apt-get install -y jq python git build-essential vim --no-install-recommends \
+    && apt-get install mongodb \
     && npm install
 
 COPY ./ ./


JianqinWang (approved):

tmacro (approved):

rahulreddy (changes requested):
Your branch naming should be `feature/` for bert-e to pick it up.

ssalaues (changes requested):
Don't need all of mongodb and would be good to pin the version being installed. I made a PR here that does this
https://github.com/scality/s3utils/pull/63

anurag4DSB:
Post discussion with @ssalaues it will be added to pull request #63 
We just need mongo-org-tools

PR #65, opened 2019-03-22
ft: S3C-1892 transfer bucket accounts

dora-korpar:
A script to migrate buckets and their contents from one user account to another.

### DIFF
--- a/migrateBuckets/migrateBuckets.js
+++ b/migrateBuckets/migrateBuckets.js
@@ -1,0 +1,355 @@
+#!/usr/bin/env node
+'use strict'; // eslint-disable-line strict
+
+const { models, versioning } = require('arsenal');
+const async = require('async');
+const AWS = require('aws-sdk');
+const bucketclient = require('bucketclient');
+const commander = require('commander');
+const http = require('http');
+const werelogs = require('werelogs');
+
+werelogs.configure({ level: 'info', dump: 'error' });
+const logger = new werelogs.Logger('s3utils:migrateBuckets')
+    .newRequestLogger();
+const { BucketInfo } = models;
+const versionIdUtils = versioning.VersionID;
+
+// configurable params
+const ACCESS_KEY = process.env.ACCESS_KEY;
+const SECRET_KEY = process.env.SECRET_KEY;
+
+commander.version('1.0.0')
+.option('-b, --bucket-name <bucketName>', 'Name of the bucket')
+.option('-s, --src-canonical-id <srcCanonicalId>',
+    'Canonical id of source account')
+.option('-d, --dest-canonical-id <destCanonicalId>',
+    'Canonical id of destination account')
+.option('-o, --dest-owner-name <destOwnerName>',
+    'Owner display name of destination account')
+.option('-e, --endpoint <endpoint>', 'S3Connector endpoint')
+.option('-u, --bucketd <bucketd>', 'Bootstrap list of bucketd endpoints')
+.option('--https-key [httpsKey]', 'HTTPS key')
+.option('--https-cert [httpsCert]', 'HTTPS cert')
+.option('--https-ca [httpsCa]', 'HTTPS ca')
+.option('-p, --profile [profile]', 'AWS credentials')
+.parse(process.argv);
+
+const {
+    bucketName,
+    srcCanonicalId,
+    destCanonicalId,
+    destOwnerName,
+    endpoint,
+    bucketd,
+    httpsKey,
+    httpsCert,
+    httpsCa,
+    profile,
+} = commander;
+
+if ((!ACCESS_KEY || !SECRET_KEY) && !profile) {
+    logger.error('Missing credentials. Set ACCESS_KEY & SECRET_KEY env ' +
+        'variables or pass --profile option');
+    process.exit(1);
+}
+
+if (!bucketName || !srcCanonicalId ||
+!destCanonicalId || !destOwnerName || !bucketd) {
+    logger.error('Missing command line parameter');
+    commander.outputHelp();
+    process.exit(1);
+}
+
+let keyInProgress;
+let versionIdInProgress;
+let nErrors = 0;
+let nObjProcessed = 0;
+
+const s3Opts = {
+    region: 'us-east-1',
+    sslEnabled: false,
+    endpoint,
+    s3ForcePathStyle: true,
+    apiVersions: { s3: '2006-03-01' },
+    signatureVersion: 'v4',
+    signatureCache: false,
+    httpOptions: {
+        maxRetries: 10,
+        timeout: 0,
+        agent: new http.Agent({ keepAlive: true }),
+    },
+};
+if (ACCESS_KEY && SECRET_KEY) {
+    s3Opts.accessKeyId = ACCESS_KEY;
+    s3Opts.secretAccessKey = SECRET_KEY;
+}
+if (profile) {
+    const credentials = new AWS.SharedIniFileCredentials({ profile });
+    AWS.config.credentials = credentials;
+}
+
+const s3 = new AWS.S3(s3Opts);
+
+let mdclient;
+try {
+    if (httpsKey && httpsCert) {
+        mdclient = new bucketclient.RESTClient(
+            bucketd, undefined, true, httpsKey, httpsCert, httpsCa);
+    } else {
+        mdclient = new bucketclient.RESTClient(bucketd, undefined, false);
+    }
+} catch (err) {
+    logger.error(`Could not instantiate bucketclient: ${err.message}`);
+    process.exit(1);
+}
+
+const splitter = '..|..';
+
+function _createUsersBucketKey(bucketName, id) {
+    return `${id}${splitter}${bucketName}`;
+}
+
+function _changeBucketAccount(bucket, log, cb) {
+    const {
+        bucketName,
+        destCanonicalId,
+        destOwnerName,
+    } = bucket;
+
+    mdclient.getBucketAttributes(bucketName, log.getSerializedUids(),
+    (err, data) => {
+        if (err) {
+            ++nErrors;
+            log.error(`Error getting bucket attributes: ${err}`);
+            return cb(err);
+        }
+        const bucketMD = BucketInfo.fromObj(BucketInfo.deSerialize(data));
+        const versioningEnabled = bucketMD.isVersioningEnabled();
+        const bucketId = bucketMD.getOwner();
+        const bucketOwner = bucketMD.getOwnerDisplayName();
+        if (bucketId === destCanonicalId && bucketOwner === destOwnerName) {
+            log.info(`Bucket ${bucketName} id already updated`);
+            return cb(null, versioningEnabled);
+        }
+        bucketMD.setOwner(destCanonicalId);
+        bucketMD.setOwnerDisplayName(destOwnerName);
+        return mdclient.putBucketAttributes(bucketName, log.getSerializedUids(),
+        bucketMD.serialize(), err => {
+            if (err) {
+                ++nErrors;
+                log.error(`Error updating bucket ${bucketName} ` +
+                    `metadata: ${err}`);
+                return cb(err);
+            }
+            log.info(`Bucket ${bucketName} id successfully updated`);
+            return cb(null, versioningEnabled);
+        });
+    });
+}
+
+function _changeUsersBucket(bucket, log, cb) {
+    const metaBucket = 'users..bucket';
+    const serUids = log.getSerializedUids();
+    const { bucketName, srcCanonicalId, destCanonicalId } = bucket;
+
+    return mdclient.listObject(metaBucket, serUids, {}, (err, data) => {
+        if (err) {
+            ++nErrors;
+            log.error(`Error listing metadata bucket: ${err}`);
+            return cb(err);
+        }
+        const objectList = JSON.parse(data);
+
+        const searchKey = _createUsersBucketKey(bucketName, srcCanonicalId);
+        const matchingBucket = objectList.Contents.find(object =>
+            object.key === searchKey);
+
+        if (!matchingBucket) {
+            log.info(`users..bucket object ${bucketName} does not need ` +
+            'to be updated');
+            return cb();
+        }
+        return async.waterfall([
+            next => mdclient.getObject(metaBucket, matchingBucket.key, serUids,
+            (err, objectMD) => {
+                if (err) {
+                    log.error('Error getting users..bucket object ' +
+                        `metadata: ${err}`);
+                    return next(err);
+                }
+                const newObjectName = _createUsersBucketKey(
+                    bucketName, destCanonicalId);
+                return next(null, newObjectName, objectMD);
+            }),
+            (newObjectName, objectMD, next) => mdclient.putObject(
+            metaBucket, newObjectName, objectMD, serUids, err => {
+                if (err) {
+                    log.error('Error updating users..bucket object ' +
+                        `${matchingBucket.key}: ${err}`);
+                    return next(err);
+                }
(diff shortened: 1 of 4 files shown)


rahulreddy (dismissed):
LGTM

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+* }
+*/
+
+// configurable params
+const ACCESS_KEY = process.env.ACCESS_KEY;
+const SECRET_KEY = process.env.SECRET_KEY;
+const ENDPOINT = process.env.ENDPOINT;
+
+commander
+.version('0.0.1')
```
jonathan-gramain: Let's be bold and call this `1.0.0` :wink:

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+const fs = require('fs');
+const http = require('http');
+const werelogs = require('werelogs');
+
+werelogs.configure({ level: 'info', dump: 'error' });
+const logger = new werelogs.Logger('s3utils:migrateBuckets')
+    .newRequestLogger();
+const { BucketInfo } = models;
+const versionIdUtils = versioning.VersionID;
+/* expected input JSON structure:
```
jonathan-gramain: I believe this has to do with the config file, hence it looks it should be above the declaration of the `--config-file` option.

I think it makes things hard to configure to ask the user to provide all of: environment variables (access key etc.), command-line options (bucket, canonical ID etc.) and a config file. I understand there can be a security concern asking to provide secret keys on the command line, but otherwise everything should be consistently passed via one or max. two means. Here are my suggestions:

- the bucketd bootstrap and https options could be passed on the command-line (`--bucketd-endpoint` with a comma-separated list and `--https-key` etc. for example)
- ENDPOINT could also be passed via command line
- it's okay to keep ACCESS_KEY and SECRET_KEY as environment variables, although another way would be to pass them with `--profile` option referring to a specific AWS client profile (there is off-the-shelf support for profiles in AWS SDK)

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+
+const {
+    bucketName,
+    srcCanonicalId,
+    destCanonicalId,
+    destOwnerName,
+    configPath,
+} = commander;
+
+if (!bucketName || !srcCanonicalId ||
```
jonathan-gramain, line 59: I believe appending `.required()` to each of those related `.option()` would have the same effect.

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+try {
+    const data = fs.readFileSync(configPath, { encoding: 'utf-8' });
+    config = JSON.parse(data);
+} catch (err) {
+    logger.error(`Could not parse provided config file: ${err.message}`);
+    process.exit(1);
+}
+
+const s3 = new AWS.S3({
+    accessKeyId: ACCESS_KEY,
```
jonathan-gramain: Maybe we should also make sure that `ACCESS_KEY`, `SECRET_KEY` and `ENDPOINT` are provided before creating the `AWS.S3` object.

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+    });
+}
+
+function _changeUsersBucket(bucket, log, cb) {
+    const metaBucket = 'users..bucket';
+    const serUids = log.getSerializedUids();
+    const { bucketName, srcCanonicalId, destCanonicalId } = bucket;
+
+    return mdclient.listObject(metaBucket, serUids, {}, (err, data) => {
+        const objectList = JSON.parse(data);
```
jonathan-gramain, line 164: I believe we should check the `err` object before going on

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+
+        const searchKey = _createUsersBucketKey(bucketName, srcCanonicalId);
+        const matchingBucket = objectList.Contents.find(object =>
+            object.key === searchKey);
+
+        if (!matchingBucket) {
+            log.debug(`users..bucket object ${bucketName} does not need ` +
```
jonathan-gramain: `log.info` could be better

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+    } = bucket;
+    const { Key, VersionId } = object;
+    const decodedId = VersionId ? versionIdUtils.decode(VersionId) : undefined;
+
+    // if versiondId is undefined, retrieval will return most current object md
+    const getOptions = { versionId: decodedId };
+
+    mdclient.getObject(bucketName, Key, log.getSerializedUids(),
+    (err, data) => {
+        const objectMD = JSON.parse(data);
```
jonathan-gramain: This should be done after the `if (err)` section

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+        const objectOwner = objectMD['owner-display-name'];
+        if (objectId === destCanonicalId && objectOwner === destOwnerName) {
+            log.debug(`Object ${Key} id already updated`);
+            return cb();
+        }
+        objectMD['owner-id'] = destCanonicalId;
+        objectMD['owner-display-name'] = destOwnerName;
+        const putOptions = {};
+        if (VersionId) {
+            putOptions.versionId = VersionId;
```
jonathan-gramain: isn't it the decoded version that we need here? Not 100% sure but it would make more sense to me API-wise since `getObject` takes the decoded version. If you already checked and are sure it's correct, just ignore this.

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+        }
+        return mdclient.putObject(bucketName, Key, JSON.stringify(objectMD),
+        log.getSerializedUids(), err => {
+            if (err) {
+                log.error(`Error updating object ${Key} metadata: ${err}`);
+                return cb(err);
+            }
+            log.debug(`Object ${Key} id successfully updated`);
```
jonathan-gramain: I'd suggest to make it a `log.info` call: it can be a lot of log, depending on the bucket size, but it's better than no feedback at all until the end of the migration.

Also you can add the `VersionId` field to the log line along with the `Key`

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+            return cb();
+        }, putOptions);
+    }, getOptions);
+}
+
+function _changeObjectsAccount(bucket, log, cb) {
+    const bucketName = bucket.bucketName;
+    const versioningEnabled = bucket.versioningEnabled;
+    const listMethod = versioningEnabled ? 'listObjectVersions' : 'listObjects';
```
jonathan-gramain: Normally you may use `listObjectVersions` on a non-versioned bucket as well, the exact results returned may be closer to a versioned bucket result than with using `listObjects` (e.g. returning a `Versions` array)

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+    const dataKey = versioningEnabled ? 'Versions' : 'Contents';
+    let moreObjects = true;
+    let objectsArr = [];
+    let marker = null;
+    async.whilst(
+        () => moreObjects === true,
```
jonathan-gramain: `() => moreObjects` would work

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+            if (marker) {
+                listParams.Marker = marker;
+            }
+            s3[listMethod](listParams, (err, data) => {
+                if (err) {
+                    return done(err);
+                }
+                objectsArr = objectsArr.concat(data[dataKey]);
+                if (data.IsTruncated) {
+                    marker = data.Contents[data[dataKey].length - 1].Key;
```
jonathan-gramain: don't you need to use `data[dataKey]` instead of `data.Contents`? But see my above comment, with `listObjectVersions` it may not be necessary to distinguish treatment between non-versioned and versioned buckets anymore.

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+        done => {
+            const listParams = { Bucket: bucketName };
+            if (marker) {
+                listParams.Marker = marker;
+            }
+            s3[listMethod](listParams, (err, data) => {
+                if (err) {
+                    return done(err);
+                }
+                objectsArr = objectsArr.concat(data[dataKey]);
```
jonathan-gramain: I fear this is going to be both very inefficient and explode memory usage (and crash the process eventually) if the bucket is a little big (10M entries for example): inefficient because we are calling `concat` repeatedly on each entry, re-creating a brand new array each time, and use memory because we are keeping all bucket entries in memory.

I believe we can instead call `_changeOneObjectAccount` after each listing iteration, on each listed entry, and not have to store them in an array.

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+}
+
+migrateBuckets(commander, err => {
+    if (err) {
+        logger.error(`Bucket migration failed: ${err}\n` +
+            'Please run script again');
+    } else {
+        logger.info('Completed bucket migration');
+    }
+});
```
jonathan-gramain, line 338: It could be good to add signal handlers, in case an admin stops the script with `SIGINT/SIGTERM` (see other scripts in s3utils for examples).

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,327 @@
...
+    secretAccessKey: SECRET_KEY,
+    region: 'us-east-1',
+    sslEnabled: false,
+    endpoint: ENDPOINT,
+    s3ForcePathStyle: true,
+    apiVersions: { s3: '2006-03-01' },
+    signatureVersion: 'v4',
+    signatureCache: false,
+    httpOptions: {
+        maxRetries: 0,
```
jonathan-gramain: Having no retries here mean that the SDK will give up at the first error. This could be an issue if listing large enough buckets, where a single listing error will halt all processing.

We could use a large enough number of retries so that we reduce this risk, and also, making sure that the script is able to recover when restarted if it stopped in the middle of a migration (I think it's the case already, just double checking).

jonathan-gramain (changes requested):

migrateBuckets/migrateBuckets.js
dora-korpar: do you think it would be better to have everything passed through the command line or through a new config that contains all those options?

jonathan-gramain: I tend to prefer the flexibility of the command line, especially if using tooling like commander, but a config file would work too, up to you.

migrateBuckets/migrateBuckets.js
dora-korpar: yeah, if the script fails at any point, it can be rerun without any problems. but i'll add in a few retries. does 3 seem reasonable?

migrateBuckets/migrateBuckets.js
dora-korpar: It looks like with the version of commander we use, surrounding the var name with <> means it's required, [] means it's optional, but it doesn't look like it throws any errors natively. https://www.npmjs.com/package/commander/v/2.20.3

migrateBuckets/migrateBuckets.js
jonathan-gramain: I'm not sure what is the best number, I'd say 10 would be good enough, it needs to be long enough to avoid failures due to transient issues, but short enough that it stops timely if there's a serious issue. AWS SDK uses exponential backoff between retries AFAIK which is desirable.

migrateBuckets/migrateBuckets.js
```diff
@@ -15,89 +14,90 @@ const logger = new werelogs.Logger('s3utils:migrateBuckets')
...
     'Canonical id of source account')
-.option('-d, --dest-canonical-id <destCanonicalId',
+.option('-d, --dest-canonical-id <destCanonicalId>',
     'Canonical id of destination account')
 .option('-o, --dest-owner-name <destOwnerName>',
     'Owner display name of destination account')
-.option('-c, --config-path <configPath>', 'Path to metadata config')
+.option('-e, --endpoint <endpoint>', 'S3Connector endpoint')
+.option('-u, --bucketd <bucketd>', 'Bootstrap list of bucketd endpoints')
+.option('-k, --https-key [httpsKey]', 'HTTPS key')
```
jonathan-gramain: minor point: for the http-xxx options I would not have provided a short-option form (single letter) but that's ok if it's there.

migrateBuckets/migrateBuckets.js
```diff
@@ -15,89 +14,90 @@ const logger = new werelogs.Logger('s3utils:migrateBuckets')
...
+    AWS.config.credentials = credentials;
+}
+
+const s3 = new AWS.S3(s3Opts);
+
 let mdclient;
 try {
-    if (config.https) {
-        const { key, cert, ca } = config.https;
+    if (httpsKey && httpsCert && httpsCa) {
```
jonathan-gramain: I think `httpsCa` may not be always provided even if `httpsKey` and `httpsCert` are (they would use the system-installed CA certificates then). I doubt I'd rather check that only the first two are defined to activate HTTPS.

migrateBuckets/migrateBuckets.js
```diff
@@ -213,88 +222,90 @@ function _changeOneObjectAccount(bucket, object, log, cb) {
...
             }
-            s3[listMethod](listParams, (err, data) => {
+            // if versioning is not enabled on bucket, VersionId will be null
+            s3.listObjectVersions(listParams, (err, data) => {
                 if (err) {
+                    ++nErrors;
                     return done(err);
                 }
-                objectsArr = objectsArr.concat(data[dataKey]);
+                objectsArr = objectsArr.concat(data.Versions);
```
jonathan-gramain: Maybe you forgot to remove `objectsArr` after the change to process versions during listing?

jonathan-gramain (approved):

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,355 @@
...
+    process.exit(1);
+}
+
+let keyInProgress;
+let versionIdInProgress;
+let nErrors = 0;
+let nObjProcessed = 0;
+
+const s3Opts = {
+    region: 'us-east-1',
```
alexandre-merle: should we make this configurable or no need ?

migrateBuckets/migrateBuckets.js
```diff
@@ -0,0 +1,355 @@
...
+                } else {
+                    moreObjects = false;
+                }
+                return async.each(data.Versions, (object, callback) =>
+                    _changeOneObjectAccount(bucket, object, log, callback),
+                    err => {
+                        if (err) {
+                            return done(err);
+                        }
+                        log.info(`All objects in bucket ${bucketName} ` +
```
alexandre-merle: this log should be after async.whilst :)

alexandre-merle (commented):
Not much to say :) one question, should we lock the bucket during the operation to avoid any weird behavior ? (if we have any way to do this)

migrateBuckets/migrateBuckets.js
dora-korpar: made it configurable just in case

dora-korpar:
@alexandre-merle I don't really think we have a way of locking the bucket. If there's an issue with some objects being left behind, the user should just be able to rerun the script, and it will transfer any new objects or objects it missed.

PR #85, opened 2020-01-14
improvement: ZENKO-2278 count items

alexanderchan-scality:
rewrite count items to perform client-side processing in order to avoid server-side aggregate that may take up mongodb resources.


rachedbenmustapha (commented):
Moving the file seems to break the chart and the operator which create jobs that run `node countItems.js`.

JianqinWang (approved):

dora-korpar (approved):

countItems.js
```diff
@@ -1,72 +1,165 @@
...
+        maxConcurrentCallsPerWorker: concurrentCursors,
+        maxRetries: 0,
+        autoStart: false,
+    },
+    require.resolve('./countItems/counter'),
+    ['clientSetup', 'countItems', 'clientTeardown']
+);
+
+function setup(callback) {
+    async.times(numWorkers, workers.clientSetup,
```
rachedbenmustapha: Is this guaranteed to work somehow, or is it just by coincidence that round-robin works perfectly and each worker gets initialized (same with `cleanup`)?

countItems.js
```diff
@@ -1,72 +1,165 @@
...
+    async.waterfall([
+        next => setup(next),
+        next => work(bucketList, next),
+    ], (err, store) => {
+        if (err) {
+            log.error('error occurred in worker process', {
+                error: err,
             });
-            return process.exit(1);
+            return cleanup(null, callback);
```
rachedbenmustapha: The error seems to be lost here, won't this code path end up trying to call save with `null` results?

countItems.js
```diff
@@ -1,72 +1,165 @@
...
-    database,
-    isLocationTransient: getIsLocationTransientCb(),
-    writeConcern: 'majority',
-    replicaSet: 'rs0',
-    readPreference: 'secondaryPreferred',
-    replicationGroupId: 'RG001',
-    logger: log,
-};
+const workers = workerFarm(
+    {
```
rachedbenmustapha: Don't these parameters cause 100 concurrent iterations? How was resource usage during tests?

rachedbenmustapha (changes requested):

countItems.js
alexanderchan-scality: I am leaning towards coincidence.
I am looking at using the `autostart` and `onChild` params to allow for a more stable method for initializing the subprocesses, and relying on signals to properly close the connections.

countItems.js
alexanderchan-scality: you are completely correct. error is lost in the callback and `null` will be written

countItems.js
alexanderchan-scality: `worker-farm` does have a queue, and the number of concurrent open cursors will be `numWorkers * concurrent cursors`.

with 10 workers and 5 concurrent cursors each:  
* mongo is running at 200-300%.
* each worker is taking up 15-25% of cpu
* network i/o is about 80MiB/s

nicolas2bert (approved):

countItems.js
rachedbenmustapha: That seems a bit high, maybe confirm with @rahulreddy what the acceptable resource usage would be.

JianqinWang (approved):

countItems.js
alexanderchan-scality: resource usage wise:
there is a positive correlation between:
* number of cursors and the cpu usage of each worker
* number of workers and the cpu usage of mongo

remeasuring the time it takes to finish and the resources used

countItems.js
```diff
@@ -145,6 +214,7 @@ async.waterfall([
         }
         return next();
     }),
+    next => waitForSetup(err => next(err)),
```
rachedbenmustapha: `err => next(err)` is just `next`

rachedbenmustapha (commented):
Changes look good. However code complexity just went up a lot, I'd be more comfortable approving if there were tests validating the new mechanics in success/error cases.

countItems/counter.js
```diff
@@ -0,0 +1,122 @@
...
+function clientTeardown(callback) {
+    if (mongoClient.client && mongoClient.client.isConnected()) {
+        log.debug('mongoclient is connected...closing client');
+        return mongoClient.close(callback);
+    }
+    return callback();
+}
+
+let ranOnce = false;
+function handleTestMessages(data) {
```
rachedbenmustapha: @alexanderchan-scality I realize this is still WIP but better catch it early: please avoid test-related code with "am I testing?" conditionals inside non-test files. Everything test related (mocks for example) should be assembled/injected from test suites, as long as the design is flexible enough to allow it.

CountItems/utils/index.js
```diff
@@ -0,0 +1,64 @@
+const fs = require('fs');
+
+function callOnce(callback) {
```
rachedbenmustapha: Not that it's a lot of code, but Arsenal has `jsutils.once`.

CountItems/CountManager.js
```diff
@@ -0,0 +1,142 @@
...
+                total: { curr: 0, prev: 0 },
+                byLocation: {},
+            },
+            stalled: 0,
+        };
+        this.workerList = [];
+        this._setupQeueu();
+    }
+
+    _setupQeueu() {
```
rachedbenmustapha: Typo here

CountItems/CountWorker.js
```diff
@@ -0,0 +1,144 @@
+const uuid = require('node-uuid');
+const { callOnce } = require('./utils');
+
+class CountWorker {
```
rachedbenmustapha, line 5: Since this not actually worker code but the representation of a worker as seen from the master process (a proxy to a worker), I suggest a different name.

CountItems/Counter.js
```diff
@@ -0,0 +1,120 @@
+const assert = require('assert');
+const async = require('async');
+const { BucketInfo } = require('arsenal').models;
+
+class Counter {
```
rachedbenmustapha: This is the actual counter worker code.

rachedbenmustapha (changes requested):

rachedbenmustapha (approved):

alexanderchan-scality:
squash/rebased

anurag4DSB (approved):

tmacro (approved):

PR #89, opened 2020-03-06
bugfix: ZENKO-2404 Handle S3 Network Errors

naren-rajendran:
### Why is this change required? What problem does it solve?
Network related errors on S3 requests stops the execution of replication script, checks are added to handle these and allow script to retry on such errors.

### DIFF
--- a/crrExistingObjects.js
+++ b/crrExistingObjects.js
@@ -17,16 +17,18 @@ let STORAGE_TYPE = process.env.STORAGE_TYPE;
 let TARGET_REPLICATION_STATUS = process.env.TARGET_REPLICATION_STATUS;
 const TARGET_PREFIX = process.env.TARGET_PREFIX;
 const WORKERS = (process.env.WORKERS &&
-                 Number.parseInt(process.env.WORKERS, 10)) || 10;
+    Number.parseInt(process.env.WORKERS, 10)) || 10;
 const MAX_UPDATES = (process.env.MAX_UPDATES &&
-                     Number.parseInt(process.env.MAX_UPDATES, 10));
+    Number.parseInt(process.env.MAX_UPDATES, 10));
 const MAX_SCANNED = (process.env.MAX_SCANNED &&
-                     Number.parseInt(process.env.MAX_SCANNED, 10));
+    Number.parseInt(process.env.MAX_SCANNED, 10));
 let KEY_MARKER = process.env.KEY_MARKER;
 let VERSION_ID_MARKER = process.env.VERSION_ID_MARKER;
 
 const LISTING_LIMIT = 1000;
 const LOG_PROGRESS_INTERVAL_MS = 10000;
+const AWS_SDK_REQUEST_RETRIES = 100;
+const AWS_SDK_REQUEST_DELAY_MS = 30;
 
 if (!BUCKETS || BUCKETS.length === 0) {
     log.fatal('No buckets given as input! Please provide ' +
@@ -56,14 +58,14 @@ const replicationStatusToProcess = TARGET_REPLICATION_STATUS.split(',');
 replicationStatusToProcess.forEach(state => {
     if (!['NEW', 'PENDING', 'COMPLETED', 'FAILED', 'REPLICA'].includes(state)) {
         log.fatal('invalid TARGET_REPLICATION_STATUS environment: must be a ' +
-                  'comma-separated list of replication statuses to requeue, ' +
-                  'as NEW, PENDING, COMPLETED, FAILED or REPLICA.');
+            'comma-separated list of replication statuses to requeue, ' +
+            'as NEW, PENDING, COMPLETED, FAILED or REPLICA.');
         process.exit(1);
     }
 });
 log.info('Objects with replication status ' +
-         `${replicationStatusToProcess.join(' or ')} ` +
-         'will be reset to PENDING to trigger CRR');
+    `${replicationStatusToProcess.join(' or ')} ` +
+    'will be reset to PENDING to trigger CRR');
 
 const options = {
     accessKeyId: ACCESS_KEY,
@@ -80,7 +82,21 @@ const options = {
         agent: new http.Agent({ keepAlive: true }),
     },
 };
-const s3 = new AWS.S3(options);
+/**
+ *  Options specific to s3 requests
+ *  `maxRetries` & `customBackoff` are set only to s3 requests
+ *  default aws sdk retry count is 3 with an exponential delay of 2^n * 30 ms
+ */
+const s3Options = {
+    maxRetries: AWS_SDK_REQUEST_RETRIES,
+    customBackoff: (retryCount, error) => {
+        log.error('aws sdk request error', { error, retryCount });
+        // computed delay is not truly exponential, it is reset to minimum after
+        // every 10 calls, with max delay of 15 seconds!
+        return AWS_SDK_REQUEST_DELAY_MS * Math.pow(2, retryCount % 10);
+    },
+};
+const s3 = new AWS.S3(Object.assign(options, s3Options));
 const bb = new BackbeatClient(options);
 
 let nProcessed = 0;
@@ -108,15 +124,15 @@ function _objectShouldBeUpdated(objMD) {
     return replicationStatusToProcess.some(filter => {
         if (filter === 'NEW') {
             return (!objMD.replicationInfo ||
-                    objMD.replicationInfo.status === '');
+                objMD.replicationInfo.status === '');
         }
         return (objMD.replicationInfo &&
-                objMD.replicationInfo.status === filter);
+            objMD.replicationInfo.status === filter);
     });
 }
 
 function _markObjectPending(bucket, key, versionId, storageClass,
-                            repConfig, cb) {
+    repConfig, cb) {
     let objMD;
     let skip = false;
     return waterfall([
@@ -241,8 +257,8 @@ function _markPending(bucket, versions, cb) {
             const storageClass = Rules[0].Destination.StorageClass || SITE_NAME;
             if (!storageClass) {
                 const errMsg =
-                      'missing SITE_NAME environment variable, must be set to' +
-                      ' the value of "site" property in the CRR configuration';
+                    'missing SITE_NAME environment variable, must be set to' +
+                    ' the value of "site" property in the CRR configuration';
                 log.error(errMsg);
                 return next(new Error(errMsg));
             }
@@ -266,7 +282,7 @@ function triggerCRROnBucket(bucketName, cb) {
         KEY_MARKER = undefined;
         VERSION_ID_MARKER = undefined;
         log.info(`resuming at: KeyMarker=${KeyMarker} ` +
-                 `VersionIdMarker=${VersionIdMarker}`);
+            `VersionIdMarker=${VersionIdMarker}`);
     }
     doWhilst(
         done => _listObjectVersions(bucket, VersionIdMarker, KeyMarker,


crrExistingObjects.js
```diff
@@ -223,13 +240,45 @@ function _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
...
         VersionIdMarker,
         KeyMarker,
-    }, cb);
+    }, (err, data) => {
+        if(err) {
+            // retry only on Network errors and if the process is set to retry
+            // on errors.
+            if (err.code && err.code === 'NetworkingError' && RETRY_ON_ERROR) {
+                const errMsg = `Error listing object versions for bucket: ` +
+                                `${bucket}, retrying request..`;
```
rachedbenmustapha: No need for double or single punctuation or caps.

Please don't make variable error messages, instead embed the bucket name in the log message attributes (`{ error: err, bucketName: bucket }`), to make it searchable by attribute instead of full-text.

For more information on logging guidelines, please see the scality/guidelines github repo.

crrExistingObjects.js
```diff
@@ -80,7 +84,20 @@ const options = {
...
 };
-const s3 = new AWS.S3(options);
+/**
+ *  Options specific to s3 requests
+ *  `maxRetries` & `customBackoff` are set only to s3 requests
+ *  default aws sdk retry count is 3 with an exponential delay of 2^n * 30 ms
+ *  aws sdk retries are used first, before trying the requests recursively
+ *  on errors!
+ */
+const s3_options = {
```
rachedbenmustapha: name should be camel case.

crrExistingObjects.js
```diff
@@ -223,13 +240,45 @@ function _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
         Prefix: TARGET_PREFIX,
         VersionIdMarker,
         KeyMarker,
-    }, cb);
+    }, (err, data) => {
+        if(err) {
+            // retry only on Network errors and if the process is set to retry
+            // on errors.
+            if (err.code && err.code === 'NetworkingError' && RETRY_ON_ERROR) {
```
rachedbenmustapha: `err.code && err.code ===` is redundant.

crrExistingObjects.js
```diff
@@ -223,13 +244,45 @@ function _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
...
-    }, cb);
+    }, (err, data) => {
+        if (err) {
+            // retry only on Network errors and if the process is set to retry
+            // on errors.
+            if (err.code && err.code === 'NetworkingError' && RETRY_ON_ERROR) {
+                const errMsg = 'Error listing object versions for bucket: ' +
+                    `${bucket}, retrying request..`;
+                log.error(errMsg, { error: err });
+                return _listObjectVersions(bucket, VersionIdMarker,
```
rachedbenmustapha: The aws sdk is using its own retry with backoff mechanism and will eventually return an error back to us, but we need one also, otherwise this could be running in a loop without really failing.

rachedbenmustapha (changes requested):

crrExistingObjects.js
naren-rajendran: @RachedBenMustapha I pushed this commit to check linter errors :), you are quick!

crrExistingObjects.js
naren-rajendran: Yes, the change works on top of backoff mechanism, the request was to keep retrying on network errors until a fair response is returned. Introduced an environment variable RETRY_ON_ERROR to handle this.

crrExistingObjects.js
naren-rajendran: Just pushed another commit, fixed it.

crrExistingObjects.js
naren-rajendran: Ah, yes.

crrExistingObjects.js
rachedbenmustapha: A better name to describe it would be `RETRY_INDEFINITELY_ON_ERROR`, which is not an acceptable behavior.

crrExistingObjects.js
naren-rajendran: Corrected.

crrExistingObjects.js
```diff
@@ -223,13 +243,43 @@ function _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
         Prefix: TARGET_PREFIX,
         VersionIdMarker,
         KeyMarker,
-    }, cb);
+    }, (err, data) => {
+        if (err) {
+            // retry only on Network errors and if the process is set to retry
+            // on errors.
+            if (err.code === 'NetworkingError' && RETRY_ON_ERROR) {
+                log.error('Error listing object versions, retrying request',
```
rahulreddy: Logging in error would mean it generates a log dump, since this is an error that is 'handled' by retrying, we can put it as log.info. Also please keep the log messages lower cased, it makes it easy to search. Here are our logging guidelines for future reference https://github.com/scality/Guidelines/blob/development/8.1/CONTRIBUTING.md#logging

crrExistingObjects.js
```diff
@@ -223,13 +243,43 @@ function _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
...
+}
+
+// get bucket replication info
+function _getBucketReplication(options, cb) {
+    return s3.getBucketReplication(options, (err, res) => {
+        if (err) {
+            // retry only on Network errors and if the process is set to retry
+            // on errors.
+            if (err.code === 'NetworkingError' && RETRY_ON_ERROR) {
+                log.error('Error getting replication info, retrying request',
```
rahulreddy: Same here, log at the info level.

crrExistingObjects.js
```diff
@@ -223,13 +243,43 @@ function _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
...
+                    { error: err });
+                return _getBucketReplication(options, cb);
+            }
+            return cb(err);
```
rahulreddy: This is the error you want to log as error.

crrExistingObjects.js
```diff
@@ -223,13 +243,43 @@ function _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
...
+        if (err) {
+            // retry only on Network errors and if the process is set to retry
+            // on errors.
+            if (err.code === 'NetworkingError' && RETRY_ON_ERROR) {
+                log.error('Error listing object versions, retrying request',
+                    { error: err, bucketName: bucket });
+                return _listObjectVersions(bucket, VersionIdMarker,
+                    KeyMarker, cb);
+            }
+            return cb(err);
```
rahulreddy: Log at error level here

crrExistingObjects.js
```diff
@@ -80,7 +84,23 @@ const options = {
...
+ *  Options specific to s3 requests
+ *  `maxRetries` & `customBackoff` are set only to s3 requests
+ *  default aws sdk retry count is 3 with an exponential delay of 2^n * 30 ms
+ *  aws sdk retries are used first, before trying the requests recursively
+ *  on errors!
+ */
+const s3Options = {
+    maxRetries: AWS_SDK_REQUEST_RETRIES,
+    customBackoff: (retryCount, error) => {
+        log.debug('AWS sdk request error', { error, retryCount });
```
jonathan-gramain: This could be logged with `log.error`.

crrExistingObjects.js
```diff
@@ -80,7 +84,23 @@ const options = {
...
+        // computed delay is not truly exponential, it is reset to minimum after
```
jonathan-gramain, line 94: What is the rationale to reset to the minimum after 10 calls? Usually when we consider the retry delay to be too large, we keep it to a maximum value.

crrExistingObjects.js
jonathan-gramain: `log` is not a request logger but a top-level `Logger` object, so it would not dump the request context. But I agree we can make it an info log.

crrExistingObjects.js
```diff
@@ -17,16 +17,20 @@ let STORAGE_TYPE = process.env.STORAGE_TYPE;
...
+    Number.parseInt(process.env.WORKERS, 10)) || 10;
 const MAX_UPDATES = (process.env.MAX_UPDATES &&
-                     Number.parseInt(process.env.MAX_UPDATES, 10));
+    Number.parseInt(process.env.MAX_UPDATES, 10));
 const MAX_SCANNED = (process.env.MAX_SCANNED &&
-                     Number.parseInt(process.env.MAX_SCANNED, 10));
+    Number.parseInt(process.env.MAX_SCANNED, 10));
 let KEY_MARKER = process.env.KEY_MARKER;
 let VERSION_ID_MARKER = process.env.VERSION_ID_MARKER;
+const RETRY_ON_ERROR = (process.env.RETRY_ON_ERROR &&
```
jonathan-gramain: I think if we want to leave a choice to the user to tune the retry behavior, it's better to default to the most robust approach (retry x times or a certain amount of time) and possibly have a tunable to disable retries or configure their number. I'm not convinced it's very useful to offer this tunable here if we provide a reasonable default behavior.

crrExistingObjects.js
jonathan-gramain: Indeed there's always a point in time where it's better to give up than keep retrying forever. Considering the use case, I think a reasonable default could be to give up after 15mn of retries since that could be caused by a rather serious network issue that requires attention from an admin.

For the backoff algorithm, you may use or get inspired from the backbeat one in lib/BackbeatTask.js

jonathan-gramain (changes requested):

crrExistingObjects.js
rachedbenmustapha: Yes exactly. To illustrate, if some action needed to be taken to correct whatever issue is happening, failing is the best way to trigger an alert and let ops people know. Not failing is the best way to delay any useful response.

crrExistingObjects.js
naren-rajendran: The delay is too long if we don't repeat in cycles, with the current setup every tenth call takes 15 seconds to retry, this could be a tunable as well. I thought it would be better if we decide what the maximum delay could be.

crrExistingObjects.js
naren-rajendran: Yes, the callback is logging it as error.

crrExistingObjects.js
naren-rajendran: Yes, the callback is logging it as error.

rahulreddy:
@jonathan-gramain , @RachedBenMustapha , @naren-scality I recommend having an discussion together to make progress on this.

crrExistingObjects.js
naren-rajendran: Removed `RETRY_ON_ERROR` and as suggested, s3 requests on failure will be retried 100 (a set constant) using the aws sdk's native feature. `customBackOff` is modified to try out these 100 requests over a span of 5 minutes (computed constant). Recursion is also removed. So the errors will eventually stop the process after 100 retries.

crrExistingObjects.js
naren-rajendran: Recursion is removed and retry (on errors) count is set to 100. `customBackOff` is modified to try these 100 requests in 5 minutes.

naren-rajendran:
As suggested, recursion is now removed and retry count (on errors) is set to 100. `customBackOff` is modified to try these errors in 5 minutes.

crrExistingObjects.js
jonathan-gramain: I mean I don't see a reason to repeat a cycle, where we could cap the max retry time and stick to it for further retries:

```
    return DELAY * pow(2, min(retryCount, 10))
```

It's not a big issue for me just find the current logic a bit odd.

rachedbenmustapha (approved):

jonathan-gramain (approved):

naren-rajendran:
@bert-e approve

PR #94, opened 2020-03-23
improvement: ZENKO-2412 use mongodb directly

naren-rajendran:
#### Why is this change required? What problem does it solve?
Replication script adds more load to cloudserver/backbeat, to reduce this calls are made to mongodb directly.

### DIFF
--- a/CrrExistingObjects/listingParser.js
+++ b/CrrExistingObjects/listingParser.js
@@ -1,0 +1,24 @@
+function listingParser(entries) {
+    if (!entries) {
+        return entries;
+    }
+    return entries.map(entry => {
+        const tmp = JSON.parse(entry.value);
+        return {
+            Key: entry.key,
+            Size: tmp['content-length'],
+            ETag: tmp['content-md5'],
+            VersionId: tmp.versionId,
+            IsNull: tmp.isNull,
+            IsDeleteMarker: tmp.isDeleteMarker,
+            LastModified: tmp['last-modified'],
+            Owner: {
+                DisplayName: tmp['owner-display-name'],
+                ID: tmp['owner-id'],
+            },
+            StorageClass: tmp['x-amz-storage-class'],
+        };
+    });
+}
+
+module.exports = listingParser;
--- a/CrrExistingObjects/metadataClient.js
+++ b/CrrExistingObjects/metadataClient.js
@@ -1,0 +1,39 @@
+const { MetadataWrapper } = require('arsenal').storage.metadata;
+const werelogs = require('werelogs');
+const listingParser = require('./listingParser');
+
+const loggerConfig = {
+    level: 'info',
+    dump: 'error',
+};
+werelogs.configure(loggerConfig);
+
+const log = new werelogs.Logger('s3utils::crrExistingObjects');
+const replicaSetHosts = process.env.MONGODB_REPLICASET;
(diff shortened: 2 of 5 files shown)


anurag4DSB:
as discussed offline, more changes are being done to this.
Will keep an eye.

CrrExistingObjects/listingParser.js
```diff
@@ -0,0 +1,24 @@
+function listingParser(entries) {
+    if (!entries) {
+        return entries;
+    }
+    return entries.map(entry => {
+        const tmp = JSON.parse(entry.value);
```
jonathan-gramain, line 6: It might be possible that `entry.value` can be undefined, for safety I would recommend handling this case specifically.

CrrExistingObjects/metadataClient.js
```diff
@@ -0,0 +1,38 @@
...
+const params = {
+    customListingParser: listingParser,
+    mongodb: {
+        replicaSetHosts,
+        writeConcern,
+        replicaSet,
+        readPreference,
+        database,
+    },
+    replicationGroupId: 'RG001',
```
jonathan-gramain: this should be configurable, as it can (should) be set differently on a multi-site deployment.

jonathan-gramain (commented):
Looks good to me overall, I left two small requests in comments.

CrrExistingObjects/metadataClient.js
naren-rajendran: Is it only `replicationGroupId`? or others too? I checked the util deployment, it does not have a config file or a reader or env property `replicationGroupId`, do you want to introduce a new env variable?

jonathan-gramain: I don't know what this variable is used for, but if it relies on it to fetch keys in mongodb, it might fail if the deployment has been configured with a different replication group ID. If the client does not rely on it to fetch keys, you can probably ignore this. (Otherwise, yes, you can make it configurable with an env var).

naren-rajendran: `replicationGroupId` is used for generating versionId's, which is in turn used in creating metadata objects. Since we are creating new objects using the script, I will introduce a new env var.

jonathan-gramain: I suggest to make it a mandatory env var, to force the user to consider it and avoid mistakes.

CrrExistingObjects/listingParser.js
naren-rajendran: I checked, so the custom parser function is set at `metadataWrapper` constructor and `metadataWrapper` handles parsing errors, https://github.com/scality/Arsenal/blob/development/8.1/lib/storage/metadata/MetadataWrapper.js#L256

jonathan-gramain (approved):
I realize that this change will tie this version of crrExistingObjects to Zenko, and make it incompatible with S3C.

So far we have been pretty liberal and use the 8.1 version occasionally on S3C, we will need to become stricter and have development/7.4 dedicated to S3C and 8.1 dedicated to Zenko from now on. This will also likely call for a couple 8.1 -> 7.4 backports of fixes that have been done on the 8.1 branch. I will create a ticket to keep track of this.

vrancurel (approved):

naren-rajendran:
@bert-e approve

PR #105, opened 2020-04-07
feature: S3C-2753 script to remove delete markers on versioning-suspe…

jonathan-gramain:
…nded bucket

Add a new script removeDeleteMarkers.js to remove all (or prefixed)
delete markers from a versioning-suspended bucket.

This is meant to be a workaround for the MD-712 issue, where listing
times out due to too many delete markers.

### DIFF
--- a/removeDeleteMarkers.js
+++ b/removeDeleteMarkers.js
@@ -1,0 +1,285 @@
+/* eslint-disable max-len */
+/* eslint-disable no-console */
+
+const http = require('http');
+const async = require('async');
+const AWS = require('aws-sdk');
+const { Logger } = require('werelogs');
+
+const log = new Logger('s3utils::removeDeleteMarkers');
+// configurable params
+
+const DEFAULT_WORKERS = 10;
+const DEFAULT_LOG_PROGRESS_INTERVAL = 10;
+const DEFAULT_LISTING_LIMIT = 1000;
+
+const USAGE = `
+removeDeleteMarkers.js
+
+This script removes delete markers from one or more versioning-suspended bucket(s)
+
+Usage:
+    node removeDeleteMarkers.js bucket1[,bucket2...]
+
+Mandatory environment variables:
+    ENDPOINT: S3 endpoint
+    ACCESS_KEY: S3 account access key
+    SECRET_KEY: S3 account secret key
+
+Optional environment variables:
+    TARGET_PREFIX: only process a specific prefix in the bucket(s)
+    WORKERS: concurrency value for listing / batch delete requests (default ${DEFAULT_WORKERS})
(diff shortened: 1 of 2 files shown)


removeDeleteMarkers.js
```diff
@@ -0,0 +1,285 @@
...
+    objectsListed: 0,
+    deleteMarkersDeleted: 0,
+    deleteMarkersErrors: 0,
+    bucketInProgress: null,
+    keyMarker: KEY_MARKER,
+    versionIdMarker: VERSION_ID_MARKER,
+};
+
+function logProgress(message) {
+    log.info(message, {
```
miniscruff, line 110: Could you just pass in `status` as the second parameter for this? Or are you copying it manually for some specific reason I do not know about?

removeDeleteMarkers.js
```diff
@@ -0,0 +1,285 @@
...
+const LOG_PROGRESS_INTERVAL = (
+    process.env.LOG_PROGRESS_INTERVAL
+        && Number.parseInt(process.env.LOG_PROGRESS_INTERVAL, 10))
+      || DEFAULT_LOG_PROGRESS_INTERVAL;
+
+AWS.config.update({
+    accessKeyId: ACCESS_KEY,
+    secretAccessKey: SECRET_KEY,
+    endpoint: ENDPOINT,
+    region: 'us-east-1',
```
miniscruff, line 86: With all the other env vars, would having aws region be one help? Maybe an optional var thats defaulted to us east 1? I don't know if us-east is chosen here for a specific reason though.

removeDeleteMarkers.js
```diff
@@ -0,0 +1,285 @@
...
+                    keyMarker: data.NextKeyMarker,
+                    versionIdMarker: data.NextVersionIdMarker,
+                });
+            } else {
+                bucketDone = true;
+            }
+            return next(null, data.DeleteMarkers);
+        }),
+        (deleteMarkers, next) => {
+            if (deleteMarkers.length === 0) {
```
miniscruff, line 186: Is it common to compare to zero length over just truthy or falsey of the value?
I would expect `if (!deleteMarkers)` to be equal here. Assuming its array like and 0 items is falsey.

nicolas2bert (approved):

naren-rajendran:
**Question:** do we ever limit the number of `WORKERS`?

naren-rajendran (approved):

jonathan-gramain:
@naren-scality yes, WORKERS limits the number of batch deletes that can happen at the same time.

jonathan-gramain:
@bert-e approve

rahulreddy (approved):

PR #106, opened 2020-04-10
bugfix: ZENKO-2536 crrExistingObjects: update microVersionId

jonathan-gramain:
- use ObjectMD model instead of raw parsed JSON object

- update the new microVersionId field before putting the metadata
  changes, to force MongoDB to apply the change and generate an event
  in the oplog

- side fix: call MetadataWrapper.close() so that the script exits
  after completion

### DIFF
--- a/crrExistingObjects.js
+++ b/crrExistingObjects.js
@@ -1,5 +1,6 @@
 const { doWhilst, eachSeries, eachLimit, waterfall, series } = require('async');
 const { Logger } = require('werelogs');
+const { ObjectMD } = require('arsenal').models;
 const metadataUtil = require('./CrrExistingObjects/metadataUtils');
 
 const log = new Logger('s3utils::crrExistingObjects');
@@ -74,11 +75,11 @@ const logProgressInterval = setInterval(_logProgress, LOG_PROGRESS_INTERVAL_MS);
 function _objectShouldBeUpdated(objMD) {
     return replicationStatusToProcess.some(filter => {
         if (filter === 'NEW') {
-            return (!objMD.replicationInfo ||
-                objMD.replicationInfo.status === '');
+            return (!objMD.getReplicationInfo() ||
+                    objMD.getReplicationInfo().status === '');
         }
-        return (objMD.replicationInfo &&
-            objMD.replicationInfo.status === filter);
+        return (objMD.getReplicationInfo() &&
+                objMD.getReplicationInfo().status === filter);
     });
 }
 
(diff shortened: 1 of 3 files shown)


crrExistingObjects.js
```diff
@@ -94,12 +95,12 @@ function _markObjectPending(bucket, key, versionId, storageClass,
...
         }, log, next),
         (mdRes, next) => {
-            objMD = mdRes;
+            objMD = new ObjectMD(mdRes);
             if (!_objectShouldBeUpdated(objMD)) {
                 skip = true;
                 return next();
             }
-            if (objMD.versionId) {
+            if (objMD.getVersionId()) {
```
miniscruff, line 103: Further follow up on the other PR, the original `versionId` would be used if `getVersionId` was instead a get method.

package.json
```diff
@@ -26,7 +26,7 @@
   },
   "homepage": "https://github.com/scality/s3utils#readme",
   "dependencies": {
-    "arsenal": "github:scality/arsenal#development/8.1",
+    "arsenal": "github:scality/arsenal#07a7d5e",
```
miniscruff: Curious as to the decision to use Github urls and git hashes instead of semantic versioning like everything else?

jonathan-gramain: I don't think we maintain the semantic versions regularly on each package, hence we use the commit hashes instead. It might not be the best I agree, here I changed it because other projects rely on commit hashes not dev branch names so it's more consistent and works even if the dev branch hasn't been updated yet.

Maybe @rahulreddy can give a more elaborate answer if you're interested, as this topic has some history.

rachedbenmustapha: There's also the general unreliability of yarn (which is an improvement over npm, but still sometimes generates installations that are not identical to previous ones, everything else being equal)

miniscruff: Hmm I know years ago this was difficult for npm / yarn to properly tree out all dependencies but as far as I know now this should generally be a non-issue. I would generally recommend using semver for 99% of use cases at this point. Calver and hash versioning have use cases but not for installed packages IMO. 

This all probably relates to release management that I have no clue about yet so obviously take anything I say with "sure but you are missing this and that".

vrancurel (approved):

package.json
rachedbenmustapha: It was not a non-issue for us, and not years ago.

miniscruff: OK, no worries. I had assumed there was some issue but that I just didn't have enough information. Curious as to what may be causing these as I have not run into it myself before.

crrExistingObjects.js
```diff
@@ -302,6 +304,7 @@ function triggerCRROnBucket(bucketName, cb) {
 series([
     next => metadataUtil.metadataClient.setup(next),
     next => eachSeries(BUCKETS, triggerCRROnBucket, next),
+    next => metadataUtil.metadataClient.close(next),
```
naren-rajendran, line 307: Curious, did `metadataClient.close()` work? When I was working on this, it threw an error IIRC! Also, checked other repos where metadataClient was used and `metadataClient.close()` was not used, I may be wrong!

jonathan-gramain: works for me :)

naren-rajendran (approved):

jonathan-gramain:
@bert-e approve

PR #107, opened 2020-04-14
bugfix: ZENKO-2555 Fix bug on updating metadata

naren-rajendran:
#### Whis is this change required? What problem does it solve?

crrExistingObjects script fails and exits when updating metadata for an object. The script is modified to log the error on such case and continue.

### DIFF
--- a/CrrExistingObjects/metadataUtils.js
+++ b/CrrExistingObjects/metadataUtils.js
@@ -25,7 +25,10 @@ function listObjectVersions(params, log, cb) {
         keyMarker: params.KeyMarker,
         versionIdMarker: params.VersionIdMarker,
     };
-
+    log.debug('listing object versions', {
+        method: 'metadataUtils.listObjectVersions',
+        listingParams,
+    });
     return metadataClient.listObject(bucketName, listingParams, log,
         (err, list) => {
             if (err) {
@@ -49,7 +52,6 @@ function _formatConfig(config) {
             },
         };
     });
-
     return {
         ReplicationConfiguration: {
             Role: role,
@@ -60,6 +62,10 @@ function _formatConfig(config) {
 
 function getBucketReplication(options, log, cb) {
     const bucketName = options.Bucket;
+    log.debug('getting bucket replication', {
+        method: 'metadataUtils.getBucketReplication',
+        bucket: bucketName,
+    });
     return metadataClient.getBucket(bucketName, log, (err, data) => {
         if (err) {
             return cb(err);
@@ -85,26 +91,33 @@ function _getNullVersion(objMD, bucketName, objectKey, log, cb) {
 }
 
 function getMetadata(params, log, cb) {
+    const { Bucket, Key } = params;
     let versionId = params.VersionId;
+    log.debug('getting object metadata', {
+        method: 'metadataUtils.getMetadata',
+        bucket: Bucket,
+        objectKey: Key,
+        versionId,
+    });
     if (versionId && versionId !== 'null') {
         versionId = versionIdUtils.decode(versionId);
     }
-
     if (versionId instanceof Error) {
         const errMsg = 'Invalid version id specified';
         return cb(errors.InvalidArgument.customizeDescription(errMsg));
     }
-
     const mdParams = {
         versionId,
     };
-    const { Bucket, Key } = params;
     return metadataClient.getBucketAndObjectMD(Bucket, Key, mdParams, log,
         (err, data) => {
             if (err) {
                 return cb(err);
             }
             const objMD = data.obj ? JSON.parse(data.obj) : undefined;
+            if (objMD === undefined) {
+                return cb(new Error('could not get metadata'));
+            }
             if (objMD && versionId === 'null') {
                 return _getNullVersion(objMD, Bucket, Key, log,
                     (err, nullVer) => {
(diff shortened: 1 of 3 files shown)


crrExistingObjects.js
```diff
@@ -95,7 +95,11 @@ function _markObjectPending(bucket, key, versionId, storageClass,
             VersionId: versionId,
         }, log, next),
         (mdRes, next) => {
+            if (mdRes === undefined) {
+                return next(new Error('could not get metadata'));
```
jonathan-gramain: If this is an error, it could be more appropriate to raise it from `getMetadata` (better details about the error, more consistent).

CrrExistingObjects/metadataUtils.js
```diff
@@ -85,26 +91,33 @@ function _getNullVersion(objMD, bucketName, objectKey, log, cb) {
...
         versionId,
     };
-    const { Bucket, Key } = params;
     return metadataClient.getBucketAndObjectMD(Bucket, Key, mdParams, log,
         (err, data) => {
             if (err) {
                 return cb(err);
             }
             const objMD = data.obj ? JSON.parse(data.obj) : undefined;
+            if (objMD === undefined) {
```
jonathan-gramain, line 118: Out of curiosity, do you know in which case objMD can be undefined?

jonathan-gramain (approved):

crrExistingObjects.js
```diff
@@ -1,9 +1,17 @@
 const { doWhilst, eachSeries, eachLimit, waterfall, series } = require('async');
-const { Logger } = require('werelogs');
+const werelogs = require('werelogs');
 const { ObjectMD } = require('arsenal').models;
 const metadataUtil = require('./CrrExistingObjects/metadataUtils');
 
-const log = new Logger('s3utils::crrExistingObjects');
+const logLevel = Number.parseInt(process.env.DEBUG, 10) === 1
```
miniscruff, line 6: these changes seem out of place in the PR and might be better split into a new one.

CrrExistingObjects/metadataUtils.js
```diff
@@ -85,26 +91,33 @@ function _getNullVersion(objMD, bucketName, objectKey, log, cb) {
 }
 
 function getMetadata(params, log, cb) {
+    const { Bucket, Key } = params;
```
miniscruff, line 94: I think these two parameters should be lower case, eg `bucket` and `key`

crrExistingObjects.js
```diff
@@ -1,9 +1,17 @@
...
+    ? 'debug' : 'info';
+const loggerConfig = {
+    level: logLevel,
+    dump: 'error',
+};
+werelogs.configure(loggerConfig);
+const log = new werelogs.Logger('s3utils::crrExistingObjects');
+log.debug('debug');
```
miniscruff: is this just a console.log at this point? Feels like it can be removed but not sure.

crrExistingObjects.js
naren-rajendran: This is just an extension of the PR, and the change is required to have more information on the process. It was discussed before I included them.

CrrExistingObjects/metadataUtils.js
naren-rajendran: It was made/kept because the original requests to s3 client expected them. Also at the time of making the original improvement (which is to use mongodbWrapper instead of cloudserver and backbeat) there was a discussion to have both options and switch based on user option. We may revisit this and hence to avoid breaking changes, it was intentionally left as it is.

crrExistingObjects.js
naren-rajendran: Yes, it was a mistake. Removed.

crrExistingObjects.js
jonathan-gramain: Agreed, it's a small complement to ease debugging that we can keep here IMO.

miniscruff: kk no problem, I like to include this additions in the description so reviewers know what to expect.

CrrExistingObjects/metadataUtils.js
```diff
@@ -25,7 +25,10 @@ function listObjectVersions(params, log, cb) {
         keyMarker: params.KeyMarker,
         versionIdMarker: params.VersionIdMarker,
     };
-
+    log.debug('listing object versions', {
+        method: 'metadataUtils.listObjectVersions',
```
miniscruff, line 29: random question here and kinda unrelated, but is there anyway to get the current method or context from `this` that could be more dynamic? Or maybe automated by werelogs in some way? I need to learn more about this but I would hate for our logs to show the wrong method cause we renamed it but forgot to update our logs. Sort of like how comments get out of date.

naren-rajendran: IMO, I don't think we can grab it from `this`, at least in the current context. Agreed that the manual handling of this could be troublesome. There are non standard ways of grabbing it apparently.

CrrExistingObjects/metadataUtils.js
naren-rajendran: The only case I can think of from the implementation is, if the key does not exist.
https://github.com/scality/cloudserver/blob/development/8.2/lib/metadata/metadataUtils.js#L62
https://github.com/scality/Arsenal/blob/development/8.1/lib/storage/metadata/mongoclient/MongoClientInterface.js#L262

CrrExistingObjects/metadataUtils.js
naren-rajendran: It would be cool to have something like `Reflections` a la `c#` to get method/function names in js.

ilkescality (approved):

rahulreddy (approved):

naren-rajendran:
@bert-e approve

PR #108, opened 2020-04-16
feature: S3C-2733 new script repairDuplicateVersions.js

jonathan-gramain:
Add a new s3utils script to repair objects that have duplicate
metadata across versions, i.e. duplicate sproxyd keys, that are at
risk of data loss due to S3C-2731.

The script takes the output log from verifyBucketSproxydKeys.js script
as input, and repairs objects that have been detected with duplicate
keys.

The repair step consists of copying each duplicated sproxyd key data
to a new key, and updating metadata of one of the object versions to
point to the new keys.

### DIFF
--- a/repairDuplicateVersions.js
+++ b/repairDuplicateVersions.js
@@ -1,0 +1,435 @@
+/* eslint-disable max-len */
+/* eslint-disable no-console */
+/* eslint-disable comma-dangle */
+
+const async = require('async');
+const crypto = require('crypto');
+const http = require('http');
+const { URL } = require('url');
+const readline = require('readline');
+
+const { jsutil, errors } = require('arsenal');
+const { Logger } = require('werelogs');
+
+const {
+    BUCKETD_HOSTPORT, SPROXYD_HOSTPORT,
+} = process.env;
+
+const USAGE = `
+repairDuplicateVersions.js
+
+This script repairs object versions that share sproxyd keys with
+another version, particularly due to bug S3C-2731.
+
+The repair action consists of copying the data location that is
+duplicated to a new sproxyd key (or set of keys for MPU), and updating
+the metadata to reflect the new location, resulting in two valid
+versions with distinct data, though identical in content.
+
+The script does not remove any version even if the duplicate was due
+to an internal retry in the metadata layer, because either version
+might be referenced by S3 clients in some cases.
+
+Usage:
+    node repairDuplicateVersions.js
+
+Standard Input:
+    The standard input must be fed with the JSON logs output by the
+    verifyBucketSproxydKeys.js s3utils script. This script only
+    processes the log entries containing the message "duplicate
+    sproxyd key found" and ignores other entries.
+
+Mandatory environment variables:
+    BUCKETD_HOSTPORT: ip:port of bucketd endpoint
+    SPROXYD_HOSTPORT: ip:port of sproxyd endpoint
+`;
+
+if (!BUCKETD_HOSTPORT) {
+    console.error('ERROR: BUCKETD_HOSTPORT not defined');
+    console.error(USAGE);
+    process.exit(1);
+}
+if (!SPROXYD_HOSTPORT) {
+    console.error('ERROR: SPROXYD_HOSTPORT not defined');
+    console.error(USAGE);
+    process.exit(1);
+}
+const log = new Logger('s3utils:repairDuplicateVersions');
+
+const httpAgent = new http.Agent({
+    keepAlive: true,
+});
+
+let sproxydAlias;
+const objectsToRepair = [];
+
+const status = {
+    logLinesRead: 0,
+    objectsRepaired: 0,
+    objectsSkipped: 0,
+    objectsErrors: 0,
+    sproxydKeysCopied: 0,
+    sproxydKeysCopyErrors: 0,
+    sproxydBytesCopied: 0,
+};
+
+function logProgress(message) {
+    log.info(message, Object.assign({}, status, {
+        objectsToRepair: objectsToRepair.length,
+    }));
+}
+
+function httpRequest(method, url, reqBody, cb) {
+    const cbOnce = jsutil.once(cb);
+    const urlObj = new URL(url);
+    const req = http.request({
+        hostname: urlObj.hostname,
+        port: urlObj.port,
+        path: `${urlObj.pathname}${urlObj.search}`,
+        method,
+        agent: httpAgent,
+    }, res => {
+        if (method === 'HEAD') {
+            return cbOnce(null, res);
+        }
+        const chunks = [];
+        res.on('data', chunk => chunks.push(chunk));
+        res.once('end', () => {
+            const body = chunks.join('');
+            // eslint-disable-next-line no-param-reassign
+            res.body = body;
+            return cbOnce(null, res);
+        });
+        return res.once('error', err => cbOnce(new Error(
+            'error reading response from HTTP request '
+                + `to ${url}: ${err.message}`
+        )));
+    });
+    req.once('error', err => cbOnce(new Error(
+        `error sending HTTP request to ${url}: ${err.message}`
+    )));
+    if (reqBody) {
+        req.setHeader('content-type', 'application/json');
+        req.setHeader('content-length', reqBody.length);
+        req.write(reqBody);
+    }
+    req.end();
+}
+
+function getSproxydAlias(cb) {
+    const url = `http://${SPROXYD_HOSTPORT}/.conf`;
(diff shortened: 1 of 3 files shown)


repairDuplicateVersions.js
```diff
@@ -0,0 +1,447 @@
...
+    objectsRepaired: 0,
+    objectsSkipped: 0,
+    objectsErrors: 0,
+    sproxydKeysCopied: 0,
+    sproxydKeysCopyErrors: 0,
+    sproxydBytesCopied: 0,
+};
+
+function logProgress(message) {
+    log.info(message, {
```
miniscruff: is there a reason you do not just pass in status? or `{...status}`?

repairDuplicateVersions.js
```diff
@@ -0,0 +1,447 @@
...
+                return undefined;
+            }
+            objectsToRepair.push({
+                objectUrl: parsedLine.objectUrl,
+                objectUrl2: parsedLine.objectUrl2,
+            });
+        } catch (err) {
+            log.info('ignoring malformed JSON line');
+        }
+        return undefined;
```
miniscruff, line 159: can just `return` here

repairDuplicateVersions.js
```diff
@@ -0,0 +1,447 @@
...
+        method: 'GET',
+        agent: httpAgent,
+    }, sourceRes => {
+        if (sourceRes.statusCode !== 200) {
+            log.error('sproxyd returned HTTP error code', {
+                objectUrl,
+                sproxydKey,
+                httpCode: sourceRes.statusCode,
+            });
+            sourceRes.on('data', () => {});
```
miniscruff: is this required?

repairDuplicateVersions.js
```diff
@@ -0,0 +1,447 @@
...
+        });
+        targetReq.once('error', err => {
+            log.error('error sending data to sproxyd', {
+                objectUrl,
+                sproxydKey: newKey,
+                error: err.message,
+            });
+            return cbOnce(errors.InternalError);
+        });
+        return undefined;
```
miniscruff: can just return here too

repairDuplicateVersions.js
```diff
@@ -0,0 +1,447 @@
+/* eslint-disable max-len */
```
miniscruff, line 1: This file doesn't have any tests, not sure the standard for testing in this project

repairDuplicateVersions.js
```diff
@@ -0,0 +1,447 @@
...
+        cb();
+    });
+}
+
+function fetchObjectMetadata(objectUrl, cb) {
+    if (!objectUrl.startsWith('s3://')) {
+        return cb(new Error(`malformed object URL ${objectUrl}: must start with "s3://"`));
+    }
+    const bucketAndObject = objectUrl.slice(5);
+    const url = `http://${BUCKETD_HOSTPORT}/default/bucket/${bucketAndObject}`;
```
rachedbenmustapha, line 172: Is there a chance that this script needs to be run at one of the sites that have enabled end-to-end TLS?

repairDuplicateVersions.js
```diff
@@ -0,0 +1,447 @@
...
+                status.objectsErrors += 1;
+            }
+            done();
+        });
+    }, cb);
+}
+
+function main() {
+    async.series([
+        done => getSproxydAlias(done),
```
rachedbenmustapha: Minor: this can be written as `[getSproxydAlias, readVerifyLog, repairObjects]`

repairDuplicateVersions.js
```diff
@@ -0,0 +1,447 @@
...
+        done => readVerifyLog(done),
+        done => repairObjects(done),
+    ], err => {
+        if (err) {
+            log.error('an error occurred during repair process', {
+                error: err.message,
```
rachedbenmustapha: Minor: the script's output will probably not be indexed, but in case it is, it's safer to do `error: { message: err.message }` otherwise the indexes choke on mismatched types for `error` being sometimes an object and sometimes a string.

repairDuplicateVersions.js
jonathan-gramain: It's a valid point, originally s3utils started as a small set of simple scripts, now that it's becoming bigger and used for more and more things the question of automated tests arise. For the detection script `verifyBucketSproxydKeys` I added a few unit tests for specific helper functions that are complex enough to benefit from them, here I did not but probably a couple functions could be tested as well.

We don't have the support needed for integration or end-to-end tests on s3utils, but with mocking we could likely enhance test coverage with some effort, but has not been done so far. When I come back to this script I will see if some tests can be added with reasonably low effort.

repairDuplicateVersions.js
jonathan-gramain: Indeed, there is one value `objectsToRepair` that is different but I could simplify for the other values using `Object.assign()`

repairDuplicateVersions.js
jonathan-gramain: Yes, though our linter rules are a bit pedantic and ask to return a value each time. We could argue that the linter is wrong and fix the linter rules but for consistency with other projects I just obeyed the linter remarks :wink:

repairDuplicateVersions.js
```diff
@@ -0,0 +1,447 @@
...
+                return cb();
+            }
+            const objMD = results.objectUrl.md;
+            objMD.location.forEach(loc => {
+                if (copiedKeys[loc.key]) {
+                    // eslint-disable-next-line no-param-reassign
+                    loc.key = copiedKeys[loc.key];
+                }
+            });
+            return putObjectMetadata(objInfo.objectUrl, objMD, err => {
```
rachedbenmustapha, line 369: On non-versioned buckets, could this mean that out of bad luck there could be a race condition if the same object is being actively written to, or is there a guard in bucketd/MD?

If there's a risk, maybe the doc could mention that it's preferable to shut down incoming requests while this script is running.

repairDuplicateVersions.js
jonathan-gramain: I'm not sure but it could likely be replaced by a `sourceRes.resume()` call instead, so that the data is consumed. I think we need to consume the data explicitly though because it's not used elsewhere.

rachedbenmustapha (approved):

repairDuplicateVersions.js
rachedbenmustapha: My personal opinion is that the linter is wrong although sadly it has authority over us.

repairDuplicateVersions.js
jonathan-gramain: Yes, it potentially could have to run at any of our S3 customers. Reading between the lines, I agree we should make it configurable with https as well.

repairDuplicateVersions.js
jonathan-gramain: This script is only relevant for versioned buckets, unless we discover a new bug that affects nonversioned buckets in the same way (unlikely). But versions could still be modified by `PutObjectAcl` or `PutObjectTagging` so the problem can exist.

Yes we could mention in the docs that it can be safer to shut down incoming traffic as the script should be quick to run it should be acceptable, even if the risk looks small since the number of affected objects should hopefully be very small compared to the total number of objects.

repairDuplicateVersions.js
miniscruff: ok wasn't sure if there was any more or fewer items here.

repairDuplicateVersions.js
miniscruff: Yea return undefined here feels weird, I wouldn't reject the PR for this if the linter says its wrong but would for sure approve of an eslint change to not require this.

repairDuplicateVersions.js
miniscruff: Gotcha, I think no project is too small for tests but I wouldn't hold up a PR for it given my late arrival. You do you :)

repairDuplicateVersions.js
jonathan-gramain: I opened an issue to keep track of it: https://scality.atlassian.net/browse/S3C-2807

repairDuplicateVersions.js
jonathan-gramain: Re-reading it, my bad, I agree we should get rid of it because we're doing `sourceReq.pipe(targetReq)` below.

jonathan-gramain: Ah no, looking even more closely, this is the error case, so we're not reading the response otherwise and need to consume it.

rachedbenmustapha (approved):

rahulreddy (approved):

jonathan-gramain:
@bert-e approve

PR #111, opened 2020-06-26
improvement: ZENKO-2661 option to pass keys to check in stdin

jonathan-gramain:
Add the KEYS_FROM_STDIN environment variable to allow to check only the specified list of keys read from stdin

The input is then a stream of JSON objects, each containing a "bucket"
and "key" attribute (key can optionally have a versionId appended to
it, separated from the key by a null character).

### DIFF
--- a/verifyBucketSproxydKeys.js
+++ b/verifyBucketSproxydKeys.js
@@ -5,6 +5,7 @@
 const http = require('http');
 const async = require('async');
 const { URL } = require('url');
+const jsonStream = require('JSONStream');
 
 const { jsutil } = require('arsenal');
 const { Logger } = require('werelogs');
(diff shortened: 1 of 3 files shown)


verifyBucketSproxydKeys.js
```diff
@@ -80,15 +86,15 @@ Optional environment variables:
...
+if (!BUCKETS && !RAFT_SESSIONS && !KEYS_FROM_STDIN) {
+    console.error('ERROR: either BUCKETS or RAFT_SESSIONS or KEYS_FROM_STDIN environment '
                   + 'variable must be defined');
     console.error(USAGE);
     process.exit(1);
 }
-if (BUCKETS && RAFT_SESSIONS) {
-    console.error('ERROR: BUCKETS and RAFT_SESSIONS environment variables '
-                  + 'cannot be both defined');
+if ((BUCKETS ? 1 : 0) + (RAFT_SESSIONS ? 1 : 0) + (KEYS_FROM_STDIN ? 1 : 0) > 1) {
```
miniscruff, line 95: A little bit of a stretch again...
Edit: That makes no sense re-reading the details above...

Can convert to booleans and just add. JS will let you add 3 booleans and its treated as 0/1.
```suggestion
if (BUCKETS + RAFT_SESSIONS + KEYS_FROM_STDIN > 1) {
```
You can also combine this with the previous check and simply do `!=1` as the error messages are about the same.

miniscruff (approved):

dora-korpar (approved):

jonathan-gramain:
@bert-e approve

PR #112, opened 2020-08-18
Feature/s3 c 3271 compare raft logs

vrancurel:


vrancurel:
not yet cleaned up / linted but the tool works

raftLogCheck.js
```diff
@@ -0,0 +1,165 @@
...
+			    if (err) {
+				console.log(err.message);
+				process.exit(1);
+			    }
+			    if (verbose) {
+				console.log('leaderLog:', leaderLog);
+			    }
+			    async.each(connected, (member, cb) => {
+				getUrl(
+				    `http://${leaderHost}:${leaderPort}/_/raft/log?begin=${seq}&limit=${limit}`,
```
jonathan-gramain: here it should be `member.host/port` not `leaderHost/Port`

raftLogCheck.js
```diff
@@ -0,0 +1,165 @@
...
+							memberLog.log[i].entries === undefined) {
+							console.log('undefined entries at', seq + i, j);
+							continue ;
+						    }
+						    if (leaderLog.log[i].entries[j].key === undefined) {
+							console.log('undefined leader key at', seq + i, j);
+							continue ;
+						    }
+						    if (leaderLog.log[i].entries[j].value === undefined) {
+							console.log('undefined leader value at', seq + i, j);
```
jonathan-gramain: not all entries have values (thinking about `type="del"`)

raftLogCheck.js
```diff
@@ -0,0 +1,165 @@
...
+						    }
+						    if (memberLog.log[i].entries[j].key === undefined) {
+							console.log('undefined member key at', seq + i, j);
+							continue ;
+						    }
+						    if (memberLog.log[i].entries[j].value === undefined) {
+							console.log('undefined member value at', seq + i, j);
+							continue ;
+						    }
+						    if (md5(memberLog.log[i].entries[j].key) !=
```
jonathan-gramain: md5 is not really needed here, we can just compare the string values against each other

jonathan-gramain (changes requested):

raftLogCheck.js
```diff
@@ -0,0 +1,183 @@
...
+							    console.log(memberLog.log[i].entries[j]);
+							}
+							continue ;
+						    }
+						    if (memberLog.log[i].entries[j].key !=
+							leaderLog.log[i].entries[j].key) {
+							console.log('key mismatch at',
+								    seq + i, j, memberLog.log[i].entries[j].key);
+						    }
+						    if (md5(memberLog.log[i].entries[j].value) !=
```
jonathan-gramain: same here :smile:

And then you can remove the dependency on md5 module as a bonus :)

package.json
```diff
@@ -26,12 +26,14 @@
   },
   "homepage": "https://github.com/scality/s3utils#readme",
   "dependencies": {
+    "JSONStream": "^1.3.5",
     "arsenal": "github:scality/arsenal#81d05b6",
     "async": "^2.6.1",
     "aws-sdk": "^2.333.0",
-    "JSONStream": "^1.3.5",
+    "md5": "^2.3.0",
```
rahulreddy: Do we really need this dependency? This should be sufficient, right?

```js
const crypto = require('crypto');
const hex = crypto.createHash('md5').update('foo').digest('hex');
```

package.json
```diff
@@ -26,12 +26,14 @@
...
     "node-uuid": "^1.4.8",
     "werelogs": "scality/werelogs",
+    "yarn": "^1.22.4",
```
rahulreddy: yarn is not needed as a dependency

raftLogCheck.js
```diff
@@ -0,0 +1,235 @@
+const http = require('http');
```
rahulreddy, line 1: It would be nice if we can support tls, quite a few of our high profile customer deployments use tls

raftLogCheck.js
```diff
@@ -0,0 +1,235 @@
...
+                                }
+                            }
+                            if (memberLog.log[i].entries[j].key !==
+                                leaderLog.log[i].entries[j].key) {
+                                log.debug('key mismatch at', {
+                                    seq: seq + i,
+                                    off: j,
+                                    key: memberLog.log[i].entries[j].key });
+                            }
+                            if (md5(memberLog.log[i].entries[j].value) !==
```
jonathan-gramain: You can remove the `md5()` calls as well as the dependency on `md5` module, I don't see any benefit in using md5 in this case.

raftLogCheck.js
```diff
@@ -0,0 +1,235 @@
...
+    log.debug('leaderLog:', { leaderLog });
+    async.each(connected, (member, cb) => {
+        getUrl(
+            `http://${member.host}:${member.adminPort}` +
+            `/_/raft/log?begin=${seq}&limit=${pageSize}`,
+            (err, memberLog) => {
+                if (err) {
+                    return cb(err);
+                }
+                log.info('getting member page', { member, seq });
```
rahulreddy: Create and use a request logger here and end it at the end of the iteratee function, if a log dump happens because of an error it will dump all logs and may crash the process.

raftLogCheck.js
```diff
@@ -0,0 +1,235 @@
...
+}
+
+const bucketdHost = myArgs[0];
+const bucketdPort = myArgs[1];
+const raftSession = myArgs[2];
+
+function getUrl(url, cb) {
+    log.debug('url', { url });
+    http.get(url, res => {
+        let body = '';
```
jonathan-gramain: It could be more efficient to use an array of chunks instead of a string to avoid copying, and at the end, join it with `.join('')`. I think it matters here because returned result could be multi-megabytes and there might be many such calls to check all the logs.

raftLogCheck.js
```diff
@@ -0,0 +1,235 @@
...
+                                            { seq: seq + i, entry: j });
+                                continue;
+                            }
+                            if (leaderLog.log[i].entries[j].key === undefined) {
+                                if (leaderLog.log[i].method === 0 || // CREATE
+                                    leaderLog.log[i].method === 1 || // DELETE
+                                    leaderLog.log[i].method === 7) { // PUT_ATTR
+                                    expectKey = false;
+                                } else {
+                                    log.warning('undefined leader key at', {
```
rahulreddy: ~log.warning~
`log.warn`

raftLogCheck.js
```diff
@@ -0,0 +1,235 @@
...
+                     i++) {
+                    if (leaderLog.log !== undefined &&
+                        leaderLog.log[i] !== undefined &&
+                        leaderLog.log[i].entries !== undefined) {
+                        for (j = 0;
+                             j < leaderLog.log[i].entries.length;
+                             j++) {
+                            let expectKey = true;
+                            let expectValue = true;
+                            if (memberLog.log === undefined ||
```
jonathan-gramain: This check could be done one level up

rahulreddy (changes requested):

raftLogCheck.js
```diff
@@ -0,0 +1,235 @@
...
+                                        seq: seq + i,
+                                        off: j,
+                                        entry: leaderLog.log[i].entries[j] });
+                                    continue;
+                                }
+                            }
+                            if (memberLog.log[i].entries[j].key ===
+                                undefined) {
+                                if (expectKey) {
+                                    log.debug('undefined member key at', {
```
jonathan-gramain: Is it supposed to be a warning log, not debug? Same question for the following debug logs

raftLogCheck.js
```diff
@@ -0,0 +1,235 @@
...
+                                    value: memberLog.log[i].entries[j].value });
+                            }
+                        }
+                    }
+                }
+                return cb();
+            });
+    }, err => {
+        if (err) {
+            log.error('error each', { error: err.message });
```
jonathan-gramain: "error each"?

raftLogCheck.js
jonathan-gramain: It's actually the internal TLS not the external nginx-configured TLS, but yes some customers use this

jonathan-gramain: repd admin does not support https yet, but a bug is open about this: https://scality.atlassian.net/browse/S3C-2612

rahulreddy: Ah right, I forgot about this.

rahulreddy (approved):

jonathan-gramain (approved):

PR #116, opened 2020-12-04
bf: S3C-3019 s3utils fails to build

naren-rajendran:
Node v8 is not sufficient to build the package mongodb-memory-server, bumping it to v10.


dora-korpar (approved):
👍

jonathan-gramain (approved):

PR #118, opened 2020-12-28
bugfix: ZENKO-2909 stalled retry

alexanderchan-scality:
Refactors existing stalledRetry script to to fix possible OOM issue when the mongo cursor's `.toArray` method is called; `toArray` will attempt to load all the entries into memory, and, if the cursor result becomes too large, the script will fail due to memory issues. 

On a collection with 5.1M Entries (average entry size 1.1KB):

stalled retry (incomplete run):
stalledRetry_mem_timeline

refactor (complete run):
stalled_retry_rewrite_mem_timeline

with the limits to match values set in stalled retry:
* batch: 10
* queue: 1000
* concurrent_requests: 5 

(blue: mem in use, gray: mem gc'd)

### DIFF
--- a/StalledRetry/CursorWrapper.js
+++ b/StalledRetry/CursorWrapper.js
@@ -1,0 +1,231 @@
+const MongoError = require('mongodb').MongoError;
+const { encode } = require('arsenal').versioning.VersionID;
+const { once } = require('arsenal').jsutil;
+
+class StalledEntry {
+    constructor(bucketName, key, versionId, storageClass, forceRetry) {
+        this.bucketName = bucketName;
+        this.key = key;
+        this.versionId = versionId;
+        this.storageClass = storageClass;
+        this.forceRetry = forceRetry;
+    }
+
+    toObject() {
+        return {
+            Bucket: this.bucketName,
+            Key: this.key,
+            VersionId: this.versionId,
+            StorageClass: this.storageClass,
+            ForceRetry: this.forceRetry,
+        };
+    }
+}
+
+function objectToEntries(bucketName, cmpDate, data) {
+    if (!data || typeof data !== 'object' ||
+        !data.value || typeof data.value !== 'object') {
+        return [];
+    }
+
+    const time = data.value['last-modified'] || null;
+    if (isNaN(Date.parse(time))) {
+        return [];
+    }
+
+    const testDate = new Date(time);
+    const withinRange = testDate <= cmpDate;
+    if (!withinRange) {
+        return [];
+    }
+
+    const storageClasses = data._id.storageClasses.split(',');
+    return storageClasses.map(i => {
+        const storageClass = i.split(':')[0];
+        return new StalledEntry(
+            bucketName,
+            data._id.key,
+            encode(data._id.versionId),
+            storageClass,
+            true
+        );
+    });
+}
+
+class RateLimitingCursor {
+    constructor(cursor, params) {
+        this.info = {
+            stalled: 0,
+        };
+
+        this.bucketName = params.bucketName;
+        this.cmpDate = params.cmpDate;
+        this.log = params.log;
+
+        this.queueLimit = params.queueLimit || 1000;
+
+        this.buffer = [];
+        this.getBatchCallbacks = [];
+        this.cursor = cursor;
+        this.cursorErr = null;
+        this.cursorEnd = false;
+
+        this.completed = false;
+
+        this._init();
+    }
+
+    _init() {
+        this.cursor.on('data', data => {
+            const entries = objectToEntries(
+                this.bucketName,
+                this.cmpDate,
+                data) || [];
+
+            if (entries.length === 0) {
+                return;
+            }
+
+            this.info.stalled += entries.length;
+            this.buffer.push(...entries);
+
+            this._fulfillGetBatch();
+
+            if (this.buffer.length >= this.queueLimit) {
+                this.cursor.pause();
+            }
+        });
+
+        this.cursor.on('end', () => {
+            this.log.debug('reached end of cursor', {
+                inQueue: this.buffer.length,
+            });
+            this.cursorEnd = true;
+            this._cleanUp();
+        });
+
+        this.cursor.on('error', err => {
+            this.log.error('encountered error while reading cursor', {
+                error: err,
+            });
+
+            if (
+                err instanceof MongoError &&
+                err.errorLabels.includes('TransientTransactionError')
+            ) {
+                this.log.info('transient error, continue reading from cursor');
+                return;
+            }
+
+            this.cursorEnd = true;
+            this.cursorErr = err;
+
+            this.buffer = [];
+            this._cleanUp();
+        });
+    }
+
+    _fulfillGetBatch() {
+        if (this.getBatchCallbacks.length === 0) {
+            return;
+        }
+
+        while (this.buffer.length > 0 && this.getBatchCallbacks.length > 0) {
+            const i = this.getBatchCallbacks.length - 1;
+            // TODO: implement a min-heap for callback on size
+            // atm, size is the same for all getBatch requests, so only the
+            // last items is checked
+            if (this.getBatchCallbacks[i].size <= this.buffer.length) {
+                const { size, cb } = this.getBatchCallbacks.pop();
+                const batch = this.buffer.splice(0, size);
+                if (batch.length === 0) {
+                    cb(null, null);
+                } else {
+                    cb(null, batch);
+                }
+            } else {
+                break;
+            }
+        }
+
+        if (this.cursor.isPaused() && this.buffer.length < this.queueLimit) {
+            this.cursor.resume();
+        }
+
+        if (this.cursorEnd && this.buffer.length === 0) {
+            this._cleanUp();
+        }
+    }
+
+    getInfo() {
+        return this.info;
+    }
+
+    _cleanUp() {
+        if (this.cursor && typeof this.cursor.destroy === 'function') {
+            this.cursor.destroy();
+            this.cursor = null;
+        }
+
+        this.getBatchCallbacks.forEach(({ size, cb }) => {
+            if (this.cursorErr !== null) {
+                return cb(this.cursorErr, null);
+            }
+
+            if (this.completed) {
+                return cb(null, null);
+            }
+
+            const batch = this.buffer.splice(0, size);
+            if (this.buffer.length === 0) {
+                this.completed = true;
+            }
+
+            if (batch.length === 0) {
+                return cb(null, null);
+            }
+
+            return cb(null, batch);
+        });
+
+        this.getBatchCallbacks = [];
+        if (this.buffer.length === 0) {
+            this.completed = true;
+        }
+    }
+
+    getBatch(size, cb) {
+        if (this.cursorErr !== null) {
+            cb(this.cursorErr, null);
+            return;
+        }
+
+        if (this.completed) {
+            cb(null, null);
+            return;
+        }
+
+        if (size === 0) {
+            cb(null, []);
+            return;
+        }
+
+        this.getBatchCallbacks.push({ size, cb: once(cb) });
+
+        if (this.cursor && this.cursor.isPaused()) {
+            this._fulfillGetBatch();
+        }
+
+        if (this.cursorEnd) {
+            this._cleanUp();
+        }
+
+        return;
+    }
+}
+
+module.exports = {
+    objectToEntries,
+    RateLimitingCursor,
+    StalledEntry,
+};
--- a/StalledRetry/MongoClientInterfaceStalled.js
+++ b/StalledRetry/MongoClientInterfaceStalled.js
@@ -1,0 +1,84 @@
+const async = require('async');
+
+const {
+    MongoClientInterface,
+} = require('arsenal').storage.metadata.mongoclient;
+
+class MongoClientInterfaceStalled extends MongoClientInterface {
+    constructor(params) {
+        super(params);
+        this.cursorWrapperFactory = params.cursorWrapperFactory;
+        this.requestHandlerFactory = params.requestHandlerFactory;
+    }
+
+    _getStalledObjectsByBucket(bucketName) {
+        const c = this.getCollection(bucketName);
+        const reducedFields = {
+            '_id': {
+                id: '$_id',
+                storageClasses: '$value.replicationInfo.storageClass',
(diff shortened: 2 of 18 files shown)


tcarmet (dismissed):
LGTM, for what I can review (mostly ci related stuff).

I would suggest to squash my commit into `remove memory mongodb dependency` or to amend title.

tmacro (dismissed):

countItems.js
```diff
@@ -1,8 +1,22 @@
 const cluster = require('cluster');
 
+const { parseEnvInt } = require('./utils');
+
+const HEAP_PROFILER_INTERVAL =
+    parseEnvInt(process.env.HEAP_PROFILER_INTERVAL, 10 * 60 * 100);
```
jonathan-gramain: Is it not using the traditional millisecond unit? I would have expected to multiply by 1000 then somewhere to convert seconds to milliseconds, but apparently not the case, it could deserve a quick comment on how much time this represents.

utils/heapProfiler.js
```diff
@@ -0,0 +1,42 @@
...
+                error: err,
+            });
+        } else {
+            logger.debug(`heapdump written to ${filename}:`);
+        }
+    });
+}
+
+module.exports = (dir, interval) => {
+    if (!!process.env.ENABLE_HEAP_PROFILER) {
```
jonathan-gramain: Is it intended to apply boolean operations on a string value coming from the env directly? Did you mean to convert the value to a boolean first (e.g. via `Number.parseInt`) so that e.g. `'0'` becomes `0` then `false`?

StalledRetry/CursorWrapper.js
```diff
@@ -0,0 +1,156 @@
...
+            }
+
+            if (this.cursorEnd && batch.length === 0) {
+                return cb(null, null);
+            }
+
+            return cb(null, batch);
+        }
+
+        return setTimeout(
```
jonathan-gramain: Why not, instead of using a random timer, queuing the requests as `(size, cb)`, and calling `getBatch` from `cursor.on('data')` and unqueuing if the request can be fullfilled. Not that the current approach is necessarily bad, actually if it is good in practice, I'm good with it, just wondering if it could cause the process to wait too much.

StalledRetry/MongoClientInterfaceStalled.js
```diff
@@ -0,0 +1,94 @@
...
+            { $match: {
+                '_id.id': { $regex: /\0/ },
+                '_id.status': { $eq: 'PENDING' },
+            } },
+        ]);
+    }
+
+    queueStalledObjects(cb) {
+        const cmpDate = new Date();
+        cmpDate.setHours(cmpDate.getHours() - 1);
```
jonathan-gramain: Maybe the "1 hour late" factor could be made configurable via an env variable

StalledRetry/MongoClientInterfaceStalled.js
```diff
@@ -0,0 +1,94 @@
...
+        const reqHandler = this.requestHandlerFactory(this.logger);
+
+        let stalledCount = 0;
+        this.db.listCollections().toArray((err, collections) => {
+            if (err) {
+                return cb(err);
+            }
+            return async.eachLimit(collections, 1, (value, next) => {
```
jonathan-gramain: if you always want one at a time, `async.eachSeries` can be used instead.

StalledRetry/MongoClientInterfaceStalled.js
```diff
@@ -0,0 +1,94 @@
...
+                const skipBucket = value.name === METASTORE ||
```
jonathan-gramain: minor style comment: could be written in more compact form as `[METASTORE, ...].includes(value.name) || ...`

StalledRetry/MongoClientInterfaceStalled.js
```diff
@@ -0,0 +1,94 @@
...
+                    bucketName,
+                    cmpDate,
+                    this._getStalledObjectsByBucket(bucketName),
+                    this.logger
+                );
+
+                return reqHandler.handleRequests(wrapper, (err, results) => {
+                    if (err) {
+                        this.logger.error(
+                            'encounted error while processing requests',
```
jonathan-gramain: encounte**re**d error

StalledRetry/MongoClientInterfaceStalled.js
```diff
@@ -0,0 +1,94 @@
...
+                            { method: 'queueStalledObjects', error: err }
+                        );
+                        return next(err);
+                    }
+
+                    this.logger.debug('completed handling requests', {
```
jonathan-gramain, line 67: even if debug log, may benefit from logging the bucket name

stalled.js
```diff
@@ -1,16 +1,44 @@
...
+ * @returns {int} - return integer value
+ */
+function parseEnvInt(env, defValue) {
+    return (env && !Number.isNaN(env))
+        ? Number.parseInt(env, 10)
+        : defValue;
+}
+
+const BATCH_SIZE = parseEnvInt(process.env.REQUEST_BATCH_SIZE, 10);
+const QUEUE_LIMT = parseEnvInt(process.env.QUEUE_LIMT, 1000);
```
jonathan-gramain: why not calling this QUEUE_LIM**I**T

countItems.js
alexanderchan-scality: this should have been 10 mins (10 * 60 * 1000)

utils/heapProfiler.js
alexanderchan-scality: will update this to something like:
```
if (process.env.ENABLE_HEAP_PROFILER &&
    process.env.ENABLE_HEAP_PROFILER !== '0')
```

eve/workers/master/pod.yaml
```diff
@@ -0,0 +1,32 @@
...
+          memory: 1Gi
+        limits:
+          cpu: "1"
+          memory: 1Gi
+      volumeMounts:
+        - name: worker-workspace
+          mountPath: /home/eve/workspace
+    - name: mongo
+      image: scality/ci-mongo:3.6.8
+      imagePullPolicy: IfNotPresent
```
tcarmet, line 22: From what I'm understanding you're changing the CI and local environments to use an actual mongodb host instead of a "in memory" server/library. (correct me if wrong)

What I wonder is the following:
* Is this explicit when a dev wants to work on s3utils? Or should it be listed somewhere in the readme to start a mongo db server on the side. For me, I see this on the `main.yml` I get it, but I wonder what is the answer for others.
* As this is a "new" change and I see this `scality/ci-mongo` beeing used a lot, does it reflect the version of mongo that s3utils will speeks to? Or should we consider upgrading it a bit, to an upstream image per example (under the namespace `mongo`  on dockerhub) worth noting that this comment may be out of scope.

tcarmet (approved):
LGTM, left some open comments, sharing thoughts I had while reviewing. But the content looks good from what I can actually review ;)

eve/workers/master/pod.yaml
alexanderchan-scality: @tcarmet 
> From what I'm understanding you're changing the CI and local environments to use an actual mongodb host instead of a "in memory" server/library. (correct me if wrong)

yes. The "in-memory" mongo introduces some flakiness, in which a teardown is not completed (port is not released), causing subsequent tests using mongo to fail.
 
> Is this explicit when a dev wants to work on s3utils? Or should it be listed somewhere in the readme to start a mongo db server on the side. For me, I see this on the main.yml I get it, but I wonder what is the answer for others.

Can you clarify what you mean by explicit? 
For development/testing, a mongodb server will be needed.

> As this is a "new" change and I see this scality/ci-mongo beeing used a lot, does it reflect the version of mongo that s3utils will speeks to? Or should we consider upgrading it a bit, to an upstream image per example (under the namespace mongo on dockerhub) worth noting that this comment may be out of scope.

Using the official upstream mongodb image is great idea. `scality/ci-mongo:3.6.8` was used to be consistent with what is used in zenko 1.2.x and cloudserver project.

countItems.js
```diff
@@ -1,8 +1,22 @@
 const cluster = require('cluster');
 
+const { parseEnvInt } = require('./utils');
+
+const HEAP_PROFILER_INTERVAL_MS =
+    parseEnvInt(process.env.HEAP_PROFILER_INTERVAL_MS, 10 * 60 * 1000);
+const HEAP_PROFILER_PATH = process.env.HEAP_PROFILER_PATH;
+
 if (cluster.isMaster) {
+    require('./utils/heapProfiler')(
```
rachedbenmustapha, line 10: Is this going to enable the heap profiler at all times indiscriminately?

StalledRetry/CursorWrapper.js
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    const testDate = new Date(time);
+    const withinRange = testDate <= cmpDate;
+    if (!withinRange) {
+        return [];
+    }
+
+    const storageClasses = data._id.storageClasses.split(',');
+    return storageClasses.map(i => {
```
rachedbenmustapha, line 43: This creates one entry per destination for `PENDING` entries if I'm understanding correctly; shouldn't it only return destinations that are in `PENDING` state? as some of the destinations could be `COMPLETED`.

StalledRetry/CursorWrapper.js
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    _init() {
+        this.cursor.on('data', data => {
+            const entries = objectToEntries(
+                this.bucketName,
+                this.cmpDate,
+                data) || [];
+
+            if (entries.length < 0) {
```
rachedbenmustapha: can this really happen?

rachedbenmustapha (commented):
Can we have a rough idea of the memory usage reduction?

StalledRetry/CursorWrapper.js
alexanderchan-scality: will fix. the logic should be `<=`

countItems.js
alexanderchan-scality: this is only enabled when the env `ENABLE_HEAP_PROFILER` is set

StalledRetry/CursorWrapper.js
alexanderchan-scality: Correct me if I am wrong here as well. In the case where the sites have different statuses, the global state would have been set to `PROCESSING`, signaling that the entry is not stalled.

The cursor will only filter for the global state, `PENDING`, so the sites are assumed to not have been processed.

alexanderchan-scality:
@RachedBenMustapha Will do. I will rerun my benchmarks for the numbers

StalledRetry/CursorWrapper.js
rachedbenmustapha: I think you're right, for some reason I thought that `PROCESSING` was a synthetic status that was not actually persisted to MD.

StalledRetry/CursorWrapper.js
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        while (this.buffer.length > 0 && this.getBatchCallbacks.length > 0) {
+            const i = this.getBatchCallbacks.length - 1;
+            // TODO: implement a min-heap for callback on size
+            // atm, size is the same for all getBatch requests, so only the
+            // last items is checked
+            if (this.getBatchCallbacks[i].size <= this.buffer.length) {
+                const { size, cb } = this.getBatchCallbacks.pop();
+                cb(null, this.buffer.splice(0, size));
```
rachedbenmustapha: Is this (and other callbacks used in this class) synchronous on purpose?

StalledRetry/CursorWrapper.js
```diff
@@ -0,0 +1,201 @@
...
+            Bucket: bucketName,
+            Key: data._id.key,
+            VersionId: encode(data._id.versionId),
+            StorageClass: storageClass,
+            ForceRetry: true,
+        };
+    });
+}
+
+class StalledCursorWrapper {
```
rachedbenmustapha: It took me a minute to understand why this cursor was special and used in place of a standard mongodb cursor, maybe it should be reflected in the name: maybe something like "rate limiting cursor", "backpressure cursor", or "trickle cursor".

StalledRetry/StalledRequestHandler.js
```diff
@@ -0,0 +1,135 @@
...
+                }
+
+                if (batch !== null) {
+                    this.queue.push({ batch, getNext: nextBatch });
+                }
+
+                return cb(null);
+            });
+
+        this.queue.error = (err, batch) => {
```
rachedbenmustapha: It feels wrong to assign the error callback at this point in the code. Is there no way of setting it up during object construction, or have a handler in the request pipeline?

StalledRetry/StalledRequestHandler.js
```diff
@@ -0,0 +1,135 @@
...
+                        error: err,
+                    });
+                    return onceCB(err);
+                }
+
+                // wait for queue to be done and source to be consumed
+                return this._waitForCompletion(err => {
+                    if (err) {
+                        this.killed = true;
+                        this.queue.error = null;
```
rachedbenmustapha: Same here, changing the error handler itself seems fragile, can it handle the case and be a no-op when needed by itself?

rachedbenmustapha (commented):
Last part of review

tmacro (approved):

jonathan-gramain (approved):

rachedbenmustapha:
@alexanderchan-scality it seems commit 803a535b9e1995f3823406d84964a0688acea139 undoes some of the previous commit's changes

alexanderchan-scality:
@RachedBenMustapha will fix the `<=` to `===`, there shouldn't be a case where that goes negative

stalled.js
```diff
@@ -1,16 +1,36 @@
...
 const ZenkoClient = require('zenkoclient');
 
-const { MongoClientInterface } = storage.metadata.mongoclient;
-const { encode } = versioning.VersionID;
+const {
+    MongoClientInterfaceStalled,
+} = require('./StalledRetry/MongoClientInterfaceStalled');
+
+const {
+    BackpressureCursor,
```
rachedbenmustapha: I believe the class is now named `RateLimitingCursor`, `BackpressureCursor ` would be undefined if imported this way.

rachedbenmustapha (changes requested):

rachedbenmustapha (approved):

PR #119, opened 2021-01-15
S3C-3806 script to delete noncurrent versions and delete markers

jonathan-gramain:
Add a new script to delete noncurrent versions and delete markers, as
part of s3utils.

I tried to test it manually as exhaustively as possible, I believe it's working well but please review thoroughly if you find any issue with it.

### DIFF
--- a/cleanupNoncurrentVersions.js
+++ b/cleanupNoncurrentVersions.js
@@ -1,0 +1,461 @@
+const http = require('http');
+
+const AWS = require('aws-sdk');
+const { doWhilst, eachSeries } = require('async');
+
+const { Logger } = require('werelogs');
+
+const log = new Logger('s3utils::cleanupNoncurrentVersions');
+const BUCKETS = process.argv[2] ? process.argv[2].split(',') : null;
+const ACCESS_KEY = process.env.ACCESS_KEY;
+const SECRET_KEY = process.env.SECRET_KEY;
+const S3_ENDPOINT = process.env.S3_ENDPOINT;
+const TARGET_PREFIX = process.env.TARGET_PREFIX;
+const MAX_DELETES = (process.env.MAX_DELETES &&
+    Number.parseInt(process.env.MAX_DELETES, 10));
+const MAX_LISTED = (process.env.MAX_LISTED &&
+    Number.parseInt(process.env.MAX_LISTED, 10));
+const MARKER = process.env.MARKER;
+const OLDER_THAN = (process.env.OLDER_THAN ?
+                    new Date(process.env.OLDER_THAN) : null);
+
+const LISTING_LIMIT = 1000;
+const LOG_PROGRESS_INTERVAL_MS = 10000;
+const AWS_SDK_REQUEST_RETRIES = 100;
+const AWS_SDK_REQUEST_INITIAL_DELAY_MS = 30;
+
+const USAGE = `
+cleanupNoncurrentVersions.js
+
+This script removes noncurrent versions and current/noncurrent delete
+markers, either all such objects or older than a specified
+last-modified date.
+
+Usage:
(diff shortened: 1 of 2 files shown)


cleanupNoncurrentVersions.js
```diff
@@ -0,0 +1,439 @@
...
+    return process.nextTick(cb);
+}
+
+function triggerDeletesOnBucket(bucketName, cb) {
+    const bucket = bucketName.trim();
+    bucketInProgress = bucket;
+    log.info(`starting task for bucket: ${bucket}`);
+    let NextKeyMarker = null;
+    let NextVersionIdMarker = null;
+    if (KEY_MARKER || VERSION_ID_MARKER) {
```
rachedbenmustapha: It seems that the key marker and delete id marker should be linked somehow to the bucket they belong to, as here they are read/written as global variables by the `triggerDeletesOnBucket` call on the first bucket even if they in fact belonged to the second bucket in the list (the use case for this would be someone resuming an interrupted run with the same command and "forgetting" to drop the buckets already done from the command line).

It would be a valid requirement IMO that they explicitly apply only to the first bucket listed (or that only one bucket is provided if they are present), in which case mentioning it in the documentation or checking that there's only one bucket argument would be enough.

Edit: just saw that the end-summary log message gives an updated list of buckets so this comment can be dismissed if we trust that users won't "forget" to use the updated bucket list when re-running.

cleanupNoncurrentVersions.js
```diff
@@ -0,0 +1,439 @@
...
+            process.nextTick(batchDeleteOnDrain);
+            batchDeleteOnDrain = null;
+        }
+        if (batchDeleteOnFullDrain && deleteQueue.length === 0) {
+            process.nextTick(batchDeleteOnFullDrain);
+            batchDeleteOnFullDrain = null;
+        }
+        if (deleteQueue.length > 0) {
+            // there are more objects to delete, keep going
+            _doBatchDelete(bucket);
```
rachedbenmustapha, line 237: The list of pending deletes may grow too big before this becomes a problem so maybe not applicable (depending on how heap vs stack works in nodejs, not sure): doing this in a `nextTick` may conserve stack space.

rachedbenmustapha (approved):
LGTM

naren-rajendran (approved):

cleanupNoncurrentVersions.js
jonathan-gramain: Indeed, not fond of this, maybe instead something that could work better is providing a single env var to pass containing the triplet bucket-key-version, although handling properly delimiters may be challenging (like internally we use `\0` for separating key and version, may not be appropriate since env vars cannot contain `\0` IIRC)

jonathan-gramain: I'll likely keep it as is unless I have a "eureka" moment later!

jonathan-gramain: Actually we could use URL encoding for all, and encode the separators as some non-encoded character

cleanupNoncurrentVersions.js
jonathan-gramain: I believe it is unnecessary since `deleteObjects` should always be asynchronous, as being an HTTP call.

cleanupNoncurrentVersions.js
jonathan-gramain: @RachedBenMustapha Did the change, now we can keep the same bucket list and a single `MARKER='bucket|key|version'` env var.

I also renamed ENDPOINT to S3_ENDPOINT to have a cleaner integration with the orchestration script that will also need BUCKETD_ENDPOINT and SPROXYD_ENDPOINT.

rachedbenmustapha: Nice approach

PR #136, opened 2021-05-06
ft: S3C-4337 find compactable ranges

vrancurel:
Very simple algorithm to find ranges that are potentially compactable.

### DIFF
--- a/ldbreclaim/ldbreclaim.go
+++ b/ldbreclaim/ldbreclaim.go
@@ -1,0 +1,290 @@
+// The ldbreclaim finds the most interesting ranges to compact
+package main
+
+import (
+	"bytes"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+
+	"github.com/golang/leveldb/db"
+	"github.com/golang/leveldb/table"
+)
+
+var (
+	verifyChecksums = flag.Bool("c", false, "Verify checksums")
+	verbose = flag.Bool("v", false, "Verbose")
+	dumpStats = flag.Bool("s", false, "Dump stats and exits")
+	numResults = flag.Int("n", 10, "Num results")
+	tempSpaceMb = flag.Int64("t", 500, "Max overwrite desired in MB")
+	minReclaimMb = flag.Int64("r", 300, "Minimum space to reclaim in MB")
+)
+
+func usage() {
+	fmt.Printf("usage: ldbreclaim [options] folder (use --help for options)\n")
+	os.Exit(1)
+}
+
+/*
+func SetVerbose(value bool) {
+	*verbose = value
+}
+*/
+
+type Ldb struct {
+	name string
+	size int64
+	startKey string
+	endKey string
+	numKeys int64 // setKeys + deletedKeys
+	numDeletedKeys int64
+	index int // for FindRange
+}
+
+// By is the type of a "less" function that defines the ordering of its Ldb arguments.
+type By func(p1, p2 *Ldb) bool
+
+// Sort is a method on the function type, By, that sorts the argument slice according to the function.
+func (by By) Sort(s []Ldb) {
+	ps := &ldbSorter{
+		ldbs:    s,
+		by:      by, // The Sort method's receiver is the function (closure) that defines the sort order.
+	}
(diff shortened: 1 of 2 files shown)


ldbreclaim/ldbreclaim.go
```diff
@@ -0,0 +1,287 @@
...
+		ldbs:    s,
+		by:      by, // The Sort method's receiver is the function (closure) that defines the sort order.
+	}
+	sort.Sort(ps)
+}
+
+// ldbSorter joins a By function and a slice of Ldbs to be sorted.
+type ldbSorter struct {
+	ldbs []Ldb
+	by      func(p1, p2 *Ldb) bool // Closure used in the Less method.
```
jonathan-gramain, line 62: could you declare `by` of type `By` instead?

ldbreclaim/ldbreclaim.go
```diff
@@ -0,0 +1,287 @@
...
+			resultsFound++
+		}
+		if resultsFound >= *numResults {
+			break
+		}
+	}
+}
+
+func Kind(k []byte) uint8 {
+        return k[len(k)-8]
```
jonathan-gramain, line 190: Could be interesting to have hints on the logic at play here, like a docstring for the `Kind` function and what is placed at `len(k)-8` or the general format of a key

jonathan-gramain (commented):
I'm not sure why dumping the ldbs with the max number of keys first will give good ranges to compact, otherwise, the method adding the number of keys and deleted keys for overlapping ranges looks legit to me.

Could a different final sorting order give better results? I don't know, maybe a ratio of deleted/total keys, or combined with the absolute number of keys to avoid dumping ranges too small, but still taking into account the number of deleted keys in that range.

vrancurel:
> Could a different final sorting order give better results? I don't know, maybe a ratio of deleted/total keys, or combined with the absolute number of keys to avoid dumping ranges too small, but still taking into account the number of deleted keys in that range.

If you just specify -t 500 -r 300 (default settings) it will display the ranges having worst case 500MB disk space and reclaim best case 300MB so 60% efficacy ratio.

vrancurel:
@cmrh asks for the output of the tool to be easily plugged into the `curl` route.

jonathan-gramain:
For reference, the compaction admin API: https://github.com/scality/MetaData/blob/development/7.10/documentation/PROTOCOL.ADMIN.md#compact-bucket-key-range

vrancurel:
> > Could a different final sorting order give better results? I don't know, maybe a ratio of deleted/total keys, or combined with the absolute number of keys to avoid dumping ranges too small, but still taking into account the number of deleted keys in that range.
> 
> If you just specify -t 500 -r 300 (default settings) it will display the ranges having worst case 500MB disk space and reclaim best case 300MB so 60% efficacy ratio.

Can you suggest a comparison function ? is the ordering total ?

cmrh (commented):
In an sample result as follows
`{001107.ldb 485027 compact-range/prefix-0346/data-file-0740.img compact-range/prefix-0898/data-file-0484.img 860474 308243 0}
{001109.ldb 484868 compact-range/prefix-0402/data-file-0774.img compact-range/prefix-0954/data-file-0543.img 796233 252451 0}
{001111.ldb 381957 compact-range/prefix-0458/data-file-0533.img compact-range/prefix-1000/data-file-1000.img 736238 196656 0}
{001104.ldb 1234225 compact-range/prefix-0499/data-file-0703.img compact-range/prefix-0800/data-file-1000.img 432878 131983 0}
`
You can either compact the data with the leveldb prefix as follows

`curl --header 'Content-Type: application/json' --request POST --data '{"prefix": "prefix-0868"}'  127.0.0.1:9146/_/buckets/compact-range/compactRange`
or alternatively use the startKey and endKey

`curl -v --header 'Content-Type: application/json' --request POST --data '{"startKey":"prefix-0458/data-file-0533.img","endKey":"prefix-1000/data-file-1000.img"}'  127.0.0.1:9146/_/buckets/compact-range/compactRange`
Note you cannot use the bucketName in the startKey/endKey

vrancurel:
Closing this PR and re-creating it in the Metadata repository as it must be ran from within the MD container.

PR #141, opened 2021-09-14
surpress error

watsaqat:
eachLimit will exit prematurely if an error is bubbled up from _requeueObject. However, suppressing the error in the async.waterfall callback allows the retry logic to continue.

### DIFF
--- a/autoRetryFailedCRR-S3C.js
+++ b/autoRetryFailedCRR-S3C.js
@@ -161,7 +161,13 @@ function _requeueObject(bucket, key, versionId, counters, cb) {
                 next(err);
             });
         },
-    ], cb);
+    ], err => {
+        if (err) {
+            log.error('error in _requeueObject waterfall',
+                      { error: err.message });
+        }
+        return cb();
+    });
     /* eslint-enable no-param-reassign */
 }
 


autoRetryFailedCRR-S3C.js
```diff
@@ -161,7 +161,14 @@ function _requeueObject(bucket, key, versionId, counters, cb) {
                 next(err);
             });
         },
-    ], cb);
+    ], err => {
+        if (err) {
+            log.error('error in _requeueObject waterfall',
+                      { error: err.message });
+            ++counters.errors;
```
jonathan-gramain: Errors are already counted when they happen in the inner call, so we should not count them a second time here.

autoRetryFailedCRR-S3C.js
```diff
@@ -161,7 +161,14 @@ function _requeueObject(bucket, key, versionId, counters, cb) {
...
+        } 
+        return cb()
```
jonathan-gramain:
```suggestion
        return cb();
```

jonathan-gramain (approved):

jonathan-gramain:
LGTM

PR #117, opened 2020-12-16
improvement: S3C-3719 auto replay failed replications

jonathan-gramain:
Fork the autoRetryFailedCRR Zenko script to make it work for S3C.

CUSTOMER-SPECIFIC SOLUTION, NOT MEANT TO BE MERGED IN DEV BRANCH

PR #142, opened 2021-09-21
improvement: S3UTILS-7 requeue failed CRR to kafka topic

jonathan-gramain:
Fork the autoRetryFailedCRR Zenko script to requeueFailedCRRCronJob:

- make it work for S3C

- requeue to the specified Kafka topic

- run in a cron schedule

This is a temporary solution until we productize the auto-retry logic.

@nicolas2bert @naren-rajendran This PR is almost identical to https://github.com/scality/s3utils/pull/117, except the script name has been changed and @watsaqat patch integrated (https://github.com/scality/s3utils/pull/141). Would you mind having a quick look since you already reviewed the original PR and don't need to go through all the script code again?

All reviewers, feel free to suggest an alternative name for the script if you don't like this one for any reason, I am not deeply satisfied with this name but could not find something better.

### DIFF
--- a/requeueFailedCRRCronJob.js
+++ b/requeueFailedCRRCronJob.js
@@ -1,0 +1,284 @@
+const { waterfall, doWhilst, eachLimit, retry } = require('async');
+const http = require('http');
+const { Producer } = require('node-rdkafka');
+const { scheduleJob } = require('node-schedule');
+
+const { errors } = require('arsenal');
+const VID_SEP = require('arsenal').versioning.VersioningConstants
+      .VersionId.Separator;
+const { Logger } = require('werelogs');
+
+const BackbeatClient = require('./BackbeatClient');
+
+const log = new Logger('s3utils::requeueFailedCRRCronJob');
+const {
+    CLOUDSERVER_ENDPOINT,
+    BACKBEAT_API_ENDPOINT,
+    ACCESS_KEY,
+    SECRET_KEY,
+    SITE_NAME,
+    KAFKA_HOSTS,
+    KAFKA_TOPIC,
+} = process.env;
+const CRON_RULE = process.env.CRON_RULE || '0 */6 * * *';
+
+
+if (!CLOUDSERVER_ENDPOINT) {
+    throw new Error('CLOUDSERVER_ENDPOINT not defined');
+}
+if (!BACKBEAT_API_ENDPOINT) {
+    throw new Error('BACKBEAT_API_ENDPOINT not defined');
+}
+if (!ACCESS_KEY) {
+    throw new Error('ACCESS_KEY not defined');
+}
+if (!SECRET_KEY) {
+    throw new Error('SECRET_KEY not defined');
+}
+if (!SITE_NAME) {
+    throw new Error('missing SITE_NAME environment variable, must be set to' +
+                    ' the value of "site" property in the CRR configuration');
+}
+if (!KAFKA_HOSTS) {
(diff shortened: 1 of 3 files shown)


jonathan-gramain:
ping

requeueFailedCRRCronJob.js
```diff
@@ -0,0 +1,282 @@
...
+const BackbeatClient = require('./BackbeatClient');
+
+const log = new Logger('s3utils::requeueFailedCRRCronJob');
+const CLOUDSERVER_ENDPOINT = process.env.CLOUDSERVER_ENDPOINT;
+const BACKBEAT_API_ENDPOINT = process.env.BACKBEAT_API_ENDPOINT;
+const ACCESS_KEY = process.env.ACCESS_KEY;
+const SECRET_KEY = process.env.SECRET_KEY;
+const SITE_NAME = process.env.SITE_NAME;
+const KAFKA_HOSTS = process.env.KAFKA_HOSTS;
+const KAFKA_TOPIC = process.env.KAFKA_TOPIC;
```
miniscruff: kinda weird but you can destructure process.env such as
```js
const { ACCESS_KEY, SITE_NAME } = process.env
```

naren-rajendran (approved):

requeueFailedCRRCronJob.js
jonathan-gramain: done

miniscruff (approved):

requeueFailedCRRCronJob.js
```diff
@@ -0,0 +1,284 @@
...
+    ACCESS_KEY,
+    SECRET_KEY,
+    SITE_NAME,
+    KAFKA_HOSTS,
+    KAFKA_TOPIC,
+} = process.env;
+const CRON_RULE = process.env.CRON_RULE || '0 */6 * * *';
+
+
+if (!CLOUDSERVER_ENDPOINT) {
```
watsaqat, line 26: `ReferenceError` could be used here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ReferenceError

requeueFailedCRRCronJob.js
```diff
@@ -0,0 +1,284 @@
...
+                    ' the value of "site" property in the CRR configuration');
+}
+if (!KAFKA_HOSTS) {
+    throw new Error('KAFKA_HOSTS not defined');
+}
+if (!KAFKA_TOPIC) {
+    throw new Error('KAFKA_TOPIC not defined');
+}
+
+const PRODUCER_MESSAGE_MAX_BYTES = 5000020;
```
watsaqat, line 49: are all the hardcoded values intentional?

watsaqat (approved):

requeueFailedCRRCronJob.js
jonathan-gramain: I am usually wary of reusing error types that have a specific meaning in the language for other purposes. The check we do here is kinda close to a variable existence check but it's more of a pre-validation of arguments passed to the script (via environment vars) than a real reference error, hence I prefer to use generic errors with an explicit message.

requeueFailedCRRCronJob.js
jonathan-gramain: This is because the script is not integrated into backbeat (not yet at least), and it does not pull backbeat as a dependency, so we just redefine what are the values in the BackbeatProducer class for this purpose.

PR #148, opened 2021-11-04
S3UTILS-12: add compareBuckets script

alexanderchan-scality:


compareBuckets/utils.js
```diff
@@ -0,0 +1,106 @@
...
+        queryParams.push(`maxKeys=${params.MaxKeys}`);
+    }
+    if (params.KeyMarker !== undefined) {
+        queryParams.push(`marker=${encodeURIComponent(params.KeyMarker)}`);
+    }
+    return `${baseURL}${queryParams.length > 0 ? '?' : ''}${queryParams.join('&')}`;
+}
+
+function listBucketMasterKeys(params, cb) {
+    const { bucket, marker, hostPort, maxKeys, workers } = params;
```
jonathan-gramain: Now that I think about it, the listing type can be set with `listingType: 'DelimiterMaster'` in the query params, so that version keys (those with `\0` in them) are not returned. It could make the listing more efficient and remove the need to filter out versions.

compareBuckets/utils.js
```diff
@@ -0,0 +1,106 @@
...
+        if (err) {
+            return cb(err);
+        }
+        if (res.statusCode !== 200) {
+            return cb(new Error(`GET ${url} returned status ${res.statusCode}`));
+        }
+        const resp = JSON.parse(res.body);
+        const { Contents, IsTruncated } = resp;
+
+        return async.filterLimit(Contents, workers, (item, itemDone) => {
```
jonathan-gramain: This logic could use the synchronous `Contents.filter(item => ...)`

tests/unit/compareBuckets/compareBuckets.js
```diff
@@ -0,0 +1,411 @@
...
+            expect(err).toBeNull();
+            expect(status.srcKeyMarker).toEqual('');
+            expect(status.dstKeyMarker).toEqual('3');
+            expect(status.missingInSrcCount).toEqual(3);
+            expect(status.missingInDstCount).toEqual(0);
+            return done();
+        });
+    });
+
+    it('should succesufully perform compare (single listing call)', done => {
```
jonathan-gramain:
```suggestion
    it('should successfully perform compare (single listing call)', done => {
```

compareBuckets/utils.js
```diff
@@ -0,0 +1,106 @@
...
+    if (params.MaxKeys) {
+        queryParams.push(`maxKeys=${params.MaxKeys}`);
+    }
+    if (params.KeyMarker !== undefined) {
+        queryParams.push(`marker=${encodeURIComponent(params.KeyMarker)}`);
+    }
+    return `${baseURL}${queryParams.length > 0 ? '?' : ''}${queryParams.join('&')}`;
+}
+
+function listBucketMasterKeys(params, cb) {
```
jonathan-gramain, line 70: We need to wrap this function with a retry loop, otherwise the script will stop at the first error, which is likely to happen on a large bucket

compareBuckets/index.js
```diff
@@ -0,0 +1,146 @@
...
+    node compareBuckets.js
+
+Mandatory environment variables:
+    SRC_BUCKETD_HOSTPORT: ip:port of the source bucketd endpoint
+    DST_BUCKETD_HOSTPORT: ip:port of the destination bucketd endpoint
+    SRC_BUCKET: bucket to be scanned
+    DST_BUCKET: bucket to be scanned
+
+Optional environment variables:
+    SRC_KEY_MARKER: key to continue listing from on the source bucket
```
jonathan-gramain: I think the user should only pass a single option here to resume from a point, up to the script to set its state and resume the listing from the same point on both bucketds.

jonathan-gramain (changes requested):
Only the missing retries look critical to me, the rest should be good. I can add the retries and I may also do some other changes.

compareBuckets/compareBuckets.js
```diff
@@ -0,0 +1,140 @@
...
+                        continue;
+                    }
+
+                    ++dstIdx;
+                    ++srcIdx;
+                }
+
+                srcContents = srcContents.slice(srcIdx);
+                dstContents = dstContents.slice(dstIdx);
+                return setTimeout(() => done(null), 500);
```
jonathan-gramain: Is there a reason for the `setTimeout`? I think it could make things painfully slow on a powerful platform.

compareBuckets/compareBuckets.js
```diff
@@ -0,0 +1,140 @@
...
+                    return listBucketMasterKeys(bucketdSrcParams,
+                        (err, isTruncated, marker, contents) => {
+                            if (err) {
+                                return _done(err);
+                            }
+
+                            srcContents = contents;
+                            srcDone = !isTruncated;
+                            srcMarker = marker;
+                            statusObj.srcKeyMarker = srcMarker;
```
jonathan-gramain: Here `bucketdSrcParams.marker` should be updated, fixed it to avoid looping over the same range

compareBuckets/utils.js
alexanderchan-scality: adding the `listingType` params to remove this

compareBuckets/utils.js
alexanderchan-scality: adding `async.retry` around the `httpRequest`

compareBuckets/compareBuckets.js
alexanderchan-scality: this was a precaution against making rapid requests. I can remove it if this presents itself as a bottleneck

compareBuckets/compareBuckets.js
alexanderchan-scality: thanks! didn't catch this with the mock, and the issue popped when testing against a real backend

jonathan-gramain (approved):

jonathan-gramain (approved):
LGTM

rahulreddy (approved):

PR #150, opened 2021-11-16
Improvement/s3 utils 11 repair duplicate sproxyd keys

watsaqat:


watsaqat:
fix for https://scality.atlassian.net/browse/S3UTILS-11

SproxydKeysScan/DuplicateKeysWindow.js
```diff
@@ -0,0 +1,126 @@
+const { Logger } = require('werelogs');
+const log = new Logger('s3utils:DuplicateKeysWinow');
```
jonathan-gramain:
```suggestion
const log = new Logger('s3utils:DuplicateKeysWindow');
```

SproxydKeysScan/SproxydKeysSubscribers.js
```diff
@@ -0,0 +1,66 @@
+const { MultiMap } = require('./DuplicateKeysWindow');
+const { repairObject } = require('../repairDuplicateVersionsSuite');
+const getObjectURL = require('../VerifyBucketSproxydKeys/getObjectURL');
+const { Logger } = require('werelogs');
+const log = new Logger('s3utils:DuplicateKeysIngestion');
```
jonathan-gramain:
```suggestion
const log = new Logger('s3utils:SproxydKeysSubscribers');
```

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+     * @param {function(err:Error)} cb - callback
+     * @returns {undefined}
+     * @callback called with (err:Error)
+     */
+    setBegin(cb) {
+        if (this.begin) {
+            return cb(null);
+        }
+        // fetch one record to get cseq
+        const requestUrl = `${this.url}/log?begin=${1}&limit=${1}&targetLeader=False`;
```
jonathan-gramain:
```suggestion
        const requestUrl = `${this.url}/log?begin=1&limit=1&targetLeader=False`;
```

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * If this.begin is undefined, latest cseq is fetched and this.begin is set to this.cseq - this.lookBack
+     * @param {function(err:Error)} cb - callback
+     * @returns {undefined}
+     * @callback called with (err:Error)
+     */
+    setBegin(cb) {
+        if (this.begin) {
+            return cb(null);
```
jonathan-gramain:
```suggestion
            return process.nextTick(cb);
```
Because it's safer to always call the callback in a next execution tick, and conventionally we usually write `cb()` in place of `cb(null)` when there is no data to return.

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+                log.error('unable to fetch cseq', { err });
+                return cb(err);
+            }
+            const body = JSON.parse(res.body);
+            this._setCseq(body);
+
+            // make sure begin is at least 1 since Raft Journal logs are 1-indexed
+            this.begin = Math.max(1, this.cseq - this.lookBack);
+            log.info(`initial begin: ${this.begin}`);
+            return cb(null);
```
jonathan-gramain:
```suggestion
            return cb();
```

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+     * @property {string} db - source bucket name
+     * @property {number} method - the type of operation that created the log object.
+     * @property {Array.<LogObjectEntry>} entries - Array of LogObjectEntry
+     */
+
+    /**
+     * @typedef {Object} RaftJournalBody
+     * @property {number} info.start - the offset that this request started reading from.
+     * @property {number} info.cseq - total number of logged objects in this raft session.
+     * @property {number} info.prune - ???
```
jonathan-gramain: You can leave this out

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+     * @returns {undefined}
+     * @callback called with (err:Error, body:RaftJournalBody)
+     */
+    getBatch(cb) {
+        const requestUrl = `${this.url}/log?begin=${this.begin}&limit=${this.limit}&targetLeader=False`;
+        return this._httpRequest('GET', requestUrl, null, (err, res) => {
+            if (err) {
+                return cb(err);
+            }
+            if (!res || !res.body) {
```
jonathan-gramain: I'd rather put this check after `if (res.statusCode != 200)`, since some/most error cases would not send a body back, it allows to return the status code in the error.

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+     * @callback called with (err:Error, extractedKeys:Array.<ExtractedKey>)
+     */
+    processBatch(body, cb) {
+        this._setCseq(body);
+
+        // { objectKey, sproxydKeys }
+        const extractedKeys = [];
+
+        body.log.forEach(record => {
+            if (record.method === 8) {
```
jonathan-gramain, line 159: Could add a comment saying that 8 is a batch, or maybe better to add a global constant at top of file `const DB_METHOD_BATCH = 8;` and use it here

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+                        return;
+                    }
+
+                    const bucket = record.db;
+
+                    let json = null;
+                    try {
+                        json = JSON.parse(entry.value);
+                    } catch (err) {
+                        log.error('json corrupted');
```
jonathan-gramain:
```suggestion
                        log.error('json corrupted', {
                            begin: this.begin,
                            limit: this.limit,
                        });
```

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+                        return;
+                    }
+
+                    // skip empty objects
```
jonathan-gramain, line 183: FYI, there used to be an older format where `location` was not an array, but we can ignore this here

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+     * @returns {undefined}
+     * @callback called with (err:Error, res: Boolean).
+     */
+    updateStatus(extractedKeys, cb) {
+        for (const entry of extractedKeys) {
+            try {
+                this.processor.insert(entry.objectKey, entry.sproxydKeys, entry.bucket);
+            } catch (err) {
+                log.error('insert key failed in updateStatus', { err, entry });
+                return cb(err, null);
```
jonathan-gramain:
```suggestion
                return cb(err);
```

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+                    // skip empty objects
+                    if (!json.location || !Array.isArray(json.location)) {
+                        return;
+                    }
+                    const sproxydKeys = json.location.map(loc => loc.key);
+                    extractedKeys.push({ objectKey, sproxydKeys, bucket });
+                });
+            }
+        });
+        log.info('processBatch succeeded');
```
jonathan-gramain:
```suggestion
        log.debug('processBatch succeeded', {
            begin: this.begin,
            limit: this.limit,
        });
```
This can be useful as debug information, but may be too low-level for info, ideally, in info level we log a regular status update (`logProgress`) with:
- how many total records and entries processed in that time period
- what is the current offset
- the latest cseq
- the lag (as `cseq - next_begin`)
- and also maybe the time spent processing the log (compared to idling) may be useful to diagnose potential load issues

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+                this.processor.insert(entry.objectKey, entry.sproxydKeys, entry.bucket);
+            } catch (err) {
+                log.error('insert key failed in updateStatus', { err, entry });
+                return cb(err, null);
+            }
+        }
+
+        // if we go over cseq, start at cseq + 1 while waiting for new raft journal entries
+        this.begin = Math.min(this.limit + this.begin, this.cseq + 1);
+        log.info('updateStatus succeeded');
```
jonathan-gramain:
```suggestion
        log.debug('updateStatus succeeded');
```

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+     * If an error occurs at any point in the process, or if there are no new objects to be processed
+     * a timeout of 5 seconds is set before polling again.
+     * @param {function(err:Error, res:number)} cb - callback. Response res is a timeout in milliseconds.
+     * @returns {undefined}
+     * @callback called with (err:Error, res:number)
+     */
+
+    runOnce(cb) {
+        return waterfall([
+            next => this.setBegin(
```
jonathan-gramain, line 233: This might be called just once when initializing the instance, but no big deal if it helps to simplify the flow I'm ok with it.

SproxydKeysScan/DuplicateKeysWindow.js
```diff
@@ -0,0 +1,126 @@
...
+     * @param {string} objectKey - Versioned object Key.
+     * @param {string} bucket - Bucket name.
+     * @returns {undefined}
+     */
+    checkDuplicate(sproxydKey, objectKey, bucket) {
+        const existingObjectKey = this.sproxydKeys.get(sproxydKey);
+        if (existingObjectKey && existingObjectKey !== objectKey) {
+            log.info(`existing object key found: 
+                Existing { sproxydKey: ${sproxydKey}, objectKey: ${existingObjectKey} } 
+                Current { sproxydKey: ${sproxydKey}, objectKey: ${objectKey}}`
```
jonathan-gramain: Making it slightly more parsable and standard, also trying to make the message more explicit:

```suggestion
            log.warn('object keys with a duplicate sproxyd key found', {
                objectKeys: [existingObjectKey, objectKey],
                sproxydKey,
            });
```

SproxydKeysScan/DuplicateKeysWindow.js
```diff
@@ -0,0 +1,126 @@
...
+            );
+            const params = {
+                objectKey,
+                existingObjectKey,
+                sproxydKey,
+                context: this,
+                bucket,
+            };
+            const handlers = this.subscribers.get('duplicateSproxyKeyFound');
```
jonathan-gramain:
```suggestion
            const handlers = this.subscribers.get('duplicateSproxydKeyFound');
```

SproxydKeysScan/SproxydKeysSubscribers.js
```diff
@@ -0,0 +1,66 @@
...
+                log.error('an error occurred repairing object', {
+                    objInfo,
+                    error: { message: err.message },
+                });
+            }
+        });
+    }
+}
+
+subscribers.set('duplicateSproxyKeyFound', new DuplicateSproxydKeyFoundHandler());
```
jonathan-gramain:
```suggestion
subscribers.set('duplicateSproxydKeyFound', new DuplicateSproxydKeyFoundHandler());
```

tests/unit/SproxydKeysScan/DuplicateKeysWindowTest.js
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+    describe('SproxydKeyProcessor', () => {
+        const windowSize = 10;
+
+        const setupProcessor = windowSize => {
+            const subscribers = new MultiMap();
+            const duplicateHandler = new DuplicateSproxydKeyFoundHandler();
+            duplicateHandler._repairObject = jest.fn().mockReturnValue((err, res) => [err, res]);
+            subscribers.set('duplicateSproxyKeyFound', duplicateHandler);
```
jonathan-gramain:
```suggestion
            subscribers.set('duplicateSproxydKeyFound', duplicateHandler);
```

SproxydKeysScan/DuplicateKeysWindow.js
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Checks if a duplicate Exists for each Sproxyd Key.
+     * @param {string} objectKey - Object Key.
+     * @param {Array.<string>} sproxydKeys - Array of Sproxyd Keys for associated Object key.
+     * @param {string} bucket - bucket name.
+     * @returns {undefined}
+     */
+    insert(objectKey, sproxydKeys, bucket) {
+        for (const sproxydKey of sproxydKeys) {
```
jonathan-gramain, line 120: That loop will likely lead to logging a message for each duplicate key found between the same objects, leading to multiple messages logged for a single pair of objects (since MPU would likely have multiple parts and also likely share all their keys in common). It might be better to build up a list of duplicate keys first, then if non-empty, log the warning.

Not critical though, we can save it for later for the sake of time.

SproxydKeysScan/SproxydKeysSubscribers.js
```diff
@@ -0,0 +1,66 @@
...
+     * @param {string} sparams.proxydKey - Share sproxyd key between the existing and new object key.
+     * @param {Class} params.context - Instance of SproxydKeysProcessor from which handle was called.
+     * @param {string} params.bucket - bucket name.
+     * @returns {undefined}
+     */
+    handle(params) {
+        const [objectUrl, objectUrl2] =
+            [params.objectKey, params.existingObjectKey]
+                .map(objectKey => this._getObjectURL(params.bucket, objectKey))
+                .sort(this._cmp);
```
jonathan-gramain, line 47: Might be simpler to use the default sort and reverse the assigned values:

```suggestion
        const [objectUrl2, objectUrl] =
            [params.objectKey, params.existingObjectKey]
                .map(objectKey => this._getObjectURL(params.bucket, objectKey))
                .sort();
```

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,85 @@
...
+const log = new Logger('s3utils:SproxydKeysScan:run');
+
+const env = {
+    SPROXYD_KEY_LIMIT: process.env.SPROXYD_KEY_LIMIT,
+    RAFT_SESSION_ID: process.env.RAFT_SESSION_ID,
+    LOOKBACK_WINDOW: process.env.LOOKBACK_WINDOW,
+};
+
+const USAGE = `
+runSproxydKeyScan.js
```
jonathan-gramain:
```suggestion
runSproxydKeysScan.js
```

Also, the "run" prefix looks redundant or too vague since all scripts can be "run", or maybe, call it `sproxydKeysScanDaemon` to make it explicit it is a daemon.

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,85 @@
...
+
+This script continously polls the Raft Journal of a given Raft session id.
+When a duplicate sproxyd key is found for objects with different versioned object keys, 
+a repair is done by creating new sproxyd keys.
+
+Usage:
+    node runSproxydKeyScan.js
```
jonathan-gramain:
```suggestion
    node runSproxydKeysScan.js
```

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,85 @@
...
+const { getSproxydAlias } = require('../repairDuplicateVersionsSuite');
+const { Logger } = require('werelogs');
+
+const log = new Logger('s3utils:SproxydKeysScan:run');
+
+const env = {
+    SPROXYD_KEY_LIMIT: process.env.SPROXYD_KEY_LIMIT,
+    RAFT_SESSION_ID: process.env.RAFT_SESSION_ID,
+    LOOKBACK_WINDOW: process.env.LOOKBACK_WINDOW,
+};
```
jonathan-gramain:
```suggestion
const env = {
    BUCKETD_HOSTPORT: process.env.BUCKETD_HOSTPORT,
    SPROXYD_HOSTPORT: process.env.SPROXYD_HOSTPORT,
    SPROXYD_KEY_LIMIT: process.env.SPROXYD_KEY_LIMIT,
    RAFT_SESSION_ID: process.env.RAFT_SESSION_ID,
    LOOKBACK_WINDOW: process.env.LOOKBACK_WINDOW,
};
```

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,85 @@
...
+When a duplicate sproxyd key is found for objects with different versioned object keys, 
+a repair is done by creating new sproxyd keys.
+
+Usage:
+    node runSproxydKeyScan.js
+
+Mandatory environment variables:
+    BUCKETD_HOSTPORT: <bucketd_host>:<bucketd_port>
+    SPROXYD_HOSTPORT: <sproxyd_host>:<sproxyd_port>
+    SPROXYD_KEY_LIMIT: Number of objects to fetch at each poll of the Raft Journal.
```
jonathan-gramain:
```suggestion
    RAFT_LOG_BATCH_SIZE: Number of records to fetch at each poll of the Raft Journal.
```

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,85 @@
...
+    RAFT_SESSION_ID: Session id from which to read Journal.
+    LOOKBACK_WINDOW: When the process is started/restarted it will begin at cseq - LOOKBACK_WINDOW 
+        unless SPROXYD_KEY_BEGIN is set explicitly. 
+    DUPLICATE_KEYS_WINDOW_SIZE: Max unique sproxydkeys that the Map will store.
+
+Optional environment variables:
+    SPROXYD_KEY_BEGIN: offset to begin scanning from. Leave this out if you wish to begin
```
jonathan-gramain:
```suggestion
    RAFT_LOG_BEGIN_SEQ: offset to begin scanning the raft journal from. Leave this out if you wish to begin 
```

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,279 @@
...
+                        log.error('in setBegin', { error: err.message });
+                        next(err);
+                    } else {
+                        next(null);
+                    }
+                }),
+            next => this.getBatch(
+                (err, res) => {
+                    if (err) {
+                        log.error('in getBatch', { error: err.message });
```
jonathan-gramain, line 245: You can skip logging (or debug log only) of the errors of type "416" (requested range not satisfiable) since those errors would occur as soon as the log has been processed entirely without any new entry to process.

jonathan-gramain (commented):
Looks good to me overall, just some minor comments

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -63,7 +63,7 @@ class RaftJournalReader {
      */
     setBegin(cb) {
         if (this.begin) {
-            return process.nextTick(cb);
+            return cb();
```
jonathan-gramain: Is this revert intentional?

jonathan-gramain (approved):

SproxydKeysScan/DuplicateKeysIngestion.js
watsaqat: ~~jest tests hung with process.nextTick(cb). I think process.nextTick was waiting for the unit tests to resolve. I didn't investigate further, seemed returning cb() was ok. @jonathan-gramain~~

tmacro (approved):

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,87 @@
+const { RaftJournalReader } = require('./DuplicateKeysIngestion');
```
jonathan-gramain: By renaming this file `index.js`, or by declaring it as the entrypoint in a `package.json` file, the script can be run directly as `node SproxydKeysScan`, it can be more convenient.

repairDuplicateVersionsSuite.js
```diff
@@ -0,0 +1,425 @@
...
+    verifyBucketSproxydKeys.js s3utils script. This script only
+    processes the log entries containing the message "duplicate
+    sproxyd key found" and ignores other entries.
+
+Mandatory environment variables:
+    BUCKETD_HOSTPORT: ip:port of bucketd endpoint
+    SPROXYD_HOSTPORT: ip:port of sproxyd endpoint
+`;
+
+if (!BUCKETD_HOSTPORT) {
```
jonathan-gramain: In the suite file, everything that was run at startup for the original `repairDuplicateVersions` script should be removed, otherwise it conflicts with running the main script (seen during testing)

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,87 @@
...
+    DUPLICATE_KEYS_WINDOW_SIZE: Max unique sproxydkeys that the Map will store.
+
+Optional environment variables:
+    RAFT_LOG_BEGIN_SEQ: offset to begin scanning from. Leave this out if you wish to begin 
+    at latest cseq - LOOKBACK_WINDOW
+`;
+for (const [key, value] of Object.entries(env)) {
+    if (!value) {
+        log.info(`${key} must be defined`);
+        log.info(USAGE);
```
jonathan-gramain: logging the USAGE string as such results in a JSON log line dumped with the serialized contents of that string that is meant to be dumped on the console directly, it would be better to use `console.log`/`console.error` instead, at least for the validation stage. (seen during testing)

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,87 @@
+const { RaftJournalReader } = require('./DuplicateKeysIngestion');
+const { getSproxydAlias } = require('../repairDuplicateVersionsSuite');
+const { Logger } = require('werelogs');
+
+const log = new Logger('s3utils:SproxydKeysScan:run');
+
+const env = {
+    BUCKETD_HOSTPORT: process.env.BUCKETD_HOSTPORT,
+    SPROXYD_HOSTPORT: process.env.SPROXYD_HOSTPORT,
+    RAFT_LOG_BATCH_SIZE: process.env.RAFT_LOG_BATCH_SIZE,
```
jonathan-gramain: I recommend setting a default value for this variable (1000 should be a good default) as the user may not know what is a good value to set.

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,87 @@
...
+    RAFT_SESSION_ID: process.env.RAFT_SESSION_ID,
+    LOOKBACK_WINDOW: process.env.LOOKBACK_WINDOW,
```
jonathan-gramain: This one also should have a default value, we can put 10000 for the default, we can revisit it later if we needed.

SproxydKeysScan/DuplicateKeysIngestion.js
jonathan-gramain: You seem to have removed the special handling of the 416 error, unwanted change?

SproxydKeysScan/DuplicateKeysIngestion.js
```diff
@@ -0,0 +1,298 @@
...
+ */
+class RaftJournalReader {
+    /**
+     *
+     * @param {(number|undefined)} begin - The first cseq number to read from. Automatically set if undefined.
+     * @param {number} limit - How many Raft Journal entries to fetch with each poll.
+     * @param {number} sessionId - Raft Session Id that the reader will poll from.
+     */
+    constructor(begin, limit, sessionId) {
+        this.lookBack = LOOKBACK_WINDOW;
```
jonathan-gramain: not strictly needed, but it would be cleaner to convert to an integer here

SproxydKeysScan/runSproxydKeysScan.js
```diff
@@ -0,0 +1,87 @@
...
+ * @returns {undefined}
+ */
+function runJournalReader() {
+    if (env.RAFT_LOG_BEGIN_SEQ === undefined) {
+        log.info('RAFT_LOG_BEGIN_SEQ is not defined. Ingestion will start at latest cseq - LOOKBACK_WINDOW');
+    }
+    const reader = new RaftJournalReader(
+        env.RAFT_LOG_BEGIN_SEQ,
+        env.RAFT_LOG_BATCH_SIZE,
+        env.RAFT_SESSION_ID
```
jonathan-gramain:
```suggestion
        Number.parseInt(env.RAFT_LOG_BEGIN_SEQ, 10),
        Number.parseInt(env.RAFT_LOG_BATCH_SIZE, 10),
        Number.parseInt(env.RAFT_SESSION_ID, 10)
```

Required to avoid adding string+string = concatenation later on in the class code, notably when doing `begin + limit` (and the class constructor shall accept only integers anyway).

SproxydKeysScan/SproxydKeysSubscribers.js
jonathan-gramain: Tests show that apparently, when there are more than 2 versions with duplicate keys, only the oldest version is repaired, leaving the other intermediate version(s) untouched and duplicates remain with the most recent master version.

I believe it's due to the algorithm used that is not correct for more than 2 keys, I think we can come up with a fixed version keeping the general principle, by conditionally updating the object key cache with a new version instead of never updating it.

PR #153, opened 2021-11-24
S3 utils 21 insert lower key version after repair

watsaqat:


SproxydKeysScan/SproxydKeysSubscribers.js
```diff
@@ -34,17 +34,23 @@ class DuplicateSproxydKeyFoundHandler {
...
-     * @param {string} sparams.proxydKey - Share sproxyd key between the existing and new object key.
+     * @param {string} params.existingObjectKey - Object Key with the same sproxyd Key that was already in Map.
+     * @param {string} params.sproxydKey - Share sproxyd key between the existing and new object key.
      * @param {Class} params.context - Instance of SproxydKeysProcessor from which handle was called.
      * @param {string} params.bucket - bucket name.
      * @returns {undefined}
      */
     handle(params) {
-        const [objectUrl, objectUrl2] =
+        const [olderVersionKey, newerVersionKey] =
```
jonathan-gramain: Can be simplified:

```suggestion
const [newerVersionKey, olderVersionKey] = [params.objectKey, params.existingObjectKey].sort()
```

SproxydKeysScan/SproxydKeysSubscribers.js
```diff
@@ -34,17 +34,23 @@ class DuplicateSproxydKeyFoundHandler {
      * Takes two object keys with an erroneously shared sproxyd key and repairs the older version of the two.
      * @param {Object} params - input params.
      * @param {string} params.objectKey - Object Key which was attempted to be inserted into the Map.
-     * @param {string} params.existingBucketMD - Object Key with the same sproxyd Key that was already in Map.
-     * @param {string} sparams.proxydKey - Share sproxyd key between the existing and new object key.
+     * @param {string} params.existingObjectKey - Object Key with the same sproxyd Key that was already in Map.
+     * @param {string} params.sproxydKey - Share sproxyd key between the existing and new object key.
```
jonathan-gramain:
```suggestion
     * @param {string} params.sproxydKey - Shared sproxyd key between the existing and new object key.
```

tests/unit/SproxydKeysScan/SproxydKeysSubscribersTest.js
```diff
@@ -0,0 +1,35 @@
...
+const range = require('lodash/range');
+
+describe('SproxydKeysSubscribers', () => {
+    describe('DuplicateSproxydKeyFoundHandler', () => {
+        const [processor, duplicateHandler] = setupProcessor(10);
+
+        const key = 'test-2021-11-24T19-41-08-494Z/2\u000098362217123979999997RG001 ';
+        const version = () => `${randomize('0', 3)}.${randomize('0', 3)}`;
+
+        test('the lower (newer) version is always repaired and then set in the map', () => {
```
jonathan-gramain: Maybe only a test description issue: we need to repair the *higher (older)* version, not the lower (newer), because we don't want to repair the master version since it would require updating the master key as well and we ignore the master keys.

What we set in the map is indeed the lower (newer) version though, I think that is correct in the code and what should be done indeed to achieve the expected result (repairing all versions except the master/newer)

jonathan-gramain (approved):

tmacro (approved):

PR #156, opened 2021-11-30
add statistics logger and log levels.

watsaqat:


watsaqat:
The approach here is extending werelogs Logger class to include summary statistics at a tunable interval.

For example, the common usage of Logger.info is
```log.info(`success happened for object ${x}`);```

the normal Logger.info api is modified to include an `exampleSuccess` event:
```log.info({ event: 'exampleSuccess' message: `success happened for object ${x}`});```
the message is handled as it normally would be by werelogs, while the event will be incremented in a counter. 

After the tunable interval time has passed we log.info something like:
```
{
   exampleSuccess: 42,
   exampleFailure: 0
}
```

File: SproxydKeysScan/Logging.js [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+
+class StatisticsLogger extends werelogs.Logger {
+    /**
+     * @constructor
+     * @param {number} logInterval - log summary statistics once every intervalSize seconds
+     */
+    constructor(logInterval) {
+        super();
+        this.interval = logInterval;
+        this.beginTime = new Date(Date.now());
```
miniscruff:
```suggestion
        this.beginTime = new Date();
```
This should do the same thing.

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+class StatisticsLogger extends werelogs.Logger {
+    /**
+     * @constructor
+     * @param {number} logInterval - log summary statistics once every intervalSize seconds
+     */
+    constructor(logInterval) {
+        super();
+        this.interval = logInterval;
```
miniscruff, line 22: I feel like a default here would be nice, maybe just one second?

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,57 @@
...
+        this.beginTime = new Date(Date.now());
+        this.statistics = new Counter();
+    }
+
+    // { event, message }
```
miniscruff: I would use a JSDoc typedef

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,57 @@
...
+    info(params) {
+        this.statistics[params.event] += 1;
+        super().info(params.message);
+    }
+
+    logInterval() {
+        const currentTime = new Date(Date.now());
+        log.info(`Summary for ${this.beginTime} to ${currentTime}:`);
+        log.info(`${this.statistics}`);
```
miniscruff:
```suggestion
        log.info(this.statistics);
```
I believe this is the same

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,57 @@
...
+
+const loggerConfig = {
+    logLevel: env.LOG_LEVEL,
+    dumpLevel: env.DUMP_LEVEL,
+};
+
+werelogs.configure(loggerConfig);
+const log = new werelogs.Logger('s3utils:SproxydKeysScan:StatisticsLogger');
+
+class Counter {
```
miniscruff: I feel like a Map would be fine, not sure what a custom class does here.

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    logInterval() {
+        const currentTime = new Date(Date.now());
+        log.info(`Summary for ${this.beginTime} to ${currentTime}:`);
+        log.info(`${this.statistics}`);
+
+        // setup for next interval
+        this.beginTime = currentTime;
+        this.statistics = new Counter();
```
miniscruff: If statistics was a map this could just be `this.statistics.clear()`

miniscruff (commented):
Small comments

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,57 @@
...
+
+    // { event, message }
+    info(params) {
+        this.statistics[params.event] += 1;
+        super().info(params.message);
+    }
+
+    logInterval() {
+        const currentTime = new Date(Date.now());
+        log.info(`Summary for ${this.beginTime} to ${currentTime}:`);
```
jonathan-gramain: Ideally those two log lines should be consolidated in a single line.

I recommend embedding `beginTime` and `currentTime` into log fields, as the second argument of `log.info`.

SproxydKeysScan/Logging.js
jonathan-gramain:
```suggestion
        log.info('status update', this.statistics);
```

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,57 @@
...
+     */
+    constructor(logInterval) {
+        super();
+        this.interval = logInterval;
+        this.beginTime = new Date(Date.now());
+        this.statistics = new Counter();
+    }
+
+    // { event, message }
+    info(params) {
```
jonathan-gramain: I would be wary of changing the prototype of the subclass function compared to the parent class, it can come with surprises.

Instead of subclassing `werelogs.Logger`, why not making a helper function or class that calls the regular logger class.

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,57 @@
...
+
+        // setup for next interval
+        this.beginTime = currentTime;
+        this.statistics = new Counter();
+    }
+
+    run() {
+        const context = this;
+        context.logInterval();
+        setTimeout(() => context.run(), context.interval * 1000);
```
jonathan-gramain: Using `setInterval` instead of `setTimeout` could avoid the need for "recursion" (even if it's in another tick) and simplify a bit.

watsaqat:
The design I had yesterday had a few glaring shortcomings:

1. there was a 1-1 mapping between StatisticsLogger class and a werelogs.Logger and
2. The api differed from werelogs.Logger

To address 1, a ProxyLogger was created that has the same functionality and usage as werelogs.Logger. Multiple ProxyLoggers can send aggregate events to the singleton AggregateLogger (renamed from StatisticsLogger).
For 2, the `eventMessage` (renamed from `event`) was moved into `data`. 

Now to change any werelogs.Logger instance to a ProxyLogger, it just needs to be required instead of werelogs.Logger. If no `eventMessage` is in the `data` object as in `log.info(message, data)`, the behavior stays the same as in werelogs.Logger. When passed in, aggregate statistics is logged on an interval.

watsaqat:
another idea (and likely a better one) is to keep all messages in the logs generic (any variables go into the data object), and then use Prometheus for aggregation and monitoring.

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    update(event) {
+        this.statistics.set(event, (this.statistics.get(event) || 0) + 1);
+    }
+
+    logInterval() {
+        const currentTime = new Date();
+        const summary = {};
+        for (const [k, v] of new Map(this.statistics).entries()) {
```
miniscruff:
```suggestion
        for (const [k, v] of this.statistics) {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#iterating_map_with_for..of

jonathan-gramain (approved):
If I may give my humble opinion, I don't see a reason to have this complexity, all we shall need in the end is a log line at regular intervals with some statistics about what's going on. Logging itself is already taken care of by werelogs, hence all we have to do is an interval timer and a call to `log.info` (which `log` object may come from an already initialized logger) to dump a few values, and reset those values after the log is complete.

I may not have the big picture of how it works (more jsdoc may help to understand here) but my feeling is that we kind of overengineered this for the need, anyway, if it works as it should, let's keep it as we want to have this merged timely.

SproxydKeysScan/index.js
```diff
@@ -27,6 +36,9 @@ Mandatory environment variables:
 Optional environment variables:
     RAFT_LOG_BEGIN_SEQ: offset to begin scanning from. Leave this out if you wish to begin 
     at latest cseq - LOOKBACK_WINDOW
+    LOG_LEVEL: defaults to info
+    DUMP_LEVEL: defaults to error
+    LOG_INTERVAL: logs summary of events in the given interval - defaults to 5 minutes
```
miniscruff, line 41: where does this get defaulted to 5 minutes?

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,110 @@
...
+            summary,
+        };
+        this.log.info('Summary update:', data);
+
+        // setup for next interval
+        this.beginTime = currentTime;
+        this.statistics.clear();
+    }
+
+    run() {
```
miniscruff, line 59: I feel like calling run in the constructor also works, I can't see a reason to create a "timed logger" and not start the timer right away?

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,110 @@
+const werelogs = require('werelogs');
+const { env } = require('./env');
+
+/**
+ * @class
+ * @classdesc - aggregates events in a map and logs them at an interval.
+ * call .run() after instantiation to being loop.
+ */
+class AggregateLogger {
```
miniscruff, line 10: I feel like these classes belong more in werelogs than in this repo.

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * @constructor
+     * @param {number} logInterval - log summary statistics once every intervalSize seconds
+     */
+    constructor(logInterval) {
+        this.log = new werelogs.Logger('s3utils:SproxydKeysScan:AggregateLogger');
+        if (AggregateLogger._instance) {
```
miniscruff, line 17: Why is this a singleton?

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,110 @@
...
+    * @constructor
+    * @param {class} logger - an instance of werelogs.Logger
+    * @returns {class} - a proxy to the logger
+    */
+    constructor(logger) {
+        this.logger = logger;
+        this.aggregateLogger = new AggregateLogger(env.LOG_INTERVAL);
+        const context = this;
+        const logLevels = new Set(['trace', 'debug', 'info', 'warn', 'error', 'fatal']);
+        const handlers = {
```
miniscruff, line 78: This whole proxy and reflect seems a bit complex for wrapping the werelogs logger, also seems a bit fragile with the "is function" check and log level check.

tests/unit/SproxydKeysScan/LoggingTest.js
```diff
@@ -0,0 +1,85 @@
+const { AggregateLogger, ProxyLoggerCreator } = require('../../../SproxydKeysScan/Logging');
+const { Logger } = require('werelogs');
+describe.only('SproxydKeysScan:Logging', () => {
```
miniscruff:
```suggestion
describe('SproxydKeysScan:Logging', () => {
```

miniscruff (changes requested):

SproxydKeysScan/Logging.js
watsaqat: to explicitly enforce one to many w.r.t. aggregatorLogger to ProxyLoggers. When we make a new proxy wereloggers.Logger instance we have to check if an existing AggregateLogger exists and this was one way of doing so

SproxydKeysScan/Logging.js
watsaqat: I forgot to this to the doc string, but this approach allows usage with both an interval or with event driven logging by calling logInterval explicitly

SproxydKeysScan/index.js
watsaqat: env.js

SproxydKeysScan/Logging.js
watsaqat: I updated the doc string to explain. If you don't call it right away, this gives you flexibility to either log at an interval or log in response to an event by calling logInterval explicitly.

SproxydKeysScan/Logging.js
watsaqat: agree, but I thought it's better to validate the design and usage for this one specific task for some time before moving it into werelogs for general use.

SproxydKeysScan/Logging.js
watsaqat: This enforces 1 to many w.r.t. AggregateLogger to proxy loggers. So, when a new a proxy logger is created, this is a way to enforce it uses the existing instance and we don't get disjoint summary logs.

SproxydKeysScan/Logging.js
watsaqat: I originally was thinking of just subclassing werelogs.Logger but some concerns were raised so I used a proxy pattern to not touch the underlying class. The checks are just saying "only intercept function calls with the names in logLevels". From what I've seen, its pretty standard usage.

miniscruff: Personally I have never seen or used Proxy or Reflect here, I would just manually create all the methods but I guess its fine. Seems a little strange to pass in a logger to just override its methods instead of creating one but thats fine I guess.

SproxydKeysScan/Logging.js
```diff
@@ -0,0 +1,111 @@
...
+            get(target, prop) {
+                if (typeof target[prop] === 'function') {
```
miniscruff: Should you add the `if logLevels.has(props)` check here so can do both checks at once. And not inside the apply method.

SproxydKeysScan/Logging.js
miniscruff: Oh I see what is going on, I was expecting the use case to be:
```js
const agg = new AggregateLogger(...);
const aLogger = new Logger(... whatever logger );
agg.Add(aLogger);
```

or something similar, the way it is now is sort of like, you create both logggers and the aggregate and they just kinda "link" together magically. Kind of confusing to me.

miniscruff (approved):

PR #168, opened 2021-12-08
improvement: S3UTILS-32 option ONLY_DELETED to cleanupNoncurrentVersions

jonathan-gramain:
Add an environment option ONLY_DELETED to the
cleanupNoncurrentVersions script, so that it only deletes noncurrent
versions if the object's current version is a delete marker. It also
then deletes the delete marker as usual.

Also updated the README.md to reflect the new option and fixed a few
missing updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,7 +504,7 @@ node cleanupNoncurrentVersions.js bucket1[,bucket2...]
 
 ## Mandatory environment variables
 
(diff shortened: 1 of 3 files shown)


jonathan-gramain:
ping

watsaqat (approved):

miniscruff (approved):

PR #188, opened 2022-01-27
ft: S3C-5190 continuous integrity check

vrancurel:
Continuous integrity check based on the existing `compareBuckets` tool.

The tool maintains 2 sets of differences, and follow the oplog to update the sets in real-time.

It persists the state in ZK/RING to avoid rescanning the bucket each time, in using the snapshot/scan/oplog pattern.

### DIFF
--- a/compareBuckets/ArrayUtils.js
+++ b/compareBuckets/ArrayUtils.js
@@ -1,0 +1,88 @@
+function indexOf(arr, value) {
+    if (!arr.length) {
+        return -1;
+    }
+    let lo = 0;
+    let hi = arr.length - 1;
+
+    while (hi - lo > 1) {
+        const i = lo + ((hi - lo) >> 1);
+        if (arr[i] > value) {
+            hi = i;
+        } else {
+            lo = i;
+        }
+    }
+    if (arr[lo] === value) {
+        return lo;
+    }
+    if (arr[hi] === value) {
+        return hi;
+    }
+    return -1;
+}
+
+function indexAtOrBelow(arr, value) {
+    let i;
+    let lo;
+    let hi;
+
+    if (!arr.length || arr[0] > value) {
+        return -1;
+    }
+    if (arr[arr.length - 1] <= value) {
+        return arr.length - 1;
+    }
+
+    lo = 0;
+    hi = arr.length - 1;
+
+    while (hi - lo > 1) {
+        i = lo + ((hi - lo) >> 1);
+        if (arr[i] > value) {
+            hi = i;
+        } else {
+            lo = i;
+        }
+    }
+
+    return lo;
+}
+
+/*
+ * perform symmetric diff in O(m + n)
+ */
+function
+symDiff(k1, k2, v1, v2, cb) {
+    let i = 0;
+    let j = 0;
+    const n = k1.length;
+    const m = k2.length;
+
+    while (i < n && j < m) {
+        if (k1[i] < k2[j]) {
+            cb(v1[i]);
+            i++;
+        } else if (k2[j] < k1[i]) {
+            cb(v2[j]);
+            j++;
+        } else {
+            i++;
+            j++;
+        }
+    }
+    while (i < n) {
+        cb(v1[i]);
+        i++;
+    }
+    while (j < m) {
+        cb(v2[j]);
+        j++;
+    }
+}
+
+module.exports = {
+    indexOf,
+    indexAtOrBelow,
+    symDiff,
+};
--- a/compareBuckets/BucketdOplogInterface.js
+++ b/compareBuckets/BucketdOplogInterface.js
@@ -1,0 +1,222 @@
+/*
+ * Main interface for bucketd oplog management
+ *
+ * persist is an interface with the following methods:
+ * - constructor(params)
+ * - load(bucketName, persistData, cb(err, offset))
+ * - save(bucketName, persistData, offset, cb(err))
+
+ * persistData is an interface with the following methods:
+ * - constuctor(params)
+ * - initState(cb(err)): initialize the structure, e.g. initial bucket scan
+ * - loadState(stream, cb(err)): load the state
+ * - saveState(stream, cb(err)): save the state
+ */
+const async = require('async');
+const BucketClient = require('bucketclient').RESTClient;
+const { jsutil } = require('arsenal');
+const LogConsumer = require('arsenal').storage.metadata.bucketclient.LogConsumer;
+const { isMasterKey } = require('arsenal/lib/versioning/Version');
+const Flusher = require('./Flusher');
+const werelogs = require('werelogs');
+
+werelogs.configure({
+    level: 'info',
+    dump: 'error',
+});
+
+class BucketdOplogInterface {
+
+    constructor(params) {
+        let bkBootstrap = ['localhost:9000'];
+        if (params && params.bootstrap) {
+            bkBootstrap = params.bootstrap;
+        }
+        this.bkClient = new BucketClient(bkBootstrap);
+        this.backendRetryTimes = 3;
+        this.backendRetryInterval = 300;
+        this.bucketdOplogQuerySize = 20;
+        this.raftId = 2; // XXX use route
+        this.logger = new werelogs.Logger('BucketdOplogInterface');
+    }
+
+    start(bucketName, persist, persistData, flusherParams, cb) {
+        async.waterfall([
+            next => {
+                let cseq = undefined;
+                // get stored offset if we have it
+                // load persistData if present
+                persist.load(bucketName, persistData, (err, offset) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    cseq = offset;
+                    return next(null, cseq, persistData);
+                });
+            },
+            (cseq, persistData, next) => {
+                if (cseq !== undefined) {
+                    this.logger.info(`skipping cseq acquisition (cseq=${cseq})`,
+                                     { bucketName });
+                    return next(null, cseq, persistData, true);
+                }
+                this.logger.info('cseq acquisition',
+                                 { bucketName });
+                async.retry(
+                    {
+                        times: this.backendRetryTimes,
+                        interval: this.backendRetryInterval,
+                    },
+                    done => {
+                        this.bkClient.getRaftLog(
+                            this.raftId,
+                            1,
+                            1,
+                            true,
+                            null,
+                            (err, stream) => {
+                                if (err) {
+                                    this.logger.info('retrying getRaftLog', { err, bucketName });
+                                    return done(err);
+                                }
+                                const chunks = [];
+                                stream.on('data', chunk => {
+                                    chunks.push(chunk);
+                                });
+                                stream.on('end', () => {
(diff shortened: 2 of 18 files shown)


vrancurel:
This is a WIP for now as some generic code (e.g. `SortedSet`) can be moved in Arsenal.

compareBuckets/BucketdOplogInterface.js
```diff
@@ -0,0 +1,222 @@
...
+    constructor(params) {
+        let bkBootstrap = ['localhost:9000'];
+        if (params && params.bootstrap) {
+            bkBootstrap = params.bootstrap;
+        }
+        this.bkClient = new BucketClient(bkBootstrap);
+        this.backendRetryTimes = 3;
+        this.backendRetryInterval = 300;
+        this.bucketdOplogQuerySize = 20;
+        this.raftId = 2; // XXX use route
```
vrancurel, line 39: I will add a `getBucketRaftId()` method in `bucketclient`.

compareBuckets/PersistRingInterface.js
```diff
@@ -0,0 +1,244 @@
...
+    dump: 'error',
+});
+
+class PersistRingInterface {
+
+    constructor() {
+        const zkConnectionString = 'localhost:2181';
+        const zkPath = '/persist-ring-interface';
+        const spBootstrap = ['localhost:8181'];
+        const spPath = '/proxy/DC1/'; // do not forget "/" at the end !!!
```
vrancurel, line 22: Those will be passed in env / params.

compareBuckets/compareBuckets.js
```diff
@@ -221,7 +233,254 @@ function compareBuckets(params, log, cb) {
...
+        /*
+          bucketdSrcParams,
+          bucketdDstParams,
+          statusObj,
+          verbose,
+          compareVersionId,
+          compareObjectSize,
+        */
+        this.log = log;
+        this.srcPersist = new PersistFileInterface();
```
vrancurel, line 332: For prod it will be using `PersistRingInterface()`

vrancurel:
Code complete: once overall code approved I will dispatch the generic code in Arsenal and do proper PRs with proper project names.

tests/unit/compareBuckets/Injector.js
```diff
@@ -0,0 +1,165 @@
...
+        let result = '';
+        const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+        const charactersLen = characters.length;
+        for (let i = 0; i < len; i++) {
+            result += characters.charAt(
+                Math.floor(
+                    this.rnd.next() *
+                        charactersLen));
+        }
+        return result;
```
miniscruff, line 41: prando seems to have a next string method, maybe something like:
```suggestion
        const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
        return this.rnd.nextString(len, characters);
```

tests/unit/compareBuckets/Injector.js
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    genBase16(len) {
```
miniscruff, line 44: same as above

tests/unit/compareBuckets/Injector.js
```diff
@@ -0,0 +1,165 @@
...
+        for (let i = 0; i < len; i++) {
+            result += characters.charAt(
+                Math.floor(
+                    this.rnd.next() *
+                        charactersLen));
+        }
+        return result;
+    }
+
+    _zeroPad(s, n, width) {
```
miniscruff, line 57:
```js
return s + n.toString().padStart(width, '0');
```
You can use padStart here I believe

tests/unit/compareBuckets/Injector.js
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    inject(bucketName, numKeys, maxSeq, op, randomSeq, prefix, suffix, keys, values, cb) {
+        const injectParams = {};
+        injectParams.numKeys = numKeys;
+        injectParams.maxSeq = maxSeq;
+        injectParams.op = op;
+        injectParams.randomSeq = randomSeq;
+        injectParams.prefix = prefix;
+        injectParams.suffix = suffix;
```
miniscruff, line 70: any reason to use the pattern of creating an empty object and filling it, over filling it when creating it? I have seen it a few times and not sure if there is a reason.

tests/unit/compareBuckets/Injector.js
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    injectExt(bucketName, params, inputKeys, outputKeys, outputValues, cb) {
+        async.timesLimit(
+            params.numKeys,
+            10,
+            (n, next) => {
+                let key;
+                if (inputKeys) {
+                    const idx = Math.floor(this.rnd.next() * inputKeys.length);
```
miniscruff, line 87: prando has a nextInt method: https://www.npmjs.com/package/prando

compareBuckets/compareBuckets.js
```diff
@@ -221,7 +233,254 @@ function compareBuckets(params, log, cb) {
...
+            // console.log('found', item.key, value, item.value);
+            assert(value.mismatch === diffMgr.MismatchNoExist);
+            diffMgr.state[source].del(item.key);
+            updateMissingCount(diffMgr, source, -1);
+        } else {
+            // console.log('not found', item.key, item.value);
+            const _value = {};
+            Object.assign(_value, item.value);
+            _value.mismatch = diffMgr.MismatchNoExist;
+            diffMgr.state[target].set(item.key, _value);
```
miniscruff, line 265:
```suggestion
            const _value = {};
            Object.assign(_value, item.value);
            _value.mismatch = diffMgr.MismatchNoExist;
            const _value = Object.assign({}, item.value, { mismatch: dffMgr.MismatchNoExist });
            diffMgr.state[target].set(item.key, _value);
```

Or using spread:
```js
diffMgr.state[target].set(item.key, { ...item.value, mismatch: diffMgr.MismatchNoExist });
```

compareBuckets/compareBuckets.js
```diff
@@ -221,7 +233,254 @@ function compareBuckets(params, log, cb) {
...
+    assert(source === diffMgr.IdxSrc ||
+           source === diffMgr.IdxDst);
+    assert(target === diffMgr.IdxSrc ||
+           target === diffMgr.IdxDst);
+    addQueue.forEach(item => {
+        // console.log('update add', item);
+        const value = diffMgr.state[source].get(item.key);
+        if (value !== undefined) {
+            // console.log('found', item.key, value, item.value);
+            assert(value.mismatch === diffMgr.MismatchNoExist);
```
miniscruff, line 257: I would use assert.strictEqual instead of assert x === y, but also I think we want to not use asserts outside of tests.

compareBuckets/PersistRingInterface.js
```diff
@@ -0,0 +1,244 @@
...
+                return cb(err);
+            }
+            return cb(null, _data.offset);
+        });
+    }
+
+    // cb(err)
+    save(bucketName, persistData, offset, cb) {
+        this.logger.info(`saving ${bucketName} offset ${offset}`);
+        async.waterfall([
```
miniscruff, line 115: instead of waterfall I think we should switch to using async/await if possible. Not sure how much work switching between the two are but something to consider.

compareBuckets/PersistRingInterface.js
```diff
@@ -0,0 +1,244 @@
...
+                    this.getZKPath(bucketName),
+                    (err, stat) => {
+                        if (err) {
+                            this.logger.error(`exists ${bucketName} error`, { err });
+                            return next(err);
+                        }
+                        let doesExist = false;
+                        if (stat) {
+                            doesExist = true;
+                        }
```
miniscruff, line 175: I think you can just do `!!doesExist` or something similar based on what stat is exactly, like maybe `stat !== undefined`.

compareBuckets/PersistMemInterface.js
```diff
@@ -0,0 +1,75 @@
...
+        this.memoryStreams = {};
+        this.offsets = {};
+        this.logger = new werelogs.Logger('PersistMemInterface');
+    }
+
+    getOffset(bucketName) {
+        if (!this.offsets[bucketName]) {
+            return {};
+        }
+        return this.offsets[bucketName];
```
miniscruff, line 23:
```suggestion
        return this.offsets[bucketName] || {};
```

compareBuckets/MongoOplogInterface.js
```diff
@@ -0,0 +1,156 @@
...
+
+const databaseName = 'metadata';
+
+class MongoOplogInterface {
+
+    constructor() {
+        this.logger = new werelogs.Logger('MongoOplogInterface');
+    }
+
+    start(bucketName, persist, persistData, stopAt, cb) {
```
miniscruff, line 30: I really think we should split this function up and as stated before maybe switch to async/await. If not splitting use `function funcName` syntax over the anonymous array functions to provide context.

compareBuckets/Flusher.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    addEvent(item, offset, opType) {
+        // console.log('addEvent', this.bucketName, item, opType);
+        if (opType === 'put') {
+            this.addedQueue.push(item);
+        } else if (opType === 'del') {
+            this.deletedQueue.push(item);
+        } else {
+            assert(false);
```
miniscruff, line 51: I would add a message to this assert at least, debugging would be very hard otherwise.

compareBuckets/BucketdOplogInterface.js
```diff
@@ -0,0 +1,222 @@
...
+        this.logger = new werelogs.Logger('BucketdOplogInterface');
+    }
+
+    start(bucketName, persist, persistData, flusherParams, cb) {
```
miniscruff, line 43: Same as the other function, this is very long and has not that much context or docs.

compareBuckets/BucketdOplogInterface.js
```diff
@@ -0,0 +1,222 @@
+/*
+ * Main interface for bucketd oplog management
```
miniscruff, line 2: Could maybe use jsdocs interface https://jsdoc.app/tags-interface.html or similar

compareBuckets/ArrayUtils.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    return lo;
+}
+
+/*
+ * perform symmetric diff in O(m + n)
+ */
+function
+symDiff(k1, k2, v1, v2, cb) {
```
miniscruff, line 56:
```suggestion
function symDiff(k1, k2, v1, v2, cb) {
```

miniscruff (changes requested):

tests/unit/compareBuckets/Injector.js
vrancurel: Yes, did not see it. Prando is really cool because it can be seeded, so make hard bugs to find easily reproducible.

PR #192, opened 2022-03-15
CountItems: fix connection to sharded mongodb

francoisferrand:
Try to read mongodb connection parameters from config file, to ensure
all mongodb client parameters are passed as needed.

Issue: S3UTILS-49

### DIFF
--- a/CountItems/utils/index.js
+++ b/CountItems/utils/index.js
@@ -23,19 +23,34 @@ function getIsLocationTransientCb(log, locationConfigFile) {
     };
 }
 
+function getMongoDbConfig(log, configFile) {
+    let config;
+    try {
+        const data = fs.readFileSync(configFile, { encoding: 'utf-8' });
+        config = JSON.parse(data).mongodb;
+    } catch (err) {
+        log.info(`could not parse config file: ${err.message}`);
+        config = {
+            database: 'metadata',
+            replicaSet: 'rs0',
+            replicationGroupId: 'RG001',
+            shardCollections: false,
+        };
+    }
+    return config;
+}
+
 function createMongoParams(log, customParams) {
-    const replicaSetHosts = process.env.MONGODB_REPLICASET;
-    const database = process.env.MONGODB_DATABASE || 'metadata';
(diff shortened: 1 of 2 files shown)


CountItems/utils/index.js
```diff
@@ -23,19 +23,34 @@ function getIsLocationTransientCb(log, locationConfigFile) {
     };
 }
 
+function getMongoDbConfig(log, configFile) {
+    let config;
+    try {
+        const data = fs.readFileSync(configFile, { encoding: 'utf-8' });
+        config = JSON.parse(data).mongodb;
+    } catch (err) {
+        log.info(`could not parse config file: ${err.message}`);
```
ghost, line 32: Maybe `warn` here? As using a config file should be the preferred way of running this function (according to your code).

ghost (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "s3utils",
-  "version": "1.12.0",
+  "version": "1.11.11",
```
lavigne958: ?

lavigne958 (changes requested):

CountItems/utils/index.js
lavigne958: Even `error`, you can't parse your config file you should fail. in my opinion.

package.json
ghost: The 1.11 branch has a wrong version in package.json (I also saw that :laughing: )
https://github.com/scality/s3utils/blob/development/1.11/package.json#L3

CountItems/utils/index.js
francoisferrand: It should not be the preferred solution at the moment, since it is not always used : e.g. it is not used in S3C.
For compatibility reason we should not warn or fail here.

package.json
lavigne958: what do we do about it then ?

francoisferrand: as discussed on slack, we should just forget 1.11, and use 1.12 in Zenko 2.2

lavigne958 (approved):

PR #215, opened 2022-06-10
feature: S3UTILS-55 block digests stream

jonathan-gramain:
BlockDigestsStream is a stream.Transform stream that groups series of
consecutive input items into blocks, and outputs each block's
attributes as a stream of objects.

The stream converts a potentially large list of items into a much
smaller list of block hashes. The output can be used in place of the
input for quick comparison purpose, by only comparing the block hashes
of two listings, where comparing their individual entries could be
costly in time and space. When differences are spotted between block
entries, one can then list only the differing block sub-range to
detect which individual entries differ.

### DIFF
--- a/CompareRaftMembers/BlockDigestsStream.js
+++ b/CompareRaftMembers/BlockDigestsStream.js
@@ -1,0 +1,120 @@
+const crypto = require('crypto');
+const stream = require('stream');
+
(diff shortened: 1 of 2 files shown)


jonathan-gramain:
ping

tmacro (approved):

watsaqat (approved):

PR #219, opened 2022-06-15
S3UTILS-58 diff stream without digest checks

jonathan-gramain:
Output differences between the input stream, consisting of
{ key: string, value: string } objects sorted by key, and a bucketd
interface that can be queried via HTTP for ranges of keys from
given buckets.

Input items are meant to be coming from a raw leveldb database
listing, transformed via a DBListStream object to be comparable with
bucketd listing.

### DIFF
--- a/CompareRaftMembers/DiffStream.js
+++ b/CompareRaftMembers/DiffStream.js
@@ -1,0 +1,223 @@
+const stream = require('stream');
+
+const BucketStream = require('./BucketStream');
+
+/**
+ * Output differences between the input stream, consisting of
+ * { key: string, value: string } objects sorted by key, and a bucketd
+ * interface that can be queried via HTTP for ranges of keys from
+ * given buckets.
+ *
+ * Input items are meant to be coming from a raw leveldb database
+ * listing, transformed via a DBListStream object to be comparable with
+ * bucketd listing.
+ *
+ * @class DiffStream
+ */
+class DiffStream extends stream.Transform {
+    /**
+     * @constructor
+     * @param {object} params - constructor params
+     * @param {string} params.bucketdHost - bucketd host name or IP address
+     * @param {number} params.bucketdPort - bucketd API port
+     * @param {number} [params.maxBufferSize=1000] - maximum number of
+     * items to bufferize
+     * @param {class} [params.BucketStreamClass=BucketStream] -
+     * override the BucketStream class, used for unit tests only
+     */
+    constructor(params) {
+        super({ objectMode: true });
+        const {
+            bucketdHost, bucketdPort,
+            maxBufferSize,
+            BucketStreamClass,
+        } = params;
+
+        this.bucketdHost = bucketdHost;
+        this.bucketdPort = bucketdPort;
+        this.maxBufferSize = maxBufferSize || 1000;
+        this.BucketStreamClass = BucketStreamClass || BucketStream;
+
+        this.inputBuffer = [];
+        this.inputBufferBucket = null;
+        this.currentMarker = null;
+    }
+
+    _transform(item, encoding, callback) {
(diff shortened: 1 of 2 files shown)


CompareRaftMembers/DiffStream.js
```diff
@@ -0,0 +1,173 @@
...
+ */
+class DiffStream extends stream.Transform {
+    /**
+     * @constructor
+     * @param {object} params - constructor params
+     * @param {string} params.bucketdHost - bucketd host name or IP address
+     * @param {number} params.bucketdPort - bucketd API port
+     * @param {number} [params.maxBufferSize=1000] - maximum number of
+     * items to bufferize
+     * @param {class} [params.BucketStreamClass=BucketStream] -
```
watsaqat, line 25: why is this needed in the params instead of modifying BucketStreamClass after instantiation in the tests?

tests/unit/CompareRaftMembers/DiffStream.js
```diff
@@ -0,0 +1,556 @@
...
+                                marker: null,
+                                lastKey: null,
+                            },
+                        ],
+                    },
+                ],
+            },
+            {
+                desc: 'with 7777 identical entries in db and bucketd',
+                dbContents: () => {
```
watsaqat: you can use getters https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get so you don't have to use explicit checks such as
`if (typeof dbContents === 'function') { .. } else { .. }`

tests/unit/CompareRaftMembers/DiffStream.js
```diff
@@ -0,0 +1,556 @@
...
+                        .on('data', data => {
+                            output.push(data);
+                        })
+                        .on('end', () => {
+                            expect(output).toEqual(expectedOutput);
+                            expect(MOCK_BUCKET_STREAM_REQUESTS_MADE)
+                                .toEqual(withDigests.expectedBucketStreamRequests);
+                            done();
+                        })
+                        .on('error', done);
```
watsaqat, line 533: if an error occurs before the 'end' event, wouldn't the test appear to succeed here?

CompareRaftMembers/DiffStream.js
```diff
@@ -0,0 +1,173 @@
...
+        if (this.inputBuffer.length > 0) {
+            this._compareInputBufferWithBucketd(null, callback);
+        } else {
+            callback();
+        }
+    }
+
+    _parseItem(item) {
+        const { key: fullKey, value } = item;
+        const slashIndex = fullKey.indexOf('/');
```
watsaqat: want to make sure that bucketName can't be suffixed with '/' for some reason as that would mess up this split.

watsaqat:
could you add some JSDocs to the DiffStream methods?

CompareRaftMembers/DiffStream.js
jonathan-gramain: yes, bucket names are DNS-compatible, hence no slash or other forbidden character

tests/unit/CompareRaftMembers/DiffStream.js
jonathan-gramain: no, since the error event will be passed an error argument, it will automatically be passed to the `done` callback, then failing the test.

CompareRaftMembers/DiffStream.js
jonathan-gramain: It's because the instanciation happens internally, potentially multiple times, from the `_compareInputBufferWithBucketd` helper.

tmacro (approved):

watsaqat (approved):

jonathan-gramain:
@watsaqat I added JSDoc, I also reorganized the code a bit and squashed

PR #223, opened 2022-06-20
Remove & upgrade dependencies from base image

francoisferrand:
- switch to the 'slim' variant, to avoid importing useless packages
- use multi-stage build to avoid including build tool in production image
- use `yarn install` to install packages

NOTE: this does not fix all issues, we would need to switch to different distribution for this (see #225 or #224), which would need better assessment of impact.

Issue: S3UTILS-69

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-name: Release
+name: release
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


miniscruff (approved):

alexanderchan-scality (approved):

alexanderchan-scality:
@francoisferrand it seems like it would also be beneficial to use `yarn install` and perform clean up after.
something like:
```
RUN apt-get autoremove --purge -y python git build-essential \
    && rm -rf /var/lib/apt/lists/* 
```

alexanderchan-scality (approved):

francoisferrand:
also fixes S3UTILS-72

PR #230, opened 2022-06-23
Switch to debian bullseye base image

francoisferrand:
- Upgrade to debian bullseye baseline
- Use build arg to specify node version once
- Switch to go-lang supervisor

Issue: S3UTILS-73

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Use separate builder to retrieve & build node modules
-FROM node:16.15.1-slim AS builder
+FROM node:16.15.1-bullseye-slim AS builder
 
 WORKDIR /usr/src/app
 
@@ -15,7 +15,7 @@ COPY ./package.json ./yarn.lock ./
 RUN yarn install --production
 
 ################################################################################
-FROM node:16.15.1-slim
+FROM node:16.15.1-bullseye-slim
 
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,35 +1,38 @@
...
     apt-get install -y --no-install-recommends \
         build-essential \
+        ca-certificates \
         git \
         jq \
         python3 \
-        ssh
+        ssh \
+        wget \
+    && rm -rf /var/lib/apt/lists/*
```
jbertran, line 17: We can `apt clean` instead - it's possible to to a dry run of that command to check if we're cleaning enough.

jbertran (commented):
Re: supervisord, have you tried the documented docker-based approach (https://github.com/ochinchina/supervisord#usage-from-a-docker-container)? It seems less brittle and less error-prone than the .tar way.

Dockerfile
francoisferrand: according to https://docs.docker.com/develop/develop-images/dockerfile_best-practices/, `apt clean` is supposed to be automatically called... but the exemple does the `rm -rf /var/lib/apt/lists/*` (just like it was before this change, see later in the diff)

so i choose to stick to the old & proven approach 😀

francoisferrand:
> Re: supervisord, have you tried the documented docker-based approach ([ochinchina/supervisord#usage-from-a-docker-container](https://github.com/ochinchina/supervisord#usage-from-a-docker-container))? It seems less brittle and less error-prone than the .tar way.

actually, the images are not tagged, and not even up-to-date.

jbertran:
:man_facepalming:  LGTM then!

jbertran (approved):

jonathan-gramain (approved):

PR #232, opened 2022-06-24
feature: S3UTILS-56 ignore special buckets

jonathan-gramain:
Ignore "users..bucket" and "mpuShadowBucketXXX" so they don't appear
in the output listing, because they are not visible from bucketd
listing.

### DIFF
--- a/CompareRaftMembers/DBListStream.js
+++ b/CompareRaftMembers/DBListStream.js
@@ -49,6 +49,11 @@ class DBListStream extends stream.Transform {
             }
             [bucket, objectKey] = [key.slice(0, slashIndex), key.slice(slashIndex + 1)];
         }
+        // ignore special buckets, that cannot be listed from bucketd
+        // like regular buckets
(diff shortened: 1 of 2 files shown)


CompareRaftMembers/DBListStream.js
```diff
@@ -49,6 +49,11 @@ class DBListStream extends stream.Transform {
             }
             [bucket, objectKey] = [key.slice(0, slashIndex), key.slice(slashIndex + 1)];
         }
+        // ignore special buckets, that cannot be listed from bucketd
+        // like regular buckets
+        if (bucket === 'users..bucket' || bucket.startsWith('mpuShadowBucket')) {
```
watsaqat, line 54: want to confirm there's not an extra '.' in `users..bucket`

watsaqat (approved):

CompareRaftMembers/DBListStream.js
jonathan-gramain: Yes the name contains two dots

tmacro (approved):

PR #235, opened 2022-06-24
bugfix: S3UTILS-75 fetch missing object locations

jonathan-gramain:
BucketStream class now supports fetching object locations that are
missing before sending the object entry on the stream output.

### DIFF
--- a/CompareRaftMembers/BucketStream.js
+++ b/CompareRaftMembers/BucketStream.js
@@ -17,17 +17,24 @@ class BucketStream extends stream.Readable {
             bucketName,
             marker,
             lastKey,
-            retryDelayMs,
-            maxRetryDelayMs,
+            retryDelayMs: userRetryDelayMs,
+            maxRetryDelayMs: userMaxRetryDelayMs,
         } = params;
         this.bucketdHost = bucketdHost;
         this.bucketdPort = bucketdPort;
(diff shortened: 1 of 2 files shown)


jonathan-gramain:
ping

tmacro (approved):

CompareRaftMembers/BucketStream.js
```diff
@@ -116,26 +116,71 @@ class BucketStream extends stream.Readable {
...
+            (err, response) => {
+                if (err) {
+                    return callback(err);
                 }
-                if (IsTruncated && !ended) {
-                    this.marker = Contents[Contents.length - 1].key;
-                } else {
-                    this.push(null);
+                if (!response) {
+                    // the object was not found
```
watsaqat, line 180: maybe adding a log here is useful?

watsaqat (approved):

CompareRaftMembers/BucketStream.js
jonathan-gramain: I think it's not too useful to log this, as it can be a normal situation depending on circumstances. Moreover, it would require carrying a logger in this case, which is possible, but I'd rather have it decoupled from the logging and emit error events that can be logged by the main script when errors occur.

PR #236, opened 2022-06-29
improvement: S3UTILS-79 document followerDiff scan procedure

jonathan-gramain:
In README.md, provide a procedure to run a full scan to detect
metadata discrepancies across active stateful members.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,13 +425,13 @@ DIFF_OUTPUT_FILE environment variable.
 In this file, it outputs each key that differs as line-separated JSON
 entries, where each entry can be one of:
 
-- [{ key, value }, null]: this key is present on this follower but not
-  on the leader
+- `[{ key, value }, null]`: this key is present on this follower but
+  not on the leader
 
-- [null, { key, value }]: this key is not present on this follower but
-  is present on the leader
+- `[null, { key, value }]`: this key is not present on this follower
+  but is present on the leader
 
-- [{ key, value: "{value1}" }, { key, value: "{value2}" }]: this key
+- `[{ key, value: "{value1}" }, { key, value: "{value2}" }]`: this key
   has a different value between this follower and the leader: "value1"
   is the value seen on the follower and "value2" the value seen on the
   leader.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -465,6 +465,217 @@ node followerDiff.js
 
 * **PARALLEL_SCANS**: number of databases to scan in parallel (default 4)
 
+## Scan Procedure
+
+In order to scan a set of raft sessions to detect discrepancies
```
watsaqat, line 470: iirc, the use-case is to detect leader-follower discrepancies, correct? If so, lets mention that

README.md
```diff
@@ -465,6 +465,217 @@ node followerDiff.js
...
+```
+ssh -n 'root'@'storage-1' 'echo 3 4 5 7 > /tmp/rs-to-scan'
+ssh -n 'root'@'storage-2' 'echo 1 2 3 5 6 7 8 > /tmp/rs-to-scan'
+ssh -n 'root'@'storage-3' 'echo 1 2 3 4 5 6 8 > /tmp/rs-to-scan'
+ssh -n 'root'@'storage-4' 'echo 1 2 3 4 5 6 7 8 > /tmp/rs-to-scan'
+ssh -n 'root'@'storage-5' 'echo 1 2 4 6 7 8 > /tmp/rs-to-scan'
+```
+
+Check that what you get looks like the above, then execute those
+commands on the shell (e.g. by piping with `| bash -x`).
```
watsaqat: maybe adding the full command to be run is better here

README.md
```diff
@@ -465,6 +465,217 @@ node followerDiff.js
...
+  range of keys multiple times outputting to the same digests
+  database, the risk being to render that range of digests much less
+  efficient to optimize the comparisons, or even useless. If an update
+  of the digests database is wanted, it's best to re-create a new one
+  from scratch (e.g. by deleting the old digests directory first).
+
+- The current version of the script does not work with Metadata bucket
+  format v1, only with v0. If the need arises, it could be made to
+  work with v1 with some more work.
+
```
watsaqat, line 736: a few Qs:
* if any of the steps fail, should the user start over from step 1, 2, etc?
* will the user always know when a step has failed? could failures be silent at any of the steps or will timeouts alert the user?

README.md
jonathan-gramain: I'd say that the *goal* is to detect any discrepancy between members, and the *means* is checking follower-leader discrepancies, due to how technically this is achievable without introducing downtime.

README.md
jonathan-gramain: Agree, I can add the command to run

README.md
jonathan-gramain: There could be ways to re-do a single step that fails, either by just relaunching it or by doing a first cleanup before re-launching the step. I should give some details about those cases indeed.

I'd need to make sure the provided commands and scripts fail in a clear way (like checking the error messages or exit code).

This also makes me think that even before running the set of steps, there are pre-conditions to be checked, like making sure followers are up-to-date with the leader for example.

README.md
```diff
@@ -465,6 +465,217 @@ node followerDiff.js
...
+```
+
+**If there are more active stateful hosts to scan, continue with Step 4
+on the next active stateful host.**
+
+### Step 8: Gather results
+
+Finally, once all scans have been done on all active stateful hosts,
+gather all diff results for later analysis.
+
```
tmacro, line 699: Could include a bit about what to expect in the results. Mine are empty files which I assume means that there are no differences. No need to get too technical just something high level about the different cases.

jonathan-gramain: I can add a pointer to where the output format is described above in the `followerDiff` tool description. Indeed empty files mean no difference (one difference per line)

README.md
jonathan-gramain: I'll add an extra paragraph to explain the general idea of the procedure.

watsaqat (approved):

tmacro (approved):

PR #238, opened 2022-07-11
Make sure crrExistingObjects does not break OOB object

francoisferrand:
The code used to re-create the replication state, which could break some
fields (esp. dataLocation) or remove some replication backends.

We now simply update the status, which is safer (more future-proof) and
simpler.

In addition, the code will now avoid updating the internal version id (not
needed), or properly setup the "null version" flags if instructed to do this
"upgrade" of the object.

Issue: S3UTILS-77

### DIFF
--- a/crrExistingObjects.js
+++ b/crrExistingObjects.js
@@ -27,6 +27,7 @@ const MAX_SCANNED = (process.env.MAX_SCANNED
     && Number.parseInt(process.env.MAX_SCANNED, 10));
 let { KEY_MARKER } = process.env;
 let { VERSION_ID_MARKER } = process.env;
+const { GENERATE_INTERNAL_VERSION_ID } = process.env;
 
 const LISTING_LIMIT = (process.env.LISTING_LIMIT
     && Number.parseInt(process.env.LISTING_LIMIT, 10)) || 1000;
@@ -121,6 +122,13 @@ function _markObjectPending(
                 // the listed version is "null", the object may have
                 // an actual internal versionId, only if the bucket
                 // was versioning-suspended when the object was put.
+                return next();
+            }
+            if (!GENERATE_INTERNAL_VERSION_ID) {
+                // When the GENERATE_INTERNAL_VERSION_ID env variable is set,
+                // matching objects with no *internal* versionId will get
+                // "updated" to get an internal versionId. The external versionId
+                // will still be "null".
                 return next();
             }
             // The object does not have an *internal* versionId, as it
--- a/crrExistingObjects.js
+++ b/crrExistingObjects.js
@@ -149,26 +149,9 @@ function _markObjectPending(
             if (skip) {
                 return next();
             }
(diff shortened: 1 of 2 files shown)


nicolas2bert (approved):

welansari (approved):

crrExistingObjects.js
```diff
@@ -149,26 +149,9 @@ function _markObjectPending(
...
-                backends,
-                content: ops,
-                destination,
-                storageClass,
-                role: Role,
-                storageType: STORAGE_TYPE,
-            };
-            objMD.setReplicationInfo(replicationInfo);
+
+            objMD.setReplicationSiteStatus(storageClass, TARGET_REPLICATION_STATUS);
```
jonathan-gramain: I think we want to set `'PENDING'` here, not `TARGET_REPLICATION_STATUS`: the target is what objects are targeted for the update, not what status we want to put in metadata.

Also, I don't know if that would work in all cases: for example, new objects don't have any replication status in their metadata, and it's an important use case of the script to be able to replicate new objects without any replication config attached to their bucket when they were created.

jonathan-gramain (changes requested):

jonathan-gramain (approved):

jbertran (approved):

welansari:
@francoisferrand tested in ARTESCA 1.4.2 on a bucket that has both OOB and replication enabled, and that contain objects with a FAILED replication status.
The S3C bucket used for OOB contains zero byte objects, non MPU normal objects and MPU objects. All objects have been tested in their versioned and non-versioned format.
The command used is the following:
```
sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf run -i --tty --rm s3utils --image=registry.scality.com/s3utils-dev/s3utils:6ea6808c548f55fa7dffdfa2140fe76080efc664 --env="ACCESS_KEY=<ACCESS_KEY>" \
--env "SECRET_KEY=<SECRET_KEY>" \
--env "ENDPOINT=artesca-data-connector-s3api.zenko.svc.cluster.local:8000" \
--env "MONGODB_REPLICASET=data-db-mongodb-sharded-mongos-0.data-db-mongodb-sharded.zenko.svc:27017" \
--env "MONGODB_AUTH_USERNAME=<MONGO_USERNAME>" \
--env "MONGODB_AUTH_PASSWORD=<MONGO_PASSWORD>" \
--env "MONGODB_SHARD_COLLECTIONS=true" \
--env "MONGODB_DATABASE=<MONGO_DB>" \
--env "REPLICATION_GROUP_ID=RG001" \
--env "STORAGE_TYPE=aws_s3" \
--env "TARGET_REPLICATION_STATUS=FAILED" \
--env "SITE_NAME=aws-walid-test" \
--restart=Never -n zenko -- node crrExistingObjects.js another-bucket
```
Replication restarted for all of the failed objects, the objects got replicated successfully and are still accessible.

PR #244, opened 2022-07-20
improvement: S3UTILS-89 RaftOplogStream

jonathan-gramain:
Create a stream class to fetch a raft session's oplog from a specified
sequence until the tail, then keep tailing it for new records.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,7 @@
     "plugins": ["jest"],
     "env": {
         "es6": true,
+        "jasmine": true,
         "jest/globals": true
     },
     "parserOptions": {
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -3,6 +3,7 @@
     "plugins": ["jest"],
     "env": {
         "es6": true,
+        "jasmine": true,
```
jonathan-gramain, line 6: Note for reviewers: this is for the linter to recognize the `fail()` syntax and not complain that it's undefined.

watsaqat (approved):

tmacro (approved):

PR #246, opened 2022-07-22
improvement: S3UTILS-86 helper to filter recently updated keys

jonathan-gramain:
Create a helper class DiffStreamOplogFilter that tails all raft
sessions from the specified cseq offset in the background, while
ingesting diff entries. It then only forwards those which do not
appear in their respective raft session's oplog to the stream's
output.

### DIFF
--- a/CompareRaftMembers/DiffStreamOplogFilter.js
+++ b/CompareRaftMembers/DiffStreamOplogFilter.js
@@ -1,0 +1,240 @@
+const async = require('async');
+const http = require('http');
+const stream = require('stream');
+
+const RaftOplogStream = require('./RaftOplogStream');
+
+const DEFAULT_MAX_BUFFERED_ENTRIES = 10000;
+const DEFAULT_RETRY_DELAY_MS = 1000;
+const DEFAULT_MAX_RETRY_DELAY_MS = 10000;
+
+const httpAgent = new http.Agent({
+    keepAlive: true,
+});
+
+class DiffStreamOplogFilter extends stream.Transform {
+    constructor(params) {
+        super({ objectMode: true });
+        const {
+            bucketdHost,
+            bucketdPort,
+            maxBufferedEntries: userMaxBufferedEntries,
+            retryDelayMs: userRetryDelayMs,
+            maxRetryDelayMs: userMaxRetryDelayMs,
+        } = params;
+        this.bucketdHost = bucketdHost;
+        this.bucketdPort = bucketdPort;
+        this.maxBufferedEntries = userMaxBufferedEntries || DEFAULT_MAX_BUFFERED_ENTRIES;
+        const retryDelayMs = userRetryDelayMs || DEFAULT_RETRY_DELAY_MS;
+        const maxRetryDelayMs = userMaxRetryDelayMs || DEFAULT_MAX_RETRY_DELAY_MS;
+        this.retryParams = {
+            times: 20,
+            interval: retryCount => Math.min(
+                // the first retry comes as "retryCount=2", hence substract 2
+                retryDelayMs * (2 ** (retryCount - 2)),
+                maxRetryDelayMs,
+            ),
+        };
+        this.bucketNameToRaftSessionId = {};
+        this._initRaftOplogStreams(params);
+
+        this.nBufferedEntries = 0;
+        this.pendingTransformCallback = null;
+        this.pendingFlushCallback = null;
+    }
+
+    _initRaftOplogStreams(params) {
+        const {
+            bucketdHost, bucketdPort,
+            excludeFromCseqs,
+        } = params;
+
+        this.raftSessionStates = {};
+        for (const rsId of Object.keys(excludeFromCseqs)) {
+            const raftSessionState = {};
+            this.raftSessionStates[rsId] = raftSessionState;
+            const oplogStream = new RaftOplogStream({
+                bucketdHost,
+                bucketdPort,
+                raftSessionId: rsId,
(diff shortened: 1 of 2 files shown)


tmacro (approved):

CompareRaftMembers/DiffStreamOplogFilter.js
```diff
@@ -0,0 +1,240 @@
...
+
+        this.raftSessionStates = {};
+        for (const rsId of Object.keys(excludeFromCseqs)) {
+            const raftSessionState = {};
+            this.raftSessionStates[rsId] = raftSessionState;
+            const oplogStream = new RaftOplogStream({
+                bucketdHost,
+                bucketdPort,
+                raftSessionId: rsId,
+                startSeq: excludeFromCseqs[rsId] + 1,
```
watsaqat, line 95: a comment here about why startSeq is defined as such would be helpful.

CompareRaftMembers/DiffStreamOplogFilter.js
```diff
@@ -0,0 +1,240 @@
...
+    _initRaftOplogStreams(params) {
+        const {
+            bucketdHost, bucketdPort,
+            excludeFromCseqs,
+        } = params;
+
+        this.raftSessionStates = {};
+        for (const rsId of Object.keys(excludeFromCseqs)) {
+            const raftSessionState = {};
+            this.raftSessionStates[rsId] = raftSessionState;
```
watsaqat: this might be better at the bottom of the block after raftSessionState is initialized with oplogStream, oplogKeys, inputBuffer

CompareRaftMembers/DiffStreamOplogFilter.js
```diff
@@ -0,0 +1,240 @@
...
+
+    _transform(diffEntry, encoding, callback) {
+        const bucketName = this._extractBucketName(diffEntry);
+        async.waterfall([
+            next => {
+                const rsId = this.bucketNameToRaftSessionId[bucketName];
+                if (rsId !== undefined) {
+                    return next(null, rsId);
+                }
+                return this._fetchBucketRaftSessionId(bucketName, next);
```
watsaqat: I would just move the behavior of lines 97-100 inside of _fetchBucketRaftSessionId

CompareRaftMembers/DiffStreamOplogFilter.js
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    _flush(callback) {
+        if (this.nBufferedEntries === 0) {
+            // end the output stream if there are no entries to forward
+            this.push(null);
+            return callback();
+        }
+        this.pendingFlushCallback = callback;
+        return undefined;
```
watsaqat, line 168: why the explicit undefined return?

CompareRaftMembers/DiffStreamOplogFilter.js
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    _destroy() {
+        for (const rsState of Object.values(this.raftSessionStates)) {
+            rsState.oplogStream.destroy();
+        }
+    }
+
+    /**
```
watsaqat, line 177: I'd probably just leave this jsDoc out since the rest of the file doesn't use them (unless you want to add them elsewhere for consistency and describe diffEntry shape)

CompareRaftMembers/DiffStreamOplogFilter.js
```diff
@@ -0,0 +1,240 @@
...
+            res.on('data', chunk => chunks.push(chunk));
+            res.once('end', () => {
+                if (res.statusCode === 404) {
+                    return cb();
+                }
+                const body = chunks.join('');
+                if (res.statusCode !== 200) {
+                    return cb(new Error(`GET ${reqParams.path} returned status ${res.statusCode}`));
+                }
+                return cb(null, { statusCode: res.statusCode, body });
```
watsaqat: statusCode should always be 200 here since we return non-200s above, right? I think just 'statusCode: 200' is easier to scan through

watsaqat (approved):

CompareRaftMembers/DiffStreamOplogFilter.js
jonathan-gramain: one of our linter rules is `consistent-return`, which is making sure that if we return somewhere (even `undefined`, here at https://github.com/scality/s3utils/pull/246/files#diff-8275bd7edd92cb9b8c115839e21be4e42f8d5e708502eff30fcf85566bd6793dR137), we have to return something everywhere in the function.

jonathan-gramain: We abuse the `return callback()` notation though, since we don't care about the returned value.

CompareRaftMembers/DiffStreamOplogFilter.js
jonathan-gramain: I think it's fine to leave this JSDoc since it can help understanding the purpose of this helper function. I actually think the `DiffStreamOplogFilter` class could benefit from a high-level JSDoc as well, to explain its concept and its input/output, which is aligned to what you're saying.

CompareRaftMembers/DiffStreamOplogFilter.js
jonathan-gramain: indeed, will change it

CompareRaftMembers/DiffStreamOplogFilter.js
jonathan-gramain: done

PR #247, opened 2022-07-22
improvement: S3UTILS-90 update followerDiff with oplog filtering

jonathan-gramain:
Update the CompareRaftMembers/followerDiff script to accept a new
environment variable EXCLUDE_FROM_CSEQS: if passed with the current
cseq mapping of the raft sessions being scanned, it filters the diff
entries matching a key present in the corresponding bucket's raft
session oplog more recently than the given cseq.

The goal is to filter false positives due to the leader being ahead of
the follower being scanned (because the follower repds are stopped,
they cannot receive the newer updates).

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,14 +465,28 @@ node followerDiff.js
 
 * **PARALLEL_SCANS**: number of databases to scan in parallel (default 4)
 
-* **EXCLUDE_FROM_CSEQS**: cseq values for a set of raft sessions, to
-  filter out diff entries which keys appear in the corresponding raft
-  journal later than the corresponding cseq value. The value must be
-  in the following JSON format:
+* **EXCLUDE_FROM_CSEQS**: mapping of raft sessions to filter on, where
+  keys are raft session IDs and values are the cseq value for that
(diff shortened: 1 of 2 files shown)


tmacro (approved):

README.md
```diff
@@ -461,10 +461,19 @@ node followerDiff.js
 ## Optional environment variables
 
 * **LISTING_DIGESTS_INPUT_DIR**: read listing digests from the
-    specified LevelDB database
+  specified LevelDB database
 
 * **PARALLEL_SCANS**: number of databases to scan in parallel (default 4)
 
+* **EXCLUDE_FROM_CSEQS**: cseq values for a set of raft sessions, to
```
watsaqat: can this sentence be rewritten in a clearer way? Maybe an example of what gets excluded could help too

jonathan-gramain: done, let me know if it's clearer (I pretty much copied what I added in https://github.com/scality/s3utils/pull/246)

jonathan-gramain: Rather than an example, I can tell in a sentence what would happen with the given example value.

watsaqat (approved):

PR #255, opened 2022-08-05
impr(S3UTILS-99): Add tool to generate UtapiV2 service level report

tmacro:


README.md
```diff
@@ -1226,3 +1226,31 @@ script outputs this status on a single line.
...
+
+## Flags
+
+```
+  -c CONFIG, --config CONFIG
+                        Specify an alternate config file (default: /conf/config.json)
+  -r MAX_RETRIES, --max-retries MAX_RETRIES
+                        Max retries before failing a request to an external service (default: 2)
+  -p PARALLEL_QUERIES, --parallel-queries PARALLEL_QUERIES
+                        Max number of parallel queries to and warp 10 (default: 5)
```
jonathan-gramain:
```suggestion
                        Max number of parallel queries to warp 10 (default: 5)
```

README.md
```diff
@@ -1226,3 +1226,31 @@ script outputs this status on a single line.
...
+  -j, --json            Output raw reports in json format (default: False)
+  -o OUTPUT, --output OUTPUT
+                        Write report to this directory (default: Current Directory)
+  --debug               Enable debug level logging (default: False)
+  --dry-run             Don't do any computation. Only validate and print the configuration. (default: False)
```
jonathan-gramain, line 1255: I'd think that usually dry-run is understood as doing exactly the intended processing but disabling any side-effect (any write/delete). Here we could call this option like `--validate-config`

jonathan-gramain (approved):

README.md
tmacro: I see your point, but I'd rather stick with --dry-run here as its more common and we can draw on that shared context

alexanderchan-scality (approved):

francoisferrand:
This PR is an issue regarding CVEs !
The image was updated a few months back to remove python from the image, as it introduces CVEs which are hard to fix (python not being upgraded/fixed quickly enough in Debian images) ; and this just puts it back in, and brings back a few CVEs.
@tmacro is there really no other way? Could we not just use node, which is already present in the image? Or use another image, there is no point using the same image if it is actually a completely different "stack" ?

tmacro:
I'm OK with it being split out into a different image, most of its usage seems to be outside the container anyway. This is just how we decided to deliver it at the time.

PR #258, opened 2022-08-31
imp: S3UTILS-101 improve stalled retry query

naren-rajendran:
Remove regex filter on id and handle master key check on the client.

### DIFF
--- a/stalledRetry.yaml
+++ b/stalledRetry.yaml
@@ -17,11 +17,11 @@ spec:
           containers:
           - name: utils
             image: zenko/s3utils:0.2
-            imagePullPolicy: IfNotPresent 
+            imagePullPolicy: IfNotPresent
             command: ["/bin/bash"]
             args:
             - -c
-            - node stalled.js 
+            - node stalled.js
             env:
             - name: ENDPOINT
               value: ""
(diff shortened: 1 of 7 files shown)


stalledRetry.yaml
```diff
@@ -53,3 +53,7 @@ spec:
               # path to write heap snapshots
             - name: HEAP_PROFILER_PATH
               value: "/path/to/write"
+              # https://www.mongodb.com/docs/v4.2/core/read-preference/index.html
+              # mongodb read preference - primary|secondary|primaryPreferred|secondaryPreferred|nearest
+            - name: MONGODB_READ_PREFERENCE
+              value: "primary"
```
rahulreddy: Is there any reason you are targeting Primary?

rahulreddy (changes requested):

stalledRetry.yaml
naren-rajendran: It was already forced to use `primary` https://github.com/scality/s3utils/pull/258/files#diff-9b354bcb233267c1a41c1bd8dd4cc82ddef2ae458fed06e07697237a420e98e4L85 that is why I wanted to have this as a variable.

naren-rajendran: This is only with stalled (not sure why), count items uses `secondaryPreferred`.

rahulreddy: Ok unless writes are needed you should go with secondary preferred

naren-rajendran: Modified.

rahulreddy (approved):

miniscruff (approved):

PR #260, opened 2022-09-27
bugfix: S3UTILS-105 fix bucket stream hanging condition

jonathan-gramain:
When all entries listed by BucketStream class (followerDiff helper) are ignored, BucketStream needs to trigger the next listing manually because it did not call `this.push()` at all and the nodejs streams engine would not call `_read` again in such case.


jonathan-gramain:
ping

dkilcy-scality (approved):

tmacro (approved):

watsaqat (approved):

PR #265, opened 2022-10-05
bugfix: S3UTILS-110 pass VersionId to PutMetadata route

jonathan-gramain:
I reverted the crrExistingObjects script to its version 1.4.1 compatible with S3C, then cherry-picked what seems like relevant changes that were applied afterwards on the MongoDB-specific version on top.

On top of this, I then added the actual fix for S3C-6244


crrExistingObjects.js
```diff
@@ -272,9 +322,7 @@ function triggerCRROnBucket(bucketName, cb) {
                     log.error('error listing object versions', { error: err });
                     return done(err);
                 }
-                const versions = data.DeleteMarkers
-                    ? data.Versions.concat(data.DeleteMarkers) : data.Versions;
-                return _markPending(bucket, versions, err => {
+                return _markPending(bucket, data.Versions.concat(data.DeleteMarkers), err => {
```
francoisferrand, line 325: we should not concat if `data.DeleteMarkers` is null?
e.g. keep the older code or `data.Versions.concat(data.DeleteMarkers || [])` ?

crrExistingObjects.js
```diff
@@ -1,9 +1,14 @@
+const http = require('http');
+
+const AWS = require('aws-sdk');
 const {
-    doWhilst, eachSeries, eachLimit, waterfall, series,
+    doWhilst, eachSeries, eachLimit, waterfall,
 } = require('async');
-const werelogs = require('werelogs');
+
 const { ObjectMD } = require('arsenal').models;
```
francoisferrand, line 8: this imports arsenal 8 models, which may not be suitable for S3C ?

crrExistingObjects.js
jonathan-gramain: I believe this extra check was introduced in the Zenko version to cope with the change in API (using metadataUtil listing instead of S3), it should not be necessary and 1.4.1 does not have this check, as the SDK always returns an empty array if there is no delete marker.

crrExistingObjects.js
jonathan-gramain: In theory they should be compatible, but a few fields have been added in 8 (I see `content-language` for example which does not look to be optional but filled with an empty string by default). While they should not cause trouble in practice I agree that we should be careful and it may be better to avoid using model 8 on S3C.

PR #269, opened 2022-11-29
ft(S3UTILS-116): Add service level metrics sidecar for utapi v2

tmacro:


README.md
```diff
@@ -1254,3 +1254,177 @@ docker run --net=host --entrypoint python3 -v /scality/ssd01/s3/scality-utapi/co
...
+
+```shell
+docker run -d \
+  --network=host \
+  -e SIDECAR_API_KEY=dev_key_change_me \
+  -e SIDECAR_SCALE_FACTOR=1.4 \
+  -e SIDECAR_WARP10_NODE="md1-cluster1:4802@127.0.0.1:4802" \
+  scality/s3utils service-level-sidecar/index.js
+
+curl -X POST -H "Authorization: Bearer dev_key_change_me" localhost:24742/api/report | jq
```
jonathan-gramain, line 1355: On my OS (ubuntu 20.04) curl support `--oauth2-bearer` so we can do this:

```suggestion
curl -X POST --oauth2-bearer dev_key_change_me localhost:24742/api/report | jq
```

but to keep compatibility with older versions it may be better to provide the `-H` explicitly.

README.md
```diff
@@ -1254,3 +1254,177 @@ docker run --net=host --entrypoint python3 -v /scality/ssd01/s3/scality-utapi/co
...
+  scality/s3utils service-level-sidecar/index.js
+```
+
+API requests to `/api/report` require this API token to be present in the `Authorization` header.
+
+```shell
+curl -X POST -H "Authorization: Bearer dev_key_change_me" localhost:24742/api/report
+```
+
+### Scale Factor
```
jonathan-gramain, line 1428: What is the use case for this? It may be worth mentioning it in the README as it doesn't look obvious.

README.md
```diff
@@ -1254,3 +1254,177 @@ docker run --net=host --entrypoint python3 -v /scality/ssd01/s3/scality-utapi/co
...
+A CA can optionally be set using `SIDECAR_TLS_CA_PATH`.
+Certificates should be mounted into the container using a docker volume.
+
+```shell
+docker run -d \
+  --network=host \
+  -v $PWD/certs:/certs \
+  -e SIDECAR_TLS_CERT_PATH=/certs/cert.pem \
+  -e SIDECAR_TLS_KEY_PATH=/certs/key.pem \
+  -e SIDECAR_TLS_CA_PATH=/certs/key.pem \ // CA path is optional
```
jonathan-gramain:
```suggestion
  -e SIDECAR_TLS_CA_PATH=/certs/ca.pem \ // CA path is optional
```

README.md
```diff
@@ -1254,3 +1254,177 @@ docker run --net=host --entrypoint python3 -v /scality/ssd01/s3/scality-utapi/co
...
+  scality/s3utils service-level-sidecar/index.js
+```
+
+To enable TLS for internal connections to vault or bucketd set `SIDECAR_VAULT_ENABLE_TLS=true` or `SIDECAR_BUCKETD_ENABLE_TLS=true` respectively.
```
jonathan-gramain, line 1462: Normally, internal TLS is either enabled everywhere or disabled everywhere, I don't think it's possible to have only vault or bucketd with TLS, hence probably a single option is enough.

README.md
```diff
@@ -1254,3 +1254,177 @@ docker run --net=host --entrypoint python3 -v /scality/ssd01/s3/scality-utapi/co
...
+
+#### Warp10
+
+The Warp 10 address is configured using `SIDECAR_WARP10_NODE`.
+The Warp 10 `nodeId` must be included (normally matches ansible inventory name plus port ie `md1-cluster1:4802`).
+
+```shell
+docker run -d \
+  --network=host \
+  -e SIDECAR_WARP10_NODE="md1-cluster1:4802@127.0.0.1:4802" \
```
jonathan-gramain, line 1504: Could you describe what one should set before/after the '@' character?

service-level-sidecar/bucketd.js
```diff
@@ -0,0 +1,65 @@
...
+            const { key, value } = data;
+            const [account, name] = key.split(mdKeySplitter);
+            return {
+                account,
+                name,
+                value: JSON.parse(value),
+            };
+        });
+
+        if (res.length !== PAGE_SIZE) {
```
jonathan-gramain, line 55: What if the number of buckets is exactly a multiple of the `PAGE_SIZE`, do we get an empty bucket list for the next result? In which case it should be good.

service-level-sidecar/report.js
```diff
@@ -0,0 +1,158 @@
...
+ */
+class MetricReport {
+    constructor(factor = 1) {
+        this.factor = factor;
+        this.count = 0;
+        this.bytes = 0;
+    }
+
+    /**
+     * Add the given data to the metric report1
```
jonathan-gramain:
```suggestion
     * Add the given data to the metric report
```

service-level-sidecar/bucketd.js
```diff
@@ -0,0 +1,65 @@
...
+            // eslint-disable-next-line no-await-in-loop
+            res = await listObjects(usersBucket, { ...listingParams, gt }, log);
+        } catch (error) {
+            log.error('Error during listing', { error });
+            throw error;
+        }
+
+        log.debug('got list of buckets from bucketd', { length: res.length });
+
+        yield res.map(data => {
```
jonathan-gramain, line 45: We can probably simplify the API and hide the internal paging by changing to:
```suggestion
        yield *res.map(data => {
```

Which would normally yield individual entries instead of the list itself. But since we're processing bucket results in batches afterwards, it may not help as we would have to batch the results anyway.

service-level-sidecar/report.js
```diff
@@ -0,0 +1,158 @@
...
+ */
+async function getServiceReport(timestamp, log) {
+    const serviceReport = new MetricReport(env.scaleFactor);
+    const accountReports = {};
+    const bucketReports = {};
+    const accountInfoCache = {};
+
+    for await (const buckets of bucketd.listBuckets(log)) {
+        log.debug('got response from bucketd', { numBuckets: buckets.length });
+        await util.promisify(async.eachLimit)(buckets, env.concurrencyLimit, async bucket => {
```
jonathan-gramain, line 97: Is `promisify` needed because we're using an old async version (2.x)?

service-level-sidecar/server.js
```diff
@@ -0,0 +1,133 @@
...
+    const info = {
+        httpCode: res.statusCode,
+        httpMessage: res.statusMessage,
+    };
+    req.log.end('finished handling request', info);
+}
+
+// Hash our api key once for reuse during the request
+// We prepend `Bearer ` to avoid having to strip it from the header value runtime for comparison
+const actualKeyHash = crypto.createHash('sha512').copy().update(`Bearer ${env.apiKey}`).digest();
```
jonathan-gramain, line 67: Why using a hash here, why not just comparing the API key directly? (we may still use `timingSafeEqual` on the compared original values)

```
`crypto.timingSafeEqual(Buffer.from(`Bearer ${env.apiKey}`), Buffer.from(req.headers.authorization))
```

service-level-sidecar/server.js
```diff
@@ -0,0 +1,133 @@
...
+
+// Register a catch-all 404 handler
+// Must be defined after all other non-error middleware
+app.use((req, res, next) => {
+    next(makeError(404, 'Not Found'));
+});
+
+// Catch all error handler
+app.use((err, req, res, next) => {
+    req.log.error('error during request', { error: err });
```
jonathan-gramain, line 122: Can we also log the URL

service-level-sidecar/warp10.js
```diff
@@ -0,0 +1,47 @@
...
+        if (params.script) {
+            payload.push(params.script);
+        }
+        return payload.join('\n');
+    }
+
+    async exec(params) {
+        const payload = this._buildExecPayload(params);
+        const resp = await this._client.exec(payload);
+        log.debug('warpscript executed', { stats: resp.meta });
```
jonathan-gramain: Might be good to log the request params, can help debugging

jonathan-gramain (approved):

README.md
tmacro: It's to account for the overhead of erasure coding on the backend

service-level-sidecar/bucketd.js
tmacro: I think so, it's been a while since I actually wrote this code. It's taken from utapi v2's codebase

service-level-sidecar/report.js
tmacro: yeah, 2.x can accept an async function as the worker, but it always returns a callback style one.

service-level-sidecar/server.js
tmacro: `timingSafeEquals` requires Buffers of equal length or it throws. This was the easiest way I could think of converting the strings + ensuring the same length without leaking info.

service-level-sidecar/bucketd.js
tmacro: I didn't know about that syntax, thats cool. Originally I had it yielding one bucket at a time but I removed it some I could get the batching downstream.

README.md
tmacro: A  centos 7 vm I have has 7.29.0 which is a few versions short of when that flag was added. I think we should stick with `-H` for compat, and also because it tells you the header format even if you're not using curl. (one may not be familiar with oauth2 headers)

service-level-sidecar/server.js
jonathan-gramain: Can you please add a comment explaining this?

service-level-sidecar/server.js
tmacro: I've added the default params to the request logger

README.md
tmacro: Yeah, I heard some mention of https being a (soft) requirement and I wasn't sure if the platforms this would be running on had internal TLS or if this would be separate. I figured I'd just err on the side being flexible.

service-level-sidecar/report.js
```diff
@@ -0,0 +1,158 @@
...
+ * Generate a service level metrics report for the given timestamp.
+ * The report is in the following format:
+ *  {
+ *    service: { obj_count, bytes_stored },
+ *    account: [ { name, arn, obj_count, bytes_stored } ],
+ *    bucket: { <account_arn>: [ { name, obj_count, bytes_stored } ] }
+ *  }
+ *
+ * @param {number} timestamp - timestamp to generate the report for
+ * @param {*} log - werelogs logger instance
```
watsaqat: missing type?

service-level-sidecar/server.js
```diff
@@ -0,0 +1,153 @@
...
+const log = require('./log');
+
+const app = express();
+
+const getServiceReportCb = util.callbackify(getServiceReport);
+
+/**
+ *
+ * @param {number} statusCode - HTTP status code
+ * @param {*} httpMessage - HTTP status message
```
watsaqat: same for these two

watsaqat (approved):

PR #270, opened 2022-12-01
improvement: S3UTILS-115 replication verification util

naren-rajendran:


VerifyReplication/index.js
```diff
@@ -0,0 +1,176 @@
...
+
+const USAGE = `
+VerifyReplication/index.js
+
+This script verifies replication by comparing source and destination objects.
+
+Usage:
+    node VerifyReplication/index.js
+
+Mandatory environment variables:
```
ghost, line 51: I do see an improvement that might be useful for large buckets, in case the tool stops for any reason: why don't we periodically store the progress of workers, so that we can either start or restart at any point from a specific object in the original bucket?
What do you think?

naren-rajendran: @williamlardier yes its on my to-do list, it would certainly be an improvement. The current iteration is for a preview (not final).

tmacro (approved):

nicolas2bert (approved):

VerifyReplication/index.js
naren-rajendran: It is adding more complexity than originally visioned. It can be implemented as an improvement, if needed.

PR #272, opened 2022-12-08
Improvement/s3 utils 117 replication verification improvements corrections

naren-rajendran:


VerifyReplication/index.js
```diff
@@ -109,6 +110,10 @@ const destinationRequestWorkers = (DST_MD_REQUEST_WORKERS
     && Number.parseInt(DST_MD_REQUEST_WORKERS, 10)) || defaults.DESTINATION_MD_REQUEST_WORKERS;
 const compareObjectSize = COMPARE_OBJECT_SIZE === '1';
 const compareObjectAllVersions = COMPARE_OBJECT_ALL_VERSIONS === '1';
+const delimiter = SRC_DELIMITER || defaults.DELIMITER;
+const prefixes = SRC_BUCKET_PREFIXES ? SRC_BUCKET_PREFIXES.split(',')
+    .filter(x => x)
+    .map(x => (x.endsWith(delimiter) ? x : `${x}${delimiter}`)) : [];
```
jonathan-gramain, line 116: I wonder if we want this `map` or if it would be better to stick with AWS semantics and not append a delimiter if there is none, trusting the original bucket prefix to be what the user wants?

jonathan-gramain (approved):

nicolas2bert (approved):

VerifyReplication/index.js
naren-rajendran: I think it's easier to isolate issues with a specific prefix and we do not want to run partial again on a successful prefix. Also the initial listing (when prefix filters are not supplied to the script) uses delimiter to grab common prefixes (which has delimiter suffixed) to run parallel listing (to increase the verification rate). I want to keep it uniform.

PR #276, opened 2022-12-20
Improvement/s3 utils 119 modify count items for storage metrics

XinLi0207:
this PR implement a `CountItemsMongoClient` class which extends `MongoClientInterface` from arsenal
override `getObjectMDStats()` function  and some utils functions to output
```
{
        "objects" : 4,
        "versions" : 6,
        "buckets" : 3,
        "bucketList" : [ 
            {
                "name" : "bucket2",
                "location" : "us-east-1",
                "isVersioned" : true,
                "ownerCanonicalId" : "2fbfcc0125094acda54f0cabb5d1e66ae78aaf67f6060e8ac5b6abc9ada01d81",
                "ingestion" : false
            }, 
            {
                "name" : "bucket1",
                "location" : "us-east-1",
                "isVersioned" : true,
                "ownerCanonicalId" : "2fbfcc0125094acda54f0cabb5d1e66ae78aaf67f6060e8ac5b6abc9ada01d81",
                "ingestion" : false
            }
        ],
        "dataManaged" : {
            "total" : {
                "curr" : 392,
                "prev" : 222
            },
            "byLocation" : {
                "us-east-1" : {
                    "curr" : 392,
                    "prev" : 222
                }
            }
        },
        "stalled" : 0,
        "dataMetrics" : {
            "bucket" : {
                "bucket1" : {
                    "usedCapacity" : {
                        "current" : 376,
                        "nonCurrent" : 0
                    },
                    "objectCount" : {
                        "current" : 2,
                        "nonCurrent" : 0,
                        "deleteMarker" : 0
                    }
                },
                "bucket2" : {
                    "usedCapacity" : {
                        "current" : 6,
                        "nonCurrent" : 0
                    },
                    "objectCount" : {
                        "current" : 1,
                        "nonCurrent" : 0,
                        "deleteMarker" : 0
                    }
                },
            },
            "location" : {
                "us-east-1" : {
                    "usedCapacity" : {
                        "current" : 392,
                        "nonCurrent" : 222
                    },
                    "objectCount" : {
                        "current" : 4,
                        "nonCurrent" : 6,
                        "deleteMarker" : 1
                    }
                }
            },
            "account" : {
                "account2" : {
                    "usedCapacity" : {
                        "current" : 392,
                        "nonCurrent" : 222
                    },
                    "objectCount" : {
                        "current" : 4,
                        "nonCurrent" : 6,
                        "deleteMarker" : 1
                    }
                }
            }
        }
    }
}
```
added 3 metric levels: bucket/location/account and collect their usedCapacity and objectCount respectively

### DIFF
--- a/CountItems/CountManager.js
+++ b/CountItems/CountManager.js
@@ -1,5 +1,7 @@
 const async = require('async');
 const { once } = require('arsenal').jsutil;
+const { validStorageMetricLevels } = require('./utils/constants');
+const { consolidateDataMetrics } = require('./utils/utils');
 
 class CountManager {
     constructor(params) {
@@ -16,6 +18,11 @@ class CountManager {
                 byLocation: {},
             },
             stalled: 0,
+            dataMetrics: {
+                bucket: {},
+                location: {},
+                account: {},
+            },
         };
         this.workerList = [];
         this._setupQueue();
@@ -40,16 +47,18 @@ class CountManager {
     }
 
     _consolidateData(results) {
-        if (!results) return;
+        if (!results) {
+            return;
+        }
         this.store.versions += results.versions;
         this.store.objects += results.objects;
         this.store.stalled += results.stalled;
         if (results.dataManaged
             && results.dataManaged.locations
             && results.dataManaged.total) {
-            const { locations } = results.dataManaged;
-            this.store.dataManaged.total.curr += results.dataManaged.total.curr;
-            this.store.dataManaged.total.prev += results.dataManaged.total.prev;
+            const { locations, total } = results.dataManaged;
+            this.store.dataManaged.total.curr += total.curr;
+            this.store.dataManaged.total.prev += total.prev;
             Object.keys(locations).forEach(site => {
                 if (!this.store.dataManaged.byLocation[site]) {
                     this.store.dataManaged.byLocation[site] = { ...locations[site] };
@@ -60,6 +69,25 @@ class CountManager {
                         += locations[site].prev;
                 }
             });
+        }
+        if (results.dataMetrics
+            && results.dataMetrics.bucket
+            && results.dataMetrics.location
+            && results.dataMetrics.account) {
+            Object.keys(results.dataMetrics).forEach(metricLevel => {
+                // metricLevel can only be 'bucket', 'location' or 'account'
+                if (validStorageMetricLevels.has(metricLevel)) {
+                    Object.keys(results.dataMetrics[metricLevel]).forEach(resourceName => {
+                        // resourceName can be the name of bucket, location or account
+                        this.store.dataMetrics[metricLevel][resourceName] = consolidateDataMetrics(
+                            this.store.dataMetrics[metricLevel][resourceName],
+                            results.dataMetrics[metricLevel][resourceName],
+                        );
+                    });
+                }
+            });
+        } else {
+            this.store.dataMetrics = results.dataMetrics;
         }
     }
 
--- a/CountItems/utils/utils.js
+++ b/CountItems/utils/utils.js
@@ -1,0 +1,40 @@
+function consolidateDataMetrics(target, source) {
+    let resTarget = {};
+    if (target && (target instanceof Object)) {
+        resTarget = {
+            usedCapacity: target.usedCapacity,
+            objectCount: target.objectCount,
+        };
+    }
+    if (!resTarget.usedCapacity) {
+        Object.assign(resTarget, {
+            usedCapacity: {
+                current: 0,
+                nonCurrent: 0,
+            },
+        });
+    }
+    if (!resTarget.objectCount) {
+        Object.assign(resTarget, {
+            objectCount: {
+                current: 0,
+                nonCurrent: 0,
+                deleteMarker: 0,
+            },
+        });
+    }
+    if (!source) {
+        return resTarget;
+    }
+    const { usedCapacity, objectCount } = source;
+    resTarget.usedCapacity.current += usedCapacity && usedCapacity.current ? usedCapacity.current : 0;
+    resTarget.usedCapacity.nonCurrent += usedCapacity && usedCapacity.nonCurrent ? usedCapacity.nonCurrent : 0;
+    resTarget.objectCount.current += objectCount && objectCount.current ? objectCount.current : 0;
+    resTarget.objectCount.nonCurrent += objectCount && objectCount.nonCurrent ? objectCount.nonCurrent : 0;
+    resTarget.objectCount.deleteMarker += objectCount && objectCount.deleteMarker ? objectCount.deleteMarker : 0;
+    return resTarget;
+}
+
+module.exports = {
+    consolidateDataMetrics,
+};
--- a/tests/unit/CountItems/utils/utils.js
+++ b/tests/unit/CountItems/utils/utils.js
@@ -1,0 +1,60 @@
+const { consolidateDataMetrics } = require('../../../../CountItems/utils/utils');
+
+describe('CountItems::utils::consolidateDataMetrics', () => {
+    const zeroValueRes = {
+        usedCapacity: { current: 0, nonCurrent: 0 },
+        objectCount: { current: 0, nonCurrent: 0, deleteMarker: 0 },
+    };
+
+    const example1 = {
+        usedCapacity: { current: 10, nonCurrent: 10 },
+        objectCount: { current: 10, nonCurrent: 10, deleteMarker: 10 },
+    };
+
+    const example2 = {
+        usedCapacity: { current: 20, nonCurrent: 20 },
+        objectCount: { current: 20, nonCurrent: 20, deleteMarker: 20 },
+    };
+
+    test('should return zero-value if target and source are both undefined', () => {
+        const res = consolidateDataMetrics(undefined, undefined);
+        expect(res).toEqual(zeroValueRes);
+    });
+
+    test('should return zero-value if target and source are both empty', () => {
+        const res = consolidateDataMetrics({}, {});
+        expect(res).toEqual(zeroValueRes);
+    });
+
+    test('should return value of target if source are empty', () => {
+        const target = example1;
+        const res = consolidateDataMetrics(target, {});
+        expect(res).toEqual(target);
+    });
+
+    test('should return value of source if target are empty', () => {
+        const source = example1;
+        const res = consolidateDataMetrics({}, source);
+        expect(res).toEqual(source);
+    });
+
+    test('should correctly consolidate source and target', () => {
+        const source = example1;
+        const target = example1;
+        const res = consolidateDataMetrics(target, source);
+        expect(res).toEqual(example2);
+    });
+
+    test('should not consolidate data other than usedCapacity and objectCount', () => {
+        const source = {
+            usedCapacity123: example1.usedCapacity,
+            objectCount456: example1.objectCount,
+        };
+        const target = {
+            usedCapacity123: example1.usedCapacity,
+            objectCount456: example1.objectCount,
+        };
+        const res = consolidateDataMetrics(target, source);
+        expect(res).toEqual(zeroValueRes);
+    });
+});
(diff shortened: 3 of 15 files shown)


anurag4DSB (approved):

CountItems/CountManager.js
```diff
@@ -61,6 +66,27 @@ class CountManager {
...
+                            });
+                        });
+                    } else {
+                        this.store.dataMetrics[metricLevel][resourceName] = results.dataMetrics[metricLevel][resourceName];
+                    }
+                });
+            });
+        } else {
+            this.store.dataMetrics = results.dataMetrics;
+        }
```
anurag4DSB: It would be great to have a comment here, to explain this logic and why it is needed.

anurag4DSB: This commend might apply here as well https://github.com/scality/Arsenal/pull/2055/files#r1054615483

XinLi0207:
hi @alexanderchan-scality, as per your comment [here](https://github.com/scality/Arsenal/pull/2055#issuecomment-1363309467), I implemented a new `CountItemsMongoClient` which extends `MongoClientInterface` from arsenal and overrides `getObjectMDStats()` and some other functions, I agree with what you said, it's indeed better to be closer to CountItems logic and should not be a common mongo client function.
I have also moved all the tests I have done in this [PR](https://github.com/scality/Arsenal/pull/2055) here as well, so now there is no change need to be done at arsneal side, I will close that PR if this PR looks good to you.

tests/utils/mongoUtils.js
```diff
@@ -0,0 +1,33 @@
...
+    debug: false,
+    instanceOpts: [
+        { port: 27018 },
+    ],
+    replSet: {
+        name: 'customSetName',
+        count: 1,
+        dbName: 'metadata',
+        storageEngine: 'ephemeralForTest',
+    },
```
ghost, line 12: Maybe we can set the version of mongodb we want to test in the binaryOpts of this object? As we use a quite old version currently, it will be better to use it in our tests as well, to make sure we are not relying on anything unsupported

CountItems/utils/utils.js
```diff
@@ -0,0 +1,42 @@
+function consolidateDataMetrics(target, source) {
+    let resTarget;
+    if (!target || !(target instanceof Object)) {
+        resTarget = {};
+    } else {
+        resTarget = {
+            usedCapacity: target.usedCapacity,
+            objectCount: target.objectCount,
+        };
+    }
```
ghost: Suggestion to simplify the code:

```suggestion
    let resTarget = {};
    if (target && (target instanceof Object)) {
        resTarget = {
            usedCapacity: target.usedCapacity,
            objectCount: target.objectCount,
        };
    }
```

CountItems/utils/utils.js
```diff
@@ -0,0 +1,42 @@
...
+    if (!source) {
+        return resTarget;
+    }
+    const { usedCapacity, objectCount } = source;
+    resTarget.usedCapacity.current += usedCapacity && usedCapacity.current ? source.usedCapacity.current : 0;
+    resTarget.usedCapacity.nonCurrent += usedCapacity && usedCapacity.nonCurrent ? source.usedCapacity.nonCurrent : 0;
+    resTarget.objectCount.current += objectCount && objectCount.current ? source.objectCount.current : 0;
+    resTarget.objectCount.nonCurrent += objectCount && objectCount.nonCurrent ? source.objectCount.nonCurrent : 0;
+    resTarget.objectCount.deleteMarker += objectCount && objectCount.deleteMarker ? source.objectCount.deleteMarker : 0;
+    return resTarget;
```
ghost:
```suggestion
    const { usedCapacity, objectCount } = source;
    resTarget.usedCapacity.current += usedCapacity && usedCapacity.current ? usedCapacity.current : 0;
    resTarget.usedCapacity.nonCurrent += usedCapacity && usedCapacity.nonCurrent ? usedCapacity.nonCurrent : 0;
    resTarget.objectCount.current += objectCount && objectCount.current ? objectCount.current : 0;
    resTarget.objectCount.nonCurrent += objectCount && objectCount.nonCurrent ? objectCount.nonCurrent : 0;
    resTarget.objectCount.deleteMarker += objectCount && objectCount.deleteMarker ? objectCount.deleteMarker : 0;
    return resTarget;
```

tests/unit/CountItems/utils/utils.js
```diff
@@ -0,0 +1,69 @@
+const { consolidateDataMetrics } = require('../../../../CountItems/utils/utils');
+
+describe('CountItems::utils::consolidateDataMetrics', () => {
+    test('should return zero-value if target and source are both undefined', () => {
+        const res = consolidateDataMetrics(undefined, undefined);
+        expect(res).toEqual({
+            usedCapacity: { current: 0, nonCurrent: 0 },
+            objectCount: { current: 0, nonCurrent: 0, deleteMarker: 0 },
+        });
```
ghost: Might be helpful to crate these objects as shared constants to avoid duplication

CountItems/CountManager.js
```diff
@@ -61,6 +70,25 @@ class CountManager {
                 }
             });
         }
+        if (results.dataMetrics
+            && results.dataMetrics.bucket
+            && results.dataMetrics.location
+            && results.dataMetrics.account) {
```
ghost, line 76: This will avoid updating the store dataMetrics if either of these 3 fields are not in the object, is this something expected? (-> or if I put in differently, can `dataMtrics` not have these three fields?)

ghost (commented):
Mostly completes my first review on arsenal before the code migration. LGTM otherwise!

My main "concern" is on the `CountItemMongoClient` that duplicates some code with Arsenal while extending the class: this may be affected by changes in the Arsenal class after a bump in s3util if we are lacking tests on some parts, as we also rely on class methods.

CountItems/CountManager.js
XinLi0207: yeah it's kind of expected, because there won't be a case that there are metrics in buckets but not in location nor account, all objects must be linked to a bucket, an account and one or multiple locations. so if there is only bucket metrics, we assume the dataMetrics is broken

tests/utils/mongoUtils.js
XinLi0207: seems `mongodb-memory-server` doesn't support specifying the mongo version we are currently using, 4.0.1

utils/S3UtilsMongoClient.js
```diff
@@ -0,0 +1,246 @@
...
+            // there can be multiple locations for an object entry
+            location: {},
+            // there will be only one account for an object entry
+            account: { [entry.value['owner-display-name']]: size },
+        };
+
+        if (!isTransient
+            || entry.value.replicationInfo.status !== 'COMPLETED') {
+            // only count it in current dataStore if object is not in transient or replication not completed
+            if (results.location[entry.value.dataStoreName]) {
```
nicolas2bert: When does this condition return true?
Is `_processEntryData` not re-initialized for every call?

utils/S3UtilsMongoClient.js
```diff
@@ -0,0 +1,246 @@
...
+                results.location[entry.value.dataStoreName] += size;
+            } else {
+                results.location[entry.value.dataStoreName] = size;
+            }
+        }
+        entry.value.replicationInfo.backends.forEach(rep => {
+            // count it in the replication destination location if replication compeleted
+            if (rep.status === 'COMPLETED') {
+                if (results.location[rep.site]) {
```
nicolas2bert: Same for this one:
The location source and destination should be different and each destination location should be different.
Worst case, if the location happens to be duplicated, we should overwrite the size.

utils/S3UtilsMongoClient.js
```diff
@@ -0,0 +1,246 @@
...
+                if (res._id.indexOf('\0') !== -1) {
+                    // versioned item
+                    targetCount = 'versionCount';
+                    targetData = 'versionData';
+
+                    if (res.value.replicationInfo.backends.length > 0
+                        && this._isReplicationEntryStalled(res, cmpDate)) {
+                        stalledCount++;
+                    }
+                } else if (!!res.value.versionId && !res.value.isNull) {
```
nicolas2bert, line 52: Why do we check for isNull here?

From what I remember, "null versions" reference the objects located in an unversioned bucket.

utils/S3UtilsMongoClient.js
```diff
@@ -0,0 +1,246 @@
...
+                        masterCount,
+                        masterData,
+                        nullCount,
+                        nullData,
+                        versionCount,
+                        versionData,
+                        deleteMarkerCount,
+                    } = res[metricLevel][resourceName];
+
+                    dataMetrics[metricLevel][resourceName].usedCapacity.current += nullData + masterData;
```
nicolas2bert, line 196: `nullData` is referencing the size of the objects in an unversioned bucket.
`masterData` is referencing the size of the master versions in a versioned/suspended bucket.

For a given bucket state one or the other shoule be 0. Is it why we add them here?

utils/S3UtilsMongoClient.js
```diff
@@ -0,0 +1,246 @@
...
+                    dataMetrics[metricLevel][resourceName].objectCount.current += nullCount + masterCount;
+
+                    if (isVersioned) {
+                        dataMetrics[metricLevel][resourceName].usedCapacity.nonCurrent += versionData - masterData;
+                        dataMetrics[metricLevel][resourceName].usedCapacity.nonCurrent = Math.max(dataMetrics[metricLevel][resourceName].usedCapacity.nonCurrent, 0);
+                        dataMetrics[metricLevel][resourceName].objectCount.nonCurrent
+                            += versionCount - masterCount - deleteMarkerCount;
```
nicolas2bert, line 204: Why do we need to subtract deleteMarkers from versions here? deleteMarker(s) are not considered as versions?

utils/S3UtilsMongoClient.js
XinLi0207: yeah you are right, we should simply overwrite the size for every destination location, there is no need to accumulate it in one bucket.

utils/S3UtilsMongoClient.js
XinLi0207: yeah, this condition can never be true, `results.location[entry.value.dataStoreName]` should simply be overwrited

utils/S3UtilsMongoClient.js
XinLi0207: the previous `countItmes` service has an issue when bucket is suspended, because of [this logic ](https://github.com/scality/Arsenal/blob/development/8.1/lib/storage/metadata/mongoclient/MongoClientInterface.js#L2077) when location's nonCurrent size is got by versioned size minus master size(it works in versioned bucket, because master data are duplicated), but the object uploaded after the bucket is suspended will only have master data in DB, so this record is not duplicated so should not be subtracted

that's why I take out of `isNull` data from master data and treat it as non versioned data

utils/S3UtilsMongoClient.js
XinLi0207: yeah, `nullData` represent current data in nonversioned bucket/ object uploaded after bucket suspended.
`masterData` represents current data in versioned bucket/object uploaded before bucket suspended
if one has value, the other is 0

utils/S3UtilsMongoClient.js
XinLi0207: we have 3 metrics for `usedCapacity`: current, nonCurrent, deleteMarker
I feel like it's more logical to have them separated, we can add them up if needed

nicolas2bert (approved):

utils/S3UtilsMongoClient.js
nicolas2bert: Would you mind adding a comment here?

XinLi0207: yeah, no problem
Thanks for reviewing!

ghost (approved):
I thought I already approved, but no :smile: 

Let's wait for the end2end test before merging, if you agree!

alexanderchan-scality (approved):

PR #277, opened 2022-12-27
Feature/s3 utils 122 data report collect metrics write capacity info

XinLi0207:
this PR is based on [this PR](https://github.com/scality/s3utils/pull/276) to reuse the MongoClientInterface inherited class

this PR mainly implemented the script for the cronjob designed [here](https://github.com/scality/citadel/blob/development/1.0/docs/design/veeam-sosapi.md#cron-job)

the script does the following things:
- collects bucket storage metrics from `INFOSTORE` collection (which is updated by `countitems` cronjob hourly)
- read `Capacity` field from bucketMD (`Capacity` field is put into bucketMD by `capacity.xml` objectPut operation)
- check the validity of the metrics retrieved above
  - if valid, write back to buketMD `Avaible`, `Used` fields after some calculation
  - if not valid, write back `-1`

### DIFF
--- a/DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
+++ b/DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
@@ -1,0 +1,133 @@
+const async = require('async');
+const { MaxParallelLimit } = require('../utils/constants');
+
+function isSOSCapacityInfoEnabled(bucketInfo) {
+    /**
+     * a valid SOS CapacityInfo Enabled bucketMD looks like
+     * {
+     *     _capabilities: {
+     *         VeeamSOSApi: {
+     *             SystemInfo: {
+     *                 ...
+     *                 ProtocolCapabilities: {
+     *                     CapacityInfo: true,
+     *                     ...
+     *                 },
+     *             ...
+     *             },
+     *             CapacityInfo: {
+     *                 ...
+     *             },
+     *         }
+     *     },
+     * }
+     */
+    return !!(bucketInfo.getCapabilities() // should exist
+        && bucketInfo.getCapabilities().VeeamSOSApi // should exist
+        && bucketInfo.getCapabilities().VeeamSOSApi.SystemInfo // should exist
+        && bucketInfo.getCapabilities().VeeamSOSApi.SystemInfo.ProtocolCapabilities // should exist
+        && bucketInfo.getCapabilities().VeeamSOSApi.SystemInfo.ProtocolCapabilities.CapacityInfo === true // should be true
+        && bucketInfo.getCapabilities().VeeamSOSApi.CapacityInfo); // should exist
+}
+
+function isValidBucketStorageMetrics(bucketMetric) {
+    return bucketMetric
+    && bucketMetric.usedCapacity
+        && typeof bucketMetric.usedCapacity.current === 'number'
+        && typeof bucketMetric.usedCapacity.nonCurrent === 'number'
+    && bucketMetric.usedCapacity.current > -1
+    && bucketMetric.usedCapacity.nonCurrent > -1;
+}
+
+function isValidCapacityValue(capacity) {
+    return (Number.isSafeInteger(capacity) && capacity >= 0);
+}
+
+function collectBucketMetricsAndUpdateBucketCapacityInfo(mongoClient, log, callback) {
+    let bucketMetrics;
+    return async.waterfall([
+        next => mongoClient.readCountItems(log, (err, doc) => {
+            if (err) {
+                log.error('an error occurred during readCountItems', {
+                    error: { message: err.message },
+                });
+                return next(err);
+            }
+            if (doc && doc.dataMetrics && doc.dataMetrics.bucket) {
+                bucketMetrics = doc.dataMetrics.bucket;
+            } else {
+                log.info('no buckets metrics in countItems, exit');
+                return callback(null);
+            }
+            return next();
+        }),
+        next => mongoClient.getBucketInfos(log, (err, bucketInfos) => {
+            if (err) {
+                log.error('error getting bucket list', {
+                    error: err,
+                });
+                return next(err);
+            }
+            return next(null, bucketInfos.bucketInfos);
+        }),
+        (bucketInfoList, next) => async.eachLimit(
+            bucketInfoList,
+            MaxParallelLimit,
+            (bucket, cb) => {
+                if (!isSOSCapacityInfoEnabled(bucket)) {
+                    return cb();
+                }
+                const bucketName = bucket.getName();
+
+                // get bucket storage used
+                let bucketStorageUsed = -1;
+                if (isValidBucketStorageMetrics(bucketMetrics[bucketName])) {
+                    bucketStorageUsed = bucketMetrics[bucketName].usedCapacity.current
+                            + bucketMetrics[bucketName].usedCapacity.nonCurrent;
+                }
+
+                // read Capacity from bucket._capabilities
+                const { Capacity } = bucket.getCapabilities().VeeamSOSApi.CapacityInfo;
+
+                let available = -1;
+                let capacity = -1;
+                if (isValidCapacityValue(Capacity)) { // is Capacity value is valid
+                    capacity = Capacity;
+                    // if bucket storage used is valid and capacity is bigger than used
+                    if (bucketStorageUsed !== -1 && (capacity - bucketStorageUsed) >= 0) {
+                        available = capacity - bucketStorageUsed;
+                    }
+                }
+
+                return mongoClient.updateBucketCapacityInfo(bucketName, {
+                    Capacity: capacity,
+                    Available: available,
+                    Used: bucketStorageUsed,
+                }, log, err => {
+                    if (err) {
+                        log.error('an error occurred during put bucket CapacityInfo attributes', {
+                            error: { message: err.message },
+                            bucketName,
+                        });
+                        return cb(err);
+                    }
+                    return cb();
+                });
+            },
+            err => {
+                if (err) {
+                    return next(err);
+                }
+                log.info('successfully updated CapacityInfo for all qualified buckets');
+                return next(null);
+            },
+        ),
+    ], callback);
+}
+
+module.exports = {
+    isSOSCapacityInfoEnabled,
+    isValidBucketStorageMetrics,
+    isValidCapacityValue,
+    collectBucketMetricsAndUpdateBucketCapacityInfo,
+};
--- a/DataReport/index.js
+++ b/DataReport/index.js
@@ -1,0 +1,39 @@
+const { Logger } = require('werelogs');
+const createMongoParams = require('../utils/createMongoParams');
+const { collectBucketMetricsAndUpdateBucketCapacityInfo } = require('./collectBucketMetricsAndUpdateBucketCapacityInfo');
+const S3UtilsMongoClient = require('../utils/S3UtilsMongoClient');
+
+const log = new Logger('s3utils:DataReport');
+
+function stop() {
+    log.info('stopping execution');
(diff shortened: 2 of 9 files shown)


DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,132 @@
...
+                    error: err,
+                });
+                return next(err);
+            }
+            return next(null, bucketInfos.bucketInfos);
+        }),
+        (bucketInfoList, next) => async.filterLimit(
+            bucketInfoList,
+            10,
+            (bucketInfo, cb) => cb(null, isSOSCapacityInfoEnabled(bucketInfo)),
```
ghost: Not sure this will be really helpful in terms of performance, as we are trying to parallelize sync functions (no I/Os), that can be done in a single loop

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,132 @@
...
+                let { Capacity } = bucket.getCapabilities().VeeamSOSApi.CapacityInfo;
+
+                let Available;
+                if (isValidCapacityValue(Capacity)) { // is Capacity value is valid
+                    if (bucketStorageUsed !== -1) { // if bucket storage used is valid
+                        // check if bucket storage used is smaller than total Capacity
+                        Available = (Capacity - bucketStorageUsed) >= 0 ? (Capacity - bucketStorageUsed) : -1;
+                    } else { // if bucket storage used is not retrievable
+                        Available = -1;
+                    }
```
ghost: Subjective nit:

```suggestion
                    // if bucket storage used is valid and bigger than the reported used capacity
                    if (bucketStorageUsed !== -1 && (Capacity - bucketStorageUsed) >= 0) { 
                        Available = Capacity - bucketStorageUsed;
                    } else { // if bucket storage used is not retrievable
                        Available = -1;
                    }
```

utils/S3UtilsMongoClient.js
```diff
@@ -241,6 +244,34 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+        m.findOneAndUpdate({
+            _id: bucketName,
+        }, {
+            $set: {
+                '_id': bucketName,
+                'value.capabilities.VeeamSOSApi.CapacityInfo': {
+                    Capacity: capacityInfo.Capacity,
+                    Available: capacityInfo.Available,
+                    Used: capacityInfo.Used,
+                    LastModified: (new Date()).toJSON(),
```
ghost: You can store it as a string, with `toISOString`, as this is stored using this format in the PUT request:

```suggestion
                    LastModified: (new Date()).toISOString(),
```

DataReport/index.js
```diff
@@ -0,0 +1,38 @@
...
+            });
+            process.exit(1);
+        }
+        collectBucketMetricsAndUpdateBucketCapacityInfo(mongoClient, log, err => {
+            if (err) {
+                log.error('error collecting bucket metrics and updating bucket capacity info', {
+                    error: err,
+                });
+                process.exit(1);
+            }
```
ghost: Just for consistency, we can call `stop()` here, to also make sure we control the return code.

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,132 @@
...
+                log.error('error getting bucket list', {
+                    error: err,
+                });
+                return next(err);
+            }
+            return next(null, bucketInfos.bucketInfos);
+        }),
+        (bucketInfoList, next) => async.filterLimit(
+            bucketInfoList,
+            10,
```
ghost: Might be great to have it as a const shared with other steps for this async waterfall

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,132 @@
...
+            (bucketInfo, cb) => cb(null, isSOSCapacityInfoEnabled(bucketInfo)),
+            (err, capacityEnabledBuckets) => {
+                next(err, capacityEnabledBuckets);
+            },
```
ghost: Are these two steps really needed, while we can check this in the `eachLimit` (while better handling I/Os), and thus ignore if bucket is not SOSApi-enabled?

ghost (commented):
Will approve after end2end testing with a Veeam12 and other PRs

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,138 @@
...
+     *             },
+     *         }
+     *     },
+     * }
+     */
+    return !!(bucketInfo.getCapabilities() // should exist
+        && bucketInfo.getCapabilities().VeeamSOSApi // should exist
+        && bucketInfo.getCapabilities().VeeamSOSApi.SystemInfo // should exist
+        && bucketInfo.getCapabilities().VeeamSOSApi.SystemInfo.ProtocolCapabilities // should exist
+        && bucketInfo.getCapabilities().VeeamSOSApi.SystemInfo.ProtocolCapabilities.CapacityInfo === true // should be true
```
anurag4DSB, line 29: Could we use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) (`.?`) here?

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,126 @@
...
+    return bucketMetric
+    && bucketMetric.usedCapacity
+        && typeof bucketMetric.usedCapacity.current === 'number'
+        && typeof bucketMetric.usedCapacity.nonCurrent === 'number'
+    && bucketMetric.usedCapacity.current > -1
+    && bucketMetric.usedCapacity.nonCurrent > -1;
+}
+
+function isValidCapacityValue(capacity) {
+    return (typeof capacity === 'number' && capacity >= 0);
```
ghost: Just wondering if we shoudl reject floating numbers? They will be accepted here if the user puts something like `123456.789`. But it's a minor concern.

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,126 @@
...
+            if (err) {
+                log.error('an error occurred during readCountItems', {
+                    error: { message: err.message },
+                });
+                return next(err);
+            }
+            if (doc && doc.dataMetrics && doc.dataMetrics.bucket) {
+                bucketMetrics = doc.dataMetrics.bucket;
+            }
+            return next();
```
ghost: Just to be safe, can we return an error if we do not have any dataMetrics? In this case, `bucketMetrics` is still undefined, but we use it line 81 as an array without any check

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,126 @@
...
+                }
+
+                return mongoClient.updateBucketCapacityInfo(bucketName, {
+                    Capacity,
+                    Available,
+                    Used: bucketStorageUsed,
+                }, log, err => {
+                    if (err) {
+                        log.error('an error occurred during put bucket CapacityInfo attributes', {
+                            error: { message: err.message },
```
ghost, line 109: Might be useful to specify in the error message the current bucket, because we may have this error message multiple time, as it's running inside a loop

utils/S3UtilsMongoClient.js
```diff
@@ -238,6 +241,34 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                },
+            },
+        }, {
+            upsert: false,
+        }, err => {
+            if (err) {
+                log.error(
+                    'updateBucketCapacityInfo: error putting bucket CapacityInfo',
+                    { error: err.message },
+                );
```
ghost: Same here for the bucket name, and maybe capacity info as well?

ghost (approved):
Some minor changes, otherwise LGTM

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,138 @@
...
+                if (isValidBucketStorageMetrics(bucketMetrics[bucketName])) {
+                    bucketStorageUsed = bucketMetrics[bucketName].usedCapacity.current
+                            + bucketMetrics[bucketName].usedCapacity.nonCurrent;
+                }
+
+                // read Capacity from bucket._capabilities
+                let { Capacity } = bucket.getCapabilities().VeeamSOSApi.CapacityInfo;
+
+
+                let Available;
```
anurag4DSB: nit for readability: would suggest using lower case variable names. 
Upper case is okay for object keys

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -0,0 +1,138 @@
...
+                    if (bucketStorageUsed !== -1) { // if bucket storage used is valid
+                        // check if bucket storage used is smaller than total Capacity
+                        Available = (Capacity - bucketStorageUsed) >= 0 ? (Capacity - bucketStorageUsed) : -1;
+                    } else { // if bucket storage used is not retrievable
+                        Available = -1;
+                    }
+                } else { // is Capacity value is not valid
+                    Capacity = -1;
+                    Available = -1;
+                }
```
anurag4DSB: nit: May be we can initialize the variables, `Available` and `Capacity` with the value `-1` value and then change it if we need to. that will reduce some conditions.

package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@senx/warp10": "^1.1.2",
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/arsenal#8.1.59",
+    "arsenal": "git+https://github.com/scality/arsenal#c460338163bcc9decde7c18d2835beeef53273e6",
```
anurag4DSB: This will need an updated tag. Commenting so that we do not forget.
Reference PR: https://github.com/scality/Arsenal/pull/2059/commits

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
anurag4DSB: We should round off.
what do you think?

anurag4DSB (commented):
over all logic looks good with the design.
Can use some refactoring using optional chaining. I commented only at one place, but there are a few opportunities.

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
XinLi0207: maybe just return without any error, when there are no buckets, there are no bucket metrics, we should just let the job finish

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
anurag4DSB: discussed further over slack, as we are always dealing in bytes, rejecting floats should be okay.

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
XinLi0207: the `jest` we are using in this project doesn't support optional chaining

anurag4DSB (approved):

PR #280, opened 2023-01-12
Improvement/s3 utils 123 sosapi cronjob alerting

XinLi0207:
there are two cronjobs defined in s3utils
- count-items: counting object number and data size in the level of account/location/bucket hourly
- update-bucket-capacity-info: it's a feature of SOSAPI, it reads metrics generated by count-items and updates bucket metrics in each bucket's metadata every 4 mins

I added alerts for the two cronjobs accordingly
- for count-items cronjob: 
  - alert when job fails
  - alert when job takes too long(longer than cronjob interval)

- for update-bucket-capacity-info cronjob:
  - alert when job fails
  - alert when job takes too long(longer than cronjob inerval)
  - alert when no successful job runs in last 10m (10m is critical for Veeam to interact with Artesca bucket)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,36 +11,25 @@ on:
         default: 'false'
         required: false
 
+env:
+  REGISTRY_NAME: registry.scality.com
+  PROJECT_NAME: ${{ github.event.repository.name }}
+
 jobs:
-  release:
+  build:
+    name: Build and Push
+    uses: ./.github/workflows/build.yml
+    with:
+      namespace: s3utils
+      tag: ${{ github.event.inputs.tag }}
+    secrets: inherit
+
+  github-release:
+    name: Create GitHub Release
     runs-on: ubuntu-latest
+    needs:
+      - build
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
-
-      - name: Login to Registry
-        uses: docker/login-action@v1
-        with:
-          registry: registry.scality.com
-          username: ${{ secrets.REGISTRY_LOGIN }}
-          password: ${{ secrets.REGISTRY_PASSWORD }}
-
-      - name: Build and push
-        uses: docker/build-push-action@v2
-        with:
-          context: .
-          file: ./Dockerfile
-          push: true
-          tags: "registry.scality.com/s3utils/s3utils:${{ github.event.inputs.tag }}"
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
-
       - name: Create Release
         uses: softprops/action-gh-release@v1
         env:
--- a/monitoring/count-items-cronjob/alerts.yaml
+++ b/monitoring/count-items-cronjob/alerts.yaml
@@ -1,0 +1,29 @@
+x-inputs:
+  - name: namespace
+    type: constant
+    value: zenko
+  - name: count_items_cronjob
+    type: constant
+    value: artesca-data-ops-count-items
+  - name: count_items_job_duration_threshold
+    type: config
+    value: 3600
+
+groups:
+  - name: count-items-cronjob/alerts.rules
+    rules:
+      - alert: CountItemsJobTakingTooLong
+        Expr: |
+          time() -
+          (sum by(job_name) (kube_job_status_failed{job_name=~"${count_items_cronjob}.*"})
+          > sum by(job_name) (kube_job_status_completion_time{job_name=~"${count_items_cronjob}.*"})
+          or sum by(job_name) (kube_job_status_completion_time{job_name=~"${count_items_cronjob}.*"}))
+          > ${count_items_job_duration_threshold}
+        Labels:
+          severity: warning
+        Annotations:
+          description: |
+            Job ${count_items_cronjob} is taking more than ${count_items_job_duration_threshold}s to complete.
+            This means that consumption information will be delayed
+            and the reported metrics might not be accurate anymore.
+          summary: count-items cronjob takes too long to finish
--- a/monitoring/update-bucket-capacity-info-cronjob/alerts.test.yaml
+++ b/monitoring/update-bucket-capacity-info-cronjob/alerts.test.yaml
@@ -1,0 +1,53 @@
+evaluation_interval: 1m
+rule_files:
+  - alerts.rendered.yaml
+
+tests:
+  - name: Update Bucket Capacity Info CronJob Test Taking Too Long
+    interval: 1m
+    input_series:
+      - series: 'kube_job_status_start_time{job="kube-state-metrics", job_name="artesca-data-ops-update-bucket-capacity-info-1", namespace="zenko"}'
+        values: '0x9'
+      - series: 'kube_job_status_completion_time{job="kube-state-metrics", job_name="artesca-data-ops-update-bucket-capacity-info-1", namespace="zenko"}'
+        values: '_x4 240x5'
+      - series: 'kube_job_status_start_time{job="kube-state-metrics", job_name="artesca-data-ops-update-bucket-capacity-info-2", namespace="zenko"}'
+        values: '_x4 240x5'
+      - series: 'kube_job_status_completion_time{job="kube-state-metrics", job_name="artesca-data-ops-update-bucket-capacity-info-2", namespace="zenko"}'
+        values: '_x9 '
+    alert_rule_test:
+      - alertname: UpdateBucketCapacityJobTakingTooLong
+        eval_time: 4m
+        exp_alerts: []
+      - alertname: UpdateBucketCapacityJobTakingTooLong
+        eval_time: 9m
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              job_name: artesca-data-ops-update-bucket-capacity-info-1
+            exp_annotations:
+              description: |
+                Job artesca-data-ops-update-bucket-capacity-info is taking more than 240s to complete.
+                This may cause bucket capacity to be out of date and Veeam SOSAPI avalability as risk.
+              summary: update-bucket-capacity-info cronjob takes too long to finish
+
+  - name: Update Bucket Capacity Info CronJob Test No Success in 10m
+    interval: 1m
+    input_series:
+      - series: 'kube_job_status_completion_time{job="kube-state-metrics", job_name="artesca-data-ops-update-bucket-capacity-info-1", namespace="zenko"}'
+        values: '0x11'
+    alert_rule_test:
+      - alertname: NoSuccessfulUpdateBucketCapacityJobRunIn10m
+        eval_time: 10m
+        exp_alerts: []
+      - alertname: NoSuccessfulUpdateBucketCapacityJobRunIn10m
+        eval_time: 11m
+        exp_alerts:
+          - exp_labels:
+              severity: critical
+              job_name: artesca-data-ops-update-bucket-capacity-info-1
+            exp_annotations:
+              description: |
+                No successful artesca-data-ops-update-bucket-capacity-info-1
+                job run in the last 600s, will cause bucket capacity to be out of date
+                and Veeam will process all buckets in non-SOSAPI mode.
+              summary: "No successful artesca-data-ops-update-bucket-capacity-info-1 job run, affecting Veeam SOSAPI"
--- a/monitoring/update-bucket-capacity-info-cronjob/alerts.yaml
+++ b/monitoring/update-bucket-capacity-info-cronjob/alerts.yaml
@@ -1,0 +1,46 @@
+# Variables which should be replaced. Similar to grafana dashboards' __inputs section
+x-inputs:
+  - name: namespace
+    type: constant
+    value: zenko
+  - name: update_bucket_capacity_info_cronjob
+    type: constant
+    value: artesca-data-ops-update-bucket-capacity-info
+  - name: update_bucket_capacity_info_job_duration_threshold
+    type: config
+    value: 240
+  - name: update_bucket_capacity_info_success_job_existence_duration_threshold
+    type: config
+    value: 600
+
+groups:
+  - name: update-bucket-capacity-info-cronjob/alerts.rules
+    rules:
+      - alert: UpdateBucketCapacityJobTakingTooLong
+        expr: |
+          time() -
+          (sum by(job_name) (kube_job_status_failed{job_name=~"${update_bucket_capacity_info_cronjob}.*"})
+          > sum by(job_name) (kube_job_status_completion_time{job_name=~"${update_bucket_capacity_info_cronjob}.*"})
+          or sum by(job_name) (kube_job_status_completion_time{job_name=~"${update_bucket_capacity_info_cronjob}.*"}))
+          > ${update_bucket_capacity_info_job_duration_threshold}
+        labels:
+          severity: warning
+        annotations:
+          description: |
+            Job ${update_bucket_capacity_info_cronjob} is taking more than ${update_bucket_capacity_info_job_duration_threshold}s to complete.
+            This may cause bucket capacity to be out of date and Veeam SOSAPI avalability as risk.
+          summary: update-bucket-capacity-info cronjob takes too long to finish
+
+      - alert: NoSuccessfulUpdateBucketCapacityJobRunIn10m
+        expr: |
+          time() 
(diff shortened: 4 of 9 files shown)


monitoring/update-bucket-capacity-info-cronjob/alerts.yaml
```diff
@@ -0,0 +1,99 @@
...
+x-inputs:
+  - name: namespace
+    type: constant
+    value: zenko
+  - name: update_bucket_capacity_info_cronjob
+    type: constant
+    value: artesca-data-ops-update-bucket-capacity-info
+  - name: update_bucket_capacity_info_job_duration_threshold
+    type: config
+    value: 240
```
XinLi0207, line 11: 4m is the interval of `update-bucket-capacity-info` cronjob, will alert if a job runs longer than this duration

monitoring/update-bucket-capacity-info-cronjob/alerts.yaml
```diff
@@ -0,0 +1,99 @@
...
+  - name: update_bucket_capacity_info_success_job_existence_duration_threshold
+    type: config
+    value: 600
```
XinLi0207, line 14: if no successful jobs in 10m which means bucket capacityInfo meta is not being updated since 10m, Veaam will refuse to interact with Artesca buckets if the data last update time is longer than 10m.

.github/workflows/test.yaml
```diff
@@ -52,6 +56,12 @@ jobs:
           username: ${{ secrets.REGISTRY_LOGIN }}
           password: ${{ secrets.REGISTRY_PASSWORD }}
 
+      - name: Push dashboards into the production namespace
+        run: |
+          oras push ${{ env.REGISTRY_NAME }}/${{ env.PROJECT_NAME }}/${{ env.PROJECT_NAME }}-dashboards:${{ github.event.inputs.tag }} \
+            count-items-cronjob/alerts.yaml:application/prometheus-alerts+yaml
+        working-directory: monitoring
+
```
JBWatenbergScality: nit: This and the following step are duplicated in `.github/workflows/test.yaml` and `.github/workflows/release.yml`, you may want to create a reusable action with those steps to avoid the duplication. 
Of course this is out of scope of the feature being implemented here, so feel free to do it later on or track it with a new ticket if you want.

monitoring/count-items-cronjob/alerts.yaml
```diff
@@ -0,0 +1,66 @@
+x-inputs:
+  - name: namespace
+    type: constant
+    value: zenko
+  - name: count_items_cronjob
+    type: constant
+    value: artesca-data-ops-count-items
```
JBWatenbergScality, line 7: nit: I wonder if we should not take the `artesca-data` prefix from an env as this is linked to the zenko CR, in practice if the CR is named differently the operator should probably rename this alert. This can also be done with the current implementation and the operator setting this var with `{crName}-{jobName}` directly

monitoring/count-items-cronjob/alerts.yaml
```diff
@@ -0,0 +1,66 @@
...
+                * ON(job_name,namespace) GROUP_RIGHT()
+                kube_job_owner{owner_name="${count_items_cronjob}"}
+              )
+              BY (owner_name),
+              "cronjob", "$1", "owner_name", "(.+)")
+          ) > ${count_items_job_duration_threshold}
+        Labels:
+          severity: critical
+        Annotations:
+          description: "Job {{$labels.job_name}} is taking more than 1h to complete"
```
JBWatenbergScality: nit: for the user to better understands the impacts of this alert it may be interesting to precise them in this description. E.g `Job {{$labels.job_name}} is taking more than 1h to complete. This means that consumption information computation will be delayed and the reported metrics might not be accurate anymore.`

monitoring/count-items-cronjob/alerts.yaml
```diff
@@ -0,0 +1,66 @@
...
+              max(
+                kube_job_status_start_time
+                * ON(job_name,namespace) GROUP_RIGHT()
+                kube_job_owner{owner_name="${count_items_cronjob}"}
+              )
+              BY (owner_name),
+              "cronjob", "$1", "owner_name", "(.+)")
+          ) > ${count_items_job_duration_threshold}
+        Labels:
+          severity: critical
```
JBWatenbergScality: Is this alert level really critical ? It doesn't prevent any service to work, only delays the computation of the metrics. Alert level might need to be confirmed with Florent.

monitoring/count-items-cronjob/alerts.yaml
```diff
@@ -0,0 +1,66 @@
...
+              )
+              BY (owner_name),
+              "cronjob", "$1", "owner_name", "(.+)")
+            , 1)
+          * ON(job_name) GROUP_LEFT()
+          (kube_job_status_failed > 0)
+        Labels:
+          severity: warning
+        Annotations:
+          description: "Job {{$labels.job_name}} failed to complete"
```
JBWatenbergScality: nit: I would also add a user friendly description of the impacts here

monitoring/update-bucket-capacity-info-cronjob/alerts.yaml
```diff
@@ -0,0 +1,99 @@
...
+              BY (owner_name),
+              "cronjob", "$1", "owner_name", "(.+)")
+          ) > ${update_bucket_capacity_info_success_job_existence_duration_threshold}
+        labels:
+          severity: critical
+        annotations:
+          description: |
+            No successful {{$labels.cronjob}}
+            job run in the last 10m, will cause bucket capacity to be out of date
+            and Veeam to fail to process.
```
JBWatenbergScality: nit: I feel this formulation to be a bit too vague. Maybe we can be a bit more precise on Veeam impacts ?

monitoring/update-bucket-capacity-info-cronjob/alerts.yaml
```diff
@@ -0,0 +1,99 @@
...
+                * ON(job_name,namespace) GROUP_RIGHT()
+                kube_job_owner{owner_name="${update_bucket_capacity_info_cronjob}"}
+              )
+              BY (owner_name),
+              "cronjob", "$1", "owner_name", "(.*)")
+          ) > ${update_bucket_capacity_info_job_duration_threshold}
+        labels:
+          severity: critical
+        annotations:
+          description: "Job {{$labels.job_name}} is taking more than 4m to complete"
```
JBWatenbergScality: nit: I would add a description on the impacts. (un-accuracy of metrics and Veeam integration at risk)

monitoring/count-items-cronjob/alerts.yaml
XinLi0207: hey JB @JBWatenbergScality, that's a good point, do you know how can we get the Zenko CR name from Prometheuse?

JBWatenbergScality: I don't think taking the name of the CR from prometheus would be the right way even through it could be possible. I would more likely say that the operator should change these alert definitions based on the CR it is reconciling.

welansari: Since you'll have to deploy the S3Utils alerts anyway in ZKOP you'll be able to pass all the inputs from there. [This](https://github.com/scality/zenko-operator/blob/c392b853d0dd16379062faa316c94dce4387cdfc/pkg/controller/zenko/reconcile_backbeat_alerting.go#L35-L67) is an example of the inputs we pass to the backbeat alerts.

The CR name and namespace is passed by default to the alerts in ZKOP [here](https://github.com/scality/zenko-operator/blob/c392b853d0dd16379062faa316c94dce4387cdfc/pkg/controller/zenko/alert_generator.go#L76-L86), you'll just have to rename the inputs to conform to what we have in ZKOP.

You can also pass `count_items_cronjob` as you already have the value [here](https://github.com/scality/zenko-operator/blob/c392b853d0dd16379062faa316c94dce4387cdfc/pkg/controller/zenko/object_builder.go#L65), and `count_items_job_duration_threshold` should probably be configurable at Zenko CR level, but this is to be done in the ZKOP PR.

welansari: Thinking about it, you probably don't even need to construct the `owner_name` by hand in the alerts by concatenating `zenkoCRName` at the beginning, there is already a function in ZKOP for that and that we use to get the job names for the alerts.
https://github.com/scality/zenko-operator/blob/c392b853d0dd16379062faa316c94dce4387cdfc/pkg/controller/zenko/object_builder.go#L361-L363

monitoring/count-items-cronjob/alerts.yaml
```diff
@@ -0,0 +1,74 @@
...
+              max(
+                kube_job_status_start_time
+                * ON(job_name,namespace) GROUP_RIGHT()
+                kube_job_owner{owner_name="${zenkoCRName}-${count_items_cronjob}"}
+              )
+              BY (owner_name),
+              "cronjob", "$1", "owner_name", "(.+)")
+            , 1)
+          * ON(job_name) GROUP_LEFT()
+          (kube_job_status_failed > 0)
```
welansari: This can be simplified using the `kube_job_status_failed` metric, it's basically a gauge that gives you the number of failed jobs. You can select the job using regex, it should look something like `kube_job_status_failed{job_name=~"artesca-data-ops-count-items-.*"}`.

One thing to keep in mind is that we already have alerts for failed jobs, so you might want to check if this alert is needed.

monitoring/count-items-cronjob/alerts.yaml
```diff
@@ -0,0 +1,74 @@
...
+              BY (job_name, owner_name, namespace)
+              == ON(owner_name) GROUP_LEFT()
+              max(
+                kube_job_status_start_time
+                * ON(job_name,namespace) GROUP_RIGHT()
+                kube_job_owner{owner_name="${zenkoCRName}-${count_items_cronjob}"}
+              )
+              BY (owner_name),
+              "cronjob", "$1", "owner_name", "(.+)")
+          ) > ${count_items_job_duration_threshold}
```
welansari: This could be simplified by using the metrics `kube_job_status_completion_time` for getting the completion time of a job and `kube_job_status_start_time` to get the start time of a job.

The alert expression should look something like the following:
```
sum by(job_name) (kube_job_status_completion_time{job_name=~"artesca-data-ops-count-items-.*"})
- sum by(job_name) (kube_job_status_start_time{job_name=~"artesca-data-ops-count-items-.*"})
> ${count_items_job_duration_threshold}
```

You can select the group of jobs named `artesca-data-ops-count-items-XXX` using `job_name=~"artesca-data-ops-count-items-.*"`

monitoring/count-items-cronjob/alerts.yaml
XinLi0207: cool, thanks for letting me know! seems the fail check not needed anymore and I enabled `removeFailedJobsOnSuccess` on those cronjobs, it can be alerted by the default rule already

monitoring/count-items-cronjob/alerts.yaml
XinLi0207: what if a job failed, will it appear in the `kube_job_status_completion_time`?

.github/workflows/release.yml
```diff
@@ -11,36 +11,25 @@ on:
...
+    name: Create GitHub Release
     runs-on: ubuntu-latest
+    needs:
+      - build
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
```
welansari, line 22: Just curious, why did we remove this step in `.github/workflows/build.yml` ?

XinLi0207: oh, it's a mistake! I should copy it to `build.yml` file
thanks a lot Walid!

monitoring/update-bucket-capacity-info-cronjob/alerts.yaml
```diff
@@ -0,0 +1,44 @@
...
+    value: 600
+
+groups:
+  - name: update-bucket-capacity-info-cronjob/alerts.rules
+    rules:
+      - alert: UpdateBucketCapacityJobTakingTooLong
+        expr: |
+          sum by(job_name) (kube_job_status_completion_time{job_name=~"${update_bucket_capacity_info_cronjob}.*"})
+          - sum by(job_name) (kube_job_status_start_time{job_name=~"${update_bucket_capacity_info_cronjob}.*"})
+          > ${update_bucket_capacity_info_job_duration_threshold}
```
welansari: This works, but it waits for the job to finish before triggering an alert. I'm not sure if this is the desired behaviour ?

One idea is to use the following :
```
time() - 
(sum by(job_name) (kube_job_status_failed{job_name=~"${update_bucket_capacity_info_cronjob}.*"})
> sum by(job_name) (kube_job_status_completion_time{job_name=~"${update_bucket_capacity_info_cronjob}.*"})
or sum by(job_name) (kube_job_status_completion_time{job_name=~"${update_bucket_capacity_info_cronjob}.*"}))
> ${update_bucket_capacity_info_job_duration_threshold}
```

This works because ZKOP forbids having multiple concurrent runs of the same job [here](https://github.com/scality/zenko-operator/blob/c392b853d0dd16379062faa316c94dce4387cdfc/pkg/controller/zenko/cronjob_generator.go#L39), this means that if it is time for a new job run and the previous job run hasn't finished yet, the CronJob skips the new job run. Which means that all you have to do is check the time difference between the last job run (that either failed or succeeded) and the current time.

To get the maximum value of two metrics you use the expression :
```

a > b or b
```
Where `a` and `b` are metrics
if `a>b` is true `a` will be returned else `b` is returned

XinLi0207: great! that's better, I updated the count-items one as well

welansari (approved):

monitoring/count-items-cronjob/alerts.yaml
```diff
@@ -0,0 +1,29 @@
...
+          or sum by(job_name) (kube_job_status_completion_time{job_name=~"${count_items_cronjob}.*"}))
+          > ${count_items_job_duration_threshold}
+        Labels:
+          severity: warning
+        Annotations:
+          description: |
+            Job ${count_items_cronjob} is taking more than 1h to complete.
+            This means that consumption information will be delayed
+            and the reported metrics might not be accurate anymore.
+          summary: count-items cronjob didn't finish after 1h
```
ghost: As we can change the frequency of this job (this will be documented very soon for the veeam use case, with warnings of course), I would suggest, if possible, to have a dynamic  `1h` based on the current frequency of the cronjob. If changing the string is implying too much work, we could rewrite it not to mention the duration?

monitoring/update-bucket-capacity-info-cronjob/alerts.test.yaml
```diff
@@ -0,0 +1,53 @@
...
+        exp_alerts: []
+      - alertname: UpdateBucketCapacityJobTakingTooLong
+        eval_time: 9m
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              job_name: artesca-data-ops-update-bucket-capacity-info-1
+            exp_annotations:
+              description: |
+                Job artesca-data-ops-update-bucket-capacity-info is taking more than 4m to complete.
```
ghost: Same comment here, considering that the doc example it changing to 5min (and many readers may blindly run this command)

monitoring/count-items-cronjob/alerts.yaml
XinLi0207: yeah good idea! I will address that

francoisferrand: it should be possible to use the variables (count_items_job_duration_threshold) in the description ;
and I am not sure, but it _may_ be possible  to use functions or operations in the template, something like `{{ ${count_items_job_duration_threshold} / 3600 }}`

XinLi0207: seems not working the '/' operator
```
/tests/monitoring/update-bucket-capacity-info-cronjob/alerts.rendered.yaml: group "update-bucket-capacity-info-cronjob/alerts.rules", rule 1, "UpdateBucketCapacityJobTakingTooLong": annotation "description": template: __alert_UpdateBucketCapacityJobTakingTooLong:2: unexpected "/" in operand
```

XinLi0207: or I just leave it `taking more than 3600s to complete`

ghost (approved):
LGTM! thanks for the minor changes

XinLi0207:
ping

PR #283, opened 2023-03-21
Improvement/s3 utils 127 2nd implementation storage metrics

XinLi0207:
the detailed [design](https://github.com/scality/citadel/pull/137) is defined here

the main changes are:
- add location metrics that is used by the account for each account
- split the whole single mongo document into one document per each account/location/bucket for scalability(because mongo has size limit for single document)
- because of the deleted and recreated entity will have stale metrics data issue, use a unique identifier(never be same) to represent each entity: account uses canonicalId, bucket uses bucketName_createionDate, location uses locationId(get from mounted locationConfig)

### DIFF
--- a/DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
+++ b/DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
@@ -44,23 +44,7 @@ function isValidCapacityValue(capacity) {
 }
 
 function collectBucketMetricsAndUpdateBucketCapacityInfo(mongoClient, log, callback) {
-    let bucketMetrics;
     return async.waterfall([
-        next => mongoClient.readCountItems(log, (err, doc) => {
-            if (err) {
-                log.error('an error occurred during readCountItems', {
-                    error: { message: err.message },
-                });
-                return next(err);
-            }
-            if (doc && doc.dataMetrics && doc.dataMetrics.bucket) {
-                bucketMetrics = doc.dataMetrics.bucket;
-            } else {
-                log.info('no buckets metrics in countItems, exit');
-                return callback(null);
-            }
-            return next();
-        }),
         next => mongoClient.getBucketInfos(log, (err, bucketInfos) => {
             if (err) {
                 log.error('error getting bucket list', {
@@ -81,37 +65,43 @@ function collectBucketMetricsAndUpdateBucketCapacityInfo(mongoClient, log, callb
 
                 // get bucket storage used
                 let bucketStorageUsed = -1;
-                if (isValidBucketStorageMetrics(bucketMetrics[bucketName])) {
-                    bucketStorageUsed = bucketMetrics[bucketName].usedCapacity.current
-                            + bucketMetrics[bucketName].usedCapacity.nonCurrent;
-                }
-
-                // read Capacity from bucket._capabilities
-                const { Capacity } = bucket.getCapabilities().VeeamSOSApi.CapacityInfo;
-
-                let available = -1;
-                let capacity = -1;
-                if (isValidCapacityValue(Capacity)) { // is Capacity value is valid
-                    capacity = Capacity;
-                    // if bucket storage used is valid and capacity is bigger than used
-                    if (bucketStorageUsed !== -1 && (capacity - bucketStorageUsed) >= 0) {
-                        available = capacity - bucketStorageUsed;
-                    }
-                }
-
-                return mongoClient.updateBucketCapacityInfo(bucketName, {
-                    Capacity: capacity,
-                    Available: available,
-                    Used: bucketStorageUsed,
-                }, log, err => {
-                    if (err) {
-                        log.error('an error occurred during put bucket CapacityInfo attributes', {
+                return mongoClient.readStorageConsumptionMetrics(`bucket_${bucketName}_${new Date(bucket.getCreationDate()).getTime()}`, log, (err, doc) => {
+                    if (err && err.message !== 'NoSuchEntity') {
+                        log.error('an error occurred during readStorageConsumptionMetrics', {
                             error: { message: err.message },
-                            bucketName,
                         });
                         return cb(err);
                     }
-                    return cb();
+                    if (isValidBucketStorageMetrics(doc)) {
+                        bucketStorageUsed = doc.usedCapacity.current
+                            + doc.usedCapacity.nonCurrent;
+                    }
+                    // read Capacity from bucket._capabilities
+                    const { Capacity } = bucket.getCapabilities().VeeamSOSApi.CapacityInfo;
+
+                    let available = -1;
+                    let capacity = -1;
+                    if (isValidCapacityValue(Capacity)) { // is Capacity value is valid
+                        capacity = Capacity;
+                        // if bucket storage used is valid and capacity is bigger than used
+                        if (bucketStorageUsed !== -1 && (capacity - bucketStorageUsed) >= 0) {
+                            available = capacity - bucketStorageUsed;
+                        }
+                    }
+
+                    return mongoClient.updateBucketCapacityInfo(bucketName, {
+                        Capacity: capacity,
+                        Available: available,
+                        Used: bucketStorageUsed,
+                    }, log, err => {
+                        if (err) {
+                            log.error('an error occurred during put bucket CapacityInfo attributes', {
+                                error: { message: err.message },
+                                bucketName,
+                            });
+                        }
+                        return cb(err);
+                    });
                 });
             },
             err => {
--- a/utils/S3UtilsMongoClient.js
+++ b/utils/S3UtilsMongoClient.js
@@ -1,8 +1,15 @@
 const { MongoClientInterface } = require('arsenal').storage.metadata.mongoclient;
+const { Long } = require('mongodb');
 const { errors } = require('arsenal');
+const async = require('async');
 const { validStorageMetricLevels } = require('../CountItems/utils/constants');
+const getLocationConfig = require('./locationConfig');
 
 const METASTORE = '__metastore';
+const INFOSTORE = '__infostore';
+const INFOSTORE_TMP = `${INFOSTORE}_tmp`;
+const __COUNT_ITEMS = 'countitems';
+
 
 class S3UtilsMongoClient extends MongoClientInterface {
     getObjectMDStats(bucketName, bucketInfo, isTransient, log, callback) {
@@ -15,9 +22,10 @@ class S3UtilsMongoClient extends MongoClientInterface {
                 'value.dataStoreName': 1,
                 'value.content-length': 1,
                 'value.versionId': 1,
-                'value.owner-display-name': 1,
+                'value.owner-id': 1,
                 'value.isDeleteMarker': 1,
                 'value.isNull': 1,
+                'value.archive': 1,
             },
         });
         const collRes = {
@@ -29,9 +37,11 @@ class S3UtilsMongoClient extends MongoClientInterface {
         const cmpDate = new Date();
         cmpDate.setHours(cmpDate.getHours() - 1);
 
+        const locationConfig = getLocationConfig(log);
+
         cursor.forEach(
             res => {
-                const { data, error } = this._processEntryData(bucketName, res, isTransient);
+                const { data, error } = this._processEntryData(bucketName, bucketInfo, res, isTransient, locationConfig);
 
                 if (error) {
                     log.error('Failed to process entry data', {
@@ -87,6 +97,28 @@ class S3UtilsMongoClient extends MongoClientInterface {
                         });
                     }
                 });
+                Object.keys(data.account).forEach(account => {
+                    if (!collRes.account[account].locations) {
+                        collRes.account[account].locations = {};
+                    }
+
+                    Object.keys(data.location).forEach(location => {
+                        if (!collRes.account[account].locations[location]) {
+                            collRes.account[account].locations[location] = {
+                                masterCount: 0,
+                                masterData: 0,
+                                nullCount: 0,
+                                nullData: 0,
+                                versionCount: 0,
+                                versionData: 0,
+                                deleteMarkerCount: 0,
+                            };
+                        }
+                        collRes.account[account].locations[location][targetData] += data.location[location];
+                        collRes.account[account].locations[location][targetCount]++;
+                        collRes.account[account].locations[location].deleteMarkerCount += res.value.isDeleteMarker ? 1 : 0;
+                    });
+                });
             },
             err => {
                 if (err) {
@@ -108,13 +140,15 @@ class S3UtilsMongoClient extends MongoClientInterface {
 
     /**
      * @param{string} bucketName -
+     * @param{object} bucketInfo - bucket attributes
      * @param{object} entry -
      * @param{string} entry._id -
      * @param{object} entry.value -
      * @param{boolean} isTransient -
+     * @param{object} locationConfig - locationConfig.json
      * @returns{object.<string, number>} results -
      */
-    _processEntryData(bucketName, entry, isTransient) {
+    _processEntryData(bucketName, bucketInfo, entry, isTransient, locationConfig) {
         if (!bucketName) {
             return {
                 data: {},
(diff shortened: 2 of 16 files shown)


utils/S3UtilsMongoClient.js
```diff
@@ -269,6 +374,68 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+            next => this.db.createCollection(INFOSTORE_TMP, (err, collection) => {
+                if (err) {
+                    return next(err);
+                }
+                tempCollection = collection;
+                return next();
+            }),
+            next => tempCollection.insertMany(updatedStorageMetricsList, next),
+            next => tempCollection.rename(INFOSTORE, { dropTarget: true }, next),
+        ], err => {
```
ghost: We can set `ordered` to false here, during the insertMany, to improve performance (see https://www.mongodb.com/docs/v4.2/reference/method/db.collection.insertMany/), as it can greatly reduce the impact of the query for a large number of documents.

utils/locationConfig.js
```diff
@@ -0,0 +1,17 @@
+const fs = require('fs');
+
+function getLocationConfig(locationConfigFile) {
+    if (!locationConfigFile) {
+        // eslint-disable-next-line no-param-reassign
+        locationConfigFile = process.env.LOCATION_CONFIG_FILE || 'conf/locationConfig.json';
+    }
```
ghost: A potential way to avoid the eslint-disable

```suggestion
function getLocationConfig(locationConfigFile) {
    const locationConfiguration = locationConfigFile ?? locationConfigFile : process.env.LOCATION_CONFIG_FILE || 'conf/locationConfig.json';
```

utils/locationConfig.js
```diff
@@ -0,0 +1,17 @@
...
+
+    if (!fs.existsSync(locationConfigFile)) {
+        return null;
+    }
+
+    const buf = fs.readFileSync(locationConfigFile);
+    return JSON.parse(buf.toString());
```
ghost: In case of an invalid location configuration I think we should not crash, but log it and return null, what do you think? `JSON.parse` can be wrapped inside a try/catch.

I think so because if we return null, then you already handle this case https://github.com/scality/s3utils/pull/283/files#diff-3a778b361ddfba6106347ff0cb053bcbb1a84374fa0d7e6e6413e7c6fb6ca5cdR166, so this way the function will return consistent results.
Alternativaly we could throw an error in these cases directly here.

utils/S3UtilsMongoClient.js
```diff
@@ -269,6 +374,68 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                    error: err.message,
+                });
+                return cb(errors.InternalError);
+            }
+            if (!doc) {
+                return cb(errors.NoSuchEntity);
+            }
+            return cb(null, doc.value);
+        });
+    }
```
ghost: The only drawback of the TMP collection is that if a request from this function is open during the renaming, it will be cancelled. Should we handle this specific case with a retry pattern?

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -81,37 +65,44 @@ function collectBucketMetricsAndUpdateBucketCapacityInfo(mongoClient, log, callb
...
-                }
-
-                return mongoClient.updateBucketCapacityInfo(bucketName, {
-                    Capacity: capacity,
-                    Available: available,
-                    Used: bucketStorageUsed,
-                }, log, err => {
-                    if (err) {
-                        log.error('an error occurred during put bucket CapacityInfo attributes', {
+                return mongoClient.readStorageConsumptionMetrics(`bucket:${bucketName}_${new Date(bucket.getCreationDate()).getTime()}`, log, (err, doc) => {
```
ghost: We want to make sure the timezone offset is returned in the date object, otherwise `.getTime()` will return a value dependent on the local timezone. As this will be dynamically computed by the UI side, it might not work, if the date is not in UTC mode.

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -81,37 +65,44 @@ function collectBucketMetricsAndUpdateBucketCapacityInfo(mongoClient, log, callb
...
+                        Used: bucketStorageUsed,
+                    }, log, err => {
+                        if (err) {
+                            log.error('an error occurred during put bucket CapacityInfo attributes', {
+                                error: { message: err.message },
+                                bucketName,
+                            });
+                            return cb(err);
+                        }
+                        return cb();
```
ghost: Nit:

```suggestion
                        if (err) {
                            log.error('an error occurred during put bucket CapacityInfo attributes', {
                                error: { message: err.message },
                                bucketName,
                            });
                        }
                        return cb(err);
```

ghost: Nit:

```suggestion
                        if (err) {
                            log.error('an error occurred during put bucket CapacityInfo attributes', {
                                error: { message: err.message },
                                bucketName,
                            });
                        }
                        return cb(err);
```

utils/S3UtilsMongoClient.js
```diff
@@ -150,6 +190,39 @@ class S3UtilsMongoClient extends MongoClientInterface {
                 results.location[rep.site] = size;
             }
         });
+
+        // count in both dataStoreName and cold location if object is restored
+        if (entry.value.archive
+            && entry.value.archive.restoreCompletedAt <= Date.now()
+            && entry.value.archive.restoreWillExpireAt >= Date.now()) {
+            entry.value.location.forEach(location => {
```
ghost: I think if we are exactly at the expiration time, it should be excluded from the check, as the object is considered expired

```suggestion
        if (entry.value.archive
            && entry.value.archive.restoreCompletedAt <= Date.now()
            && entry.value.archive.restoreWillExpireAt > Date.now()) {
            entry.value.location.forEach(location => {
```

utils/S3UtilsMongoClient.js
```diff
@@ -231,6 +304,38 @@ class S3UtilsMongoClient extends MongoClientInterface {
                 });
             }
         });
+
+        // parse all locations and reflect the data in the account
+        Object.keys((res.account || {})).forEach(account => {
+            if (!dataMetrics.account[account].locations) {
+                dataMetrics.account[account].locations = {};
+            }
+            Object.keys(res.location).forEach(location => {
```
ghost: For consistency

```suggestion
        Object.keys((res.account || {})).forEach(account => {
            if (!dataMetrics.account[account].locations) {
                dataMetrics.account[account].locations = {};
            }
            Object.keys(res.location || {}).forEach(location => {
```

utils/S3UtilsMongoClient.js
ghost: Also, should be better handle error in the `next => tempCollection.rename(INFOSTORE, { dropTarget: true }, next)` case? If the renaming failed, we just want to make sure to properly log it, as the metrics won't be updated. Or we could also retry. If the query failed due to a transient error, it's a waste of resources not to retry to apply them and loose everything by crashing the job.

ghost (commented):
1st iteration, gloabally it implements well the design!

KazToozs (commented):
First iteration: first 2 commits look ok to me

utils/S3UtilsMongoClient.js
ghost, line 1: Adding another quick comment on the issue we discussed in the design, about being abit more robust if a bucket collection exists but is not in the metastore collection. Can we handle this case? As disucsseds, the job should be able to start in this case, and ignore the "bad buckets"

utils/locationConfig.js
XinLi0207: I agree with catch the json parse error and return null and let [this logic](https://github.com/scality/s3utils/pull/283/files#diff-3a778b361ddfba6106347ff0cb053bcbb1a84374fa0d7e6e6413e7c6fb6ca5cdR166) to catch it

utils/S3UtilsMongoClient.js
XinLi0207: yeah both renaming and read operation could not succeed because of race condition, I added retry logic for both of them

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
XinLi0207: yeah getTime() is the utc timestamp

utils/S3UtilsMongoClient.js
XinLi0207: yeah! added in this [commit](https://github.com/scality/s3utils/pull/283/commits/2546e24005b09596242e469adf0305343d0fd542)

KazToozs (approved):
Took a long time but I honestly didn't come up with any issues, LGTM

utils/S3UtilsMongoClient.js
```diff
@@ -269,6 +374,138 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+        ], err => {
+            if (err) {
+                log.error('updateStorageConsumptionMetrics: error updating count items', {
+                    error: err.message,
+                });
+                return cb(errors.InternalError);
+            }
+            return cb();
+        });
+    }
```
ghost: I tested the current count-items job in my cluster where I have thousands of buckets, but this logic:
- puts all the buckets name under `_id: countitems` (`bucketList`). I think this can be deleted, unless we need to expose the list of all buckets. Note that this may fail if the list is too large, as everything is in a single document.
- It only creates one document of each type. See below

Here is my (truncated) output:

```json
{ "_id" : "countitems", "value" : { "objects" : 1057057, "versions" : 0, "buckets" : 1911, "bucketList" : [ { "name" : "my-bucket-2019", "location" : "us-east-1", "isVersioned" : false, "ownerCanonicalId" : "7a9747d06c80a1f046fdd134522699213df25632a7de36b55ee78d069e2425fd", "ingestion" : false }, { "name" : "test", "location" : "us-east-1", "isVersioned" : false, "ownerCanonicalId" : "7a9747d06c80a1f046fdd134522699213df25632a7de36b55ee78d069e2425fd", "ingestion" : false }, { "name" : "my-bucket-1162", "location" : "us-east-1", "isVersioned" : false, "ownerCanonicalId" : "7a9747d06c80a1f046fdd134522699213df25632a7de36b55ee78d069e2425fd", "ingestion" : false }, { "name" : "my-bucket-1246", "location" : "us-east-1", "isVersioned" : false, "ownerCanonicalId" : "7a9747d06c80a1f046fdd134522699213df25632a7de36b55ee78d069e2425fd", "ingestion" : false }, { "name" : "my-bucket-1192", "loca.... [ all the buckets here) ],
{ "_id" : "location:fd2234c7-c258-11ed-aae3-ea8ec16f01ad", "value" : { "measuredOn" : ISODate("2023-03-31T06:55:20.786Z"), "usedCapacity" : { "current" : 0, "nonCurrent" : 0 }, "objectCount" : { "current" : 1057057, "nonCurrent" : 0, "deleteMarker" : 0 } } }
{ "_id" : "account:7a9747d06c80a1f046fdd134522699213df25632a7de36b55ee78d069e2425fd", "value" : { "measuredOn" : ISODate("2023-03-31T06:55:20.786Z"), "usedCapacity" : { "current" : 0, "nonCurrent" : 0 }, "objectCount" : { "current" : 1057057, "nonCurrent" : 0, "deleteMarker" : 0 }, "locations" : { "fd2234c7-c258-11ed-aae3-ea8ec16f01ad" : { "usedCapacity" : { "current" : 0, "nonCurrent" : 0 }, "objectCount" : { "current" : 1057057, "nonCurrent" : 0, "deleteMarker" : 0 } } } } }
{ "_id" : "bucket:test_1680008330869", "value" : { "measuredOn" : ISODate("2023-03-31T06:55:20.786Z"), "usedCapacity" : { "current" : 0, "nonCurrent" : 0 }, "objectCount" : { "current" : 1057057, "nonCurrent" : 0, "deleteMarker" : 0 } } }
```

```bash
data-db-mongodb-sharded-shard-0:PRIMARY> db.getCollection('__infostore').find().count()
4
```

For some reason I see only one document of each type (like, only 1 `bucket:test_...`) while I have many others, I believe this is not on purpose.

utils/S3UtilsMongoClient.js
```diff
@@ -269,6 +374,138 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
     }
+
+    updateStorageConsumptionMetrics(countItems, dataMetrics, log, cb) {
+        const updatedStorageMetricsList = [{ _id: __COUNT_ITEMS, value: countItems }];
+        // iterate every resource through dataMetrics and add to updatedStorageMetricsList
+        Object.keys(dataMetrics).forEach(metricLevel => {
+            if (validStorageMetricLevels.has(metricLevel)) {
+                Object.keys(dataMetrics[metricLevel]).forEach(resource => {
+                    updatedStorageMetricsList.push({
+                        _id: `${metricLevel}:${resource}`,
```
ghost:
```suggestion
                        _id: `${metricLevel}_${resource}`,
```

for consistency with the design (unless you found an issue of using `_`)

utils/S3UtilsMongoClient.js
ghost: After adding some logs this logic seem fine but the function never receives all the resources for some reason

```
size of datametrics {
  bucket: {
    test_1680008330869: { usedCapacity: [Object], objectCount: [Object] }
  },
  location: {
    'fd2234c7-c258-11ed-aae3-ea8ec16f01ad': { usedCapacity: [Object], objectCount: [Object] }
  },
  account: {
    '7a9747d06c80a1f046fdd134522699213df25632a7de36b55ee78d069e2425fd': {
      usedCapacity: [Object],
      objectCount: [Object],
      locations: [Object]
    }
  }
}
size of dataMetrics[metricLevel] {
  test_1680008330869: {
    usedCapacity: { current: 0, nonCurrent: 0 },
    objectCount: { current: 1057057, nonCurrent: 0, deleteMarker: 0 }
  }
}
size of dataMetrics[metricLevel][resource] {
  test_1680008330869: {
    usedCapacity: { current: 0, nonCurrent: 0 },
    objectCount: { current: 1057057, nonCurrent: 0, deleteMarker: 0 }
  }
}
size of dataMetrics[metricLevel] {
  'fd2234c7-c258-11ed-aae3-ea8ec16f01ad': {
    usedCapacity: { current: 0, nonCurrent: 0 },
    objectCount: { current: 1057057, nonCurrent: 0, deleteMarker: 0 }
  }
}
size of dataMetrics[metricLevel][resource] {
  'fd2234c7-c258-11ed-aae3-ea8ec16f01ad': {
    usedCapacity: { current: 0, nonCurrent: 0 },
    objectCount: { current: 1057057, nonCurrent: 0, deleteMarker: 0 }
  }
}
size of dataMetrics[metricLevel] {
  '7a9747d06c80a1f046fdd134522699213df25632a7de36b55ee78d069e2425fd': {
    usedCapacity: { current: 0, nonCurrent: 0 },
    objectCount: { current: 1057057, nonCurrent: 0, deleteMarker: 0 },
    locations: { 'fd2234c7-c258-11ed-aae3-ea8ec16f01ad': [Object] }
  }
}
size of dataMetrics[metricLevel][resource] {
  '7a9747d06c80a1f046fdd134522699213df25632a7de36b55ee78d069e2425fd': {
    usedCapacity: { current: 0, nonCurrent: 0 },
    objectCount: { current: 1057057, nonCurrent: 0, deleteMarker: 0 },
    locations: { 'fd2234c7-c258-11ed-aae3-ea8ec16f01ad': [Object] }
  }
}
going to put 4 [
  {
    _id: 'countitems',
    value: {
      objects: 1057057,
      versions: 0,
      buckets: 1911,
      bucketList: [Array],
      dataManaged: [Object],
      stalled: 0
    }
  },
  {
    _id: 'bucket:test_1680008330869',
    value: {
      measuredOn: 2023-03-31T07:17:12.556Z,
      usedCapacity: [Object],
      objectCount: [Object]
    }
  },
  {
    _id: 'location:fd2234c7-c258-11ed-aae3-ea8ec16f01ad',
    value: {
      measuredOn: 2023-03-31T07:17:12.557Z,
      usedCapacity: [Object],
      objectCount: [Object]
    }
  },
  {
    _id: 'account:7a9747d06c80a1f046fdd134522699213df25632a7de36b55ee78d069e2425fd',
    value: {
      measuredOn: 2023-03-31T07:17:12.558Z,
      usedCapacity: [Object],
      objectCount: [Object],
      locations: [Object]
    }
  }
]
```

image

ghost (changes requested):

utils/S3UtilsMongoClient.js
XinLi0207: for the the existing bucketList in _countitems document
seems it is used by [pensieve-api](https://github.com/scality/pensieve-api/blob/development/1.4/apiv1/restapi/status/item_counts_stats.go), but just I don't know where they are presented on the UI

ghost: Ok, it makes sense then, closing this comment!

utils/S3UtilsMongoClient.js
ghost: After further testing we confirmed a bucket must have objects inside to appear in the __infostore collection.
So we created 5700 buckets, all with objects, and two buckets with millions of objects.

The job ran just fine, in 34s.

ghost: Closing this comment

ghost (approved):

PR #286, opened 2023-05-11
Improvement/s3 utils 131 correct creation date from userbuckets

XinLi0207:
use bucket's creationDate from `__usersbucket` collection instead of `__metastore` to be aligned with bucket's creationDate used by UI

### DIFF
--- a/tests/functional/collectBucketMetricsAndUpdateBucketCapacityInfo.js
+++ b/tests/functional/collectBucketMetricsAndUpdateBucketCapacityInfo.js
@@ -1,13 +1,15 @@
 const async = require('async');
 const werelogs = require('werelogs');
 const assert = require('assert');
+const { constants } = require('arsenal');
 const S3UtilsMongoClient = require('../../utils/S3UtilsMongoClient');
-const { testBucketMD, testBucketCreationDate } = require('../constants');
+const { testBucketMD, testBucketCreationDate, testUserBucketInfo } = require('../constants');
 const { collectBucketMetricsAndUpdateBucketCapacityInfo } = require('../../DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo');
 
 const logger = new werelogs.Logger('collectBucketMetricsAndUpdateBucketCapacityInfo::Test::Functional');
 const { MONGODB_REPLICASET } = process.env;
 const dbName = 'collectBucketMetricsAndUpdateBucketCapacityInfoTest';
+const USERSBUCKET = '__usersbucket';
 
 
 describe('collectBucketMetricsAndUpdateBucketCapacityInfo', () => {
@@ -76,6 +78,17 @@ describe('collectBucketMetricsAndUpdateBucketCapacityInfo', () => {
             next => client.createBucket(testBucketName, {
                 ...testBucketMD,
             }, logger, next),
+            next => client.putObject(
+                USERSBUCKET,
+                `${testBucketMD._owner}${constants.splitter}${testBucketName}`,
+                testUserBucketInfo.value,
+                {
+                    versioning: false,
+                    versionId: null,
+                },
+                logger,
+                next,
+            ), // put bucket entry in __usersbucket
             next => collectBucketMetricsAndUpdateBucketCapacityInfo(client, logger, err => {
                 assert.equal(err, null);
                 next();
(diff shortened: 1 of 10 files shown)


JBWatenbergScality:
I just tried the dev image on my ptf, ran the job manually, refreshed the UI, and ... 
image

the magic occurred. Thanks for the quick fix !! Feature wise it works ! I'm not sure to have time to review the implementation details through this afternoon :/

XinLi0207:
> I just tried the dev image on my ptf, ran the job manually, refreshed the UI, and ... image
> 
> the magic occurred. Thanks for the quick fix !! Feature wise it works ! I'm not sure to have time to review the implementation details through this afternoon :/

thanks @JBWatenbergScality for the quick testing!
no metrics for the other three buckets is that because of no objects?

JBWatenbergScality:
Yes others buckets are empty !

tests/functional/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -76,6 +77,17 @@ describe('collectBucketMetricsAndUpdateBucketCapacityInfo', () => {
             next => client.createBucket(testBucketName, {
                 ...testBucketMD,
             }, logger, next),
+            next => client.putObject(
+                USERSBUCKET,
+                `${testBucketMD._owner}..|..${testBucketName}`,
```
ghost: I believe this is exposed by arsenal in its constants file, if no, it can be better to centralize all these separators in one place in s3utils, in a constant we can reuse.

utils/S3UtilsMongoClient.js
```diff
@@ -39,72 +41,61 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
-                                    nullData: 0,
-                                    versionCount: 0,
-                                    versionData: 0,
-                                    deleteMarkerCount: 0,
-                                };
-                            }
-                            collRes[metricLevel][resourceName][targetData] += data[metricLevel][resourceName];
-                            collRes[metricLevel][resourceName][targetCount]++;
-                            collRes[metricLevel][resourceName].deleteMarkerCount += res.value.isDeleteMarker ? 1 : 0;
+        while (await cursor.hasNext()) {
```
ghost: It could work here to parallelize a bit with `async` the process, to divide the overall latency in caser the number of buckets is high. Note that I do not consider this as a blocker for this PR, but might cause issues with the job duration in the future.

If we can process more buckets at a time to get their "right" creation date, it can be benefical, but it also complexifies the code structure a lot...

ghost (approved):
LGTM, some comments but not blocking this particular PR

Thanks @JBWatenbergScality for the testing!

utils/S3UtilsMongoClient.js
```diff
@@ -170,10 +176,10 @@ class S3UtilsMongoClient extends MongoClientInterface {
                 error: new Error('empty locationConfig'),
             };
         }
-
         const results = {
             // there will be only one bucket for an object entry, and use `bucketName_creationDate` as key
-            bucket: { [`${bucketName}_${new Date(bucketInfo.getCreationDate()).getTime()}`]: size },
+            // creationDate is got from __userbucket collection
```
francoisferrand: nit:
```suggestion
            // creationDate comes from __userbucket collection
```

utils/S3UtilsMongoClient.js
```diff
@@ -39,72 +41,61 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
-                                    versionData: 0,
-                                    deleteMarkerCount: 0,
-                                };
-                            }
-                            collRes[metricLevel][resourceName][targetData] += data[metricLevel][resourceName];
-                            collRes[metricLevel][resourceName][targetCount]++;
-                            collRes[metricLevel][resourceName].deleteMarkerCount += res.value.isDeleteMarker ? 1 : 0;
+        while (await cursor.hasNext()) {
+            const res = await cursor.next();
+            const bucketCreationDate = await (new Promise((resolve, reject) => {
```
francoisferrand: do we need to compute the bucketCreationDate _inside_ the loop: the loop iterates over a single bucket, so it should be enough to read the bucketCreationDate once before going through the cursor ?

e.g. similar to how we get the bucketInfo once at the beginning (actually, as parameter) of this function

XinLi0207: changed to read the usersbucket once to get all buckets' creationDates
https://github.com/scality/s3utils/pull/286/commits/0f66c93a8e3950a55c845009f53ec4ee2311832d

utils/S3UtilsMongoClient.js
XinLi0207: changed to read the usersbucket once to get all buckets' creationDates
https://github.com/scality/s3utils/commit/0f66c93a8e3950a55c845009f53ec4ee2311832d

tests/functional/collectBucketMetricsAndUpdateBucketCapacityInfo.js
XinLi0207: thanks for pointing out! we do have this var in arsenal

anurag4DSB (approved):

PR #290, opened 2023-06-05
S3UTILS-133 compareFollowerDbs

jonathan-gramain:
Add a new tool to compare follower databases offline, and report differences in a JSON output file in the same format than `followerDiff`.

It can be used in place of `followerDiff` to support frequent leader changes, however it has some drawbacks:

- it requires stopping the leader manually at least once

- it requires copying follower's databases so that the script can scan databases from two followers from the same location

### DIFF
--- a/CompareRaftMembers/KeyValueStreamDiff.js
+++ b/CompareRaftMembers/KeyValueStreamDiff.js
@@ -1,0 +1,113 @@
+const async = require('async');
+const stream = require('stream');
+
+const { jsutil } = require('arsenal');
+
+/**
+ * Output differences between two streams, where each stream emits
+ * { key: string, value: string } objects sorted by key.
+ *
+ * The output consists of entries of one of the following types:
+ *
+ * - [{ key, value }, null]: this key is present on stream #1 but not
+ *   on stream #2
+ *
+ * - [null, { key, value }]: this key is not present on stream #1 but
+ *   is present on stream #2
+ *
+ * - [{ key, value: "{value1}" }, { key, value: "{value2}" }]: this key
+ *   has a different value between stream #1 and stream #2: "value1"
+ *   is the value seen on stream #1 and "value2" the value seen on
+ *   stream #2.
+ *
+ * @class KeyValueStreamDiff
+ */
+class KeyValueStreamDiff extends stream.Readable {
+    /**
+     * @constructor
+     * @param {stream.Readable} stream1 - first stream to compare
+     * @param {stream.Readable} stream2 - second stream to compare
+     */
+    constructor(stream1, stream2) {
+        super({ objectMode: true });
+
+        this._streams = [stream1, stream2];
+        this._lastItems = [null, null];
+        this._eof = [false, false];
+        this._reading = false;
+
+        this._streams.forEach((stream, streamIndex) => {
+            stream
+                .on('data', data => {
+                    this._lastItems[streamIndex] = data;
+                    stream.pause();
+                    this._checkDiff();
+                })
+                .on('end', () => {
+                    this._eof[streamIndex] = true;
+                    this._checkDiff();
+                })
+                .on('error', err => {
+                    // eslint-disable-next-line no-console
+                    console.error(`error from stream #${streamIndex}: ${err.message}`);
+                    this.destroy(err);
+                });
+        });
+    }
+
+    _read() {
+        if (!this._reading) {
+            this._reading = true;
+            this._checkDiff();
+        }
+    }
+
+    _consumedItem(streamIndex) {
+        this._lastItems[streamIndex] = null;
+        this._streams[streamIndex].resume();
+    }
+
+    _pushDiff(item1, item2) {
+        this.push([item1, item2]);
+        this._reading = false;
+    }
+
+    _checkDiff() {
+        // bail out if:
+        // - consumer not actively reading
+        // - or at least one of the input streams is waiting for an item to read
+        if (!this._reading
+            || (!this._eof[0] && !this._lastItems[0])
+            || (!this._eof[1] && !this._lastItems[1])) {
+            return undefined;
+        }
+        // compute the difference between the last item read from each
+        // stream (or null if EOF was reached)
+        if (this._lastItems[0] && this._lastItems[1]) {
+            if (this._lastItems[0].key < this._lastItems[1].key) {
+                this._pushDiff(this._lastItems[0], null);
+                this._consumedItem(0);
+            } else if (this._lastItems[0].key > this._lastItems[1].key) {
+                this._pushDiff(null, this._lastItems[1]);
+                this._consumedItem(1);
+            } else {
+                if (this._lastItems[0].value !== this._lastItems[1].value) {
+                    this._pushDiff(this._lastItems[0], this._lastItems[1]);
+                }
+                this._consumedItem(0);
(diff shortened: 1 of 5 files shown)


jonathan-gramain:
ping

tests/unit/CompareRaftMembers/KeyValueStreamDiff.js
```diff
@@ -0,0 +1,309 @@
...
+                ],
+                expectedOutput: [
+                    [
+                        null,
+                        { key: 'bucket/key1', value: '{}' },
+                    ],
+                ],
+            },
+            {
+                desc: 'with respectively one and zero entry in each stream',
```
watsaqat, line 53: could you add a note why this is not symmetric w.r.t. to the above case and needs to be tested seperately? Or is this intended as a sanity check asserting stream params can be passed in either order?

README.md
```diff
@@ -858,6 +882,402 @@ summary](#compare-followers-databases-against-leader-view).
...
+  * This configuration would cause diff entries of which bucket/key
+    appear in one of the following to be discarded from the output:
+
+    * oplog of raft session 1 after cseq=1234
+
+    * or oplog of raft session 4 after cseq=4567
+
+    * or oplog of raft session 6 after cseq=6789
+
+    * or any other raft session's oplog at any cseq
```
watsaqat, line 947: if we're excluding all other rs not in the mapping, maybe EXCLUDE_FROM_CSEQS isn't the best naming. I think the mapping acts more like an inclusion set rather than an exclusion set?

README.md
```diff
@@ -858,6 +882,402 @@ summary](#compare-followers-databases-against-leader-view).
...
+
+* **BUCKETD_HOSTPORT**: ip:port of bucketd endpoint, needed when
+  EXCLUDE_FROM_CSEQS is set
```
watsaqat: could you note why this is needed when `EXCLUDE_FROM_CSEQS is set`? Not immediately obvious

README.md
```diff
@@ -858,6 +882,402 @@ summary](#compare-followers-databases-against-leader-view).
...
+leader](#scan-procedure-with-stable-leader) procedure instead.
+
+In order to scan a set of raft sessions to detect discrepancies
+between raft members, you may use the following procedure involving
+the scan tool [compareFollowerDbs](#compare-two-followers-database-sets).
+
+The general idea is to compare sets of databases belonging to two
+different followers at a time.
+
+**Note**: Each repd node (including the current leader) has to be
```
watsaqat, line 967: a link/reference to the stopping and starting commands could be helpful here

CompareRaftMembers/KeyValueStreamDiff.js
```diff
@@ -0,0 +1,114 @@
...
+     */
+    constructor(stream1, stream2) {
+        super({ objectMode: true });
+
+        this._streams = [stream1, stream2];
+        this._lastItems = [null, null];
+        this._eof = [false, false];
+        this._reading = false;
+
+        for (let streamIndex = 0; streamIndex < 2; ++streamIndex) {
```
watsaqat: `this._stream.forEach((stream, streamIndex) => ...`

CompareRaftMembers/compareFollowerDbs.js
```diff
@@ -0,0 +1,272 @@
...
+        format:
+            {"rsId":cseq[,"rsId":cseq...]}
+        Example:
+            {"1":1234,"4":4567,"6":6789}
+        This configuration would cause diff entries which bucket/key
+        appear in either of the following to be discarded from the output:
+        - oplog of raft session 1 after cseq=1234
+        - or oplog of raft session 4 after cseq=4567
+        - or oplog of raft session 6 after cseq=6789
+        - or any other raft session's oplog at any cseq
```
watsaqat, line 66: https://github.com/scality/s3utils/pull/290/files#r1232668423

CompareRaftMembers/compareFollowerDbs.js
```diff
@@ -0,0 +1,272 @@
...
+    BUCKETD_HOSTPORT: ip:port of bucketd endpoint, needed when EXCLUDE_FROM_CSEQS is set
+`;
+
+for (const envVar of [
```
watsaqat, line 70: we should move these env var existence checks into arsenal at some point

something like:
```
checkVarExistence(['DATABASES1', 'DATABASES2'])
```

watsaqat (approved):

README.md
jonathan-gramain: Here the `EXCLUDE_FROM` refers to the exclusion from the computation of everything that happens after each cseq of their respective raft session's oplog (or everything otherwise). I don't think it's the best name ever but I haven't found something better (and it's the same name in the other script variation `followerDiff`, it's also nice to keep them in sync).

README.md
jonathan-gramain: OK will complete the description

README.md
jonathan-gramain: The stop/start commands are fully included in the procedure itself. (e.g. https://github.com/scality/s3utils/pull/290/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R770), here the point is mostly to warn the user that repds will be stopped temporarily, which can have an impact on the platform. I could mention this fact though.

tests/unit/CompareRaftMembers/KeyValueStreamDiff.js
jonathan-gramain: it's a sanity check, the function logic is symmetric although the result itself is different in each case (but symmetric). I don't think it's super useful to add a comment about this but I don't mind adding it

CompareRaftMembers/compareFollowerDbs.js
jonathan-gramain: I wouldn't be fond of executing thinks like `process.exit()` in a library function, as I think it's a bad practice to hide control flow changes except in very specific circumstances. Now why not having the rest of the logic in Arsenal, but it's very small and we would still end up doing `if (checkVars(...)) { process.exit(1) }` at least so I don't think there would be any real benefit (but some overhead in finding out where the function is etc.). It's still a pattern we have regularly in the various scripts I agree but we don't do much more than logging and exiting in that case.

CompareRaftMembers/KeyValueStreamDiff.js
jonathan-gramain: done

CompareRaftMembers/compareFollowerDbs.js
jonathan-gramain: What could make more sense IMHO is using an existing library more often to handle a lot of parameter passing etc. for us and automate the sanity checks. Scuba uses `convict` (https://github.com/scality/scuba/blob/development/1.0/lib/config.ts#L118) which also supports environment variables, maybe we could benefit from it as well in the future (TBH I don't want to tackle this for this internal scan script as I already tested it and I'm lazy :laughing: )

watsaqat (approved):

CompareRaftMembers/compareFollowerDbs.js
tmacro: covict supports cli flags as well

tmacro (approved):

PR #293, opened 2023-07-18
impr: S3UTILS-139 repairDuplicateVersionIds

jonathan-gramain:
New script derived from repairDuplicateVersions to repair object metadata that has a duplicate versionId field, notably due to bug S3C-7861 (CRR with incompatible S3C versions between source and target).

It takes log lines from verifyBucketSproxydKeys as input, and for each line matching a duplicate versionId detection, repairs the affected object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1527,6 +1527,43 @@ we suggest to disable those operations on the clients during the time
 the script is running (or alternatively, disable all writes), to avoid
 any such risk.
 
+
+# Repair object metadata with duplicate 'versionId' field
+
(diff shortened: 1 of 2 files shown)


KazToozs (approved):
cool signal handling for the exiting, not something I'd seen yet in JS. LGTM otherwise

tmacro (approved):

jonathan-gramain:
@KazToozs @tmacro I added an extra commit to improve the safety of the repair, indeed, previously it would be able to repair a master with the original scanned version ID, even if the master key was updated in the meantime.

PR #295, opened 2023-08-18
Fix count-items failures

francoisferrand:
- Properly handle cold objects which are currently restored
- Skip over blank PHD markers
- Do not call callback multiple times on process entry error
- Bump node to 16.20.2 (CVE fixes)

Issue: S3UTILS-141
Issue: S3UTILS-142

### DIFF
--- a/utils/S3UtilsMongoClient.js
+++ b/utils/S3UtilsMongoClient.js
@@ -190,11 +190,11 @@ class S3UtilsMongoClient extends MongoClientInterface {
             return { error: new Error('no bucket name provided') };
         }
 
-        if (entry.value.isPHD && !entry.value.hasOwnProperty('content-length')) {
-            // In some case, PHD are re-created form scratch and do not hold any information
-            // (when deleting a delete marker which is the latest version). This should be very
-            // transient (they should be cleaned after 5 seconds), but should not create any
-            // error.
+        if (entry.value.isPHD) {
+            // PHD are created transiently in place of a master when it is deleted, until
+            // they get replaced with the "new" master. They may either hold no information
+            // (and cannot be processed) or information related to the earlier master (and
+            // thus not correct): so best to just ignore them.
             return {};
         }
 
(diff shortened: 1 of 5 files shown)


utils/S3UtilsMongoClient.js
```diff
@@ -172,29 +183,28 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
     _processEntryData(bucketName, bucketInfo, entry, bucketCreationDate, isTransient, locationConfig) {
         if (!bucketName) {
-            return {
-                data: {},
-                error: new Error('no bucket name provided'),
-            };
+            return { error: new Error('no bucket name provided') };
+        }
+
+        if (entry.value.isPHD && !entry.value.hasOwnProperty('content-length')) {
```
nicolas2bert: PHD-flagged master suggests that the master is no more considered valid. Do we need to exclude all master flagged with PHD to ensure we don't provide incorrect information to users? 

Additionally, the presence of the PHD flag might signify that the data for the old master hasn't been deleted (yet). This could explain the rationale behind retaining it.

francoisferrand: > PHD-flagged master suggests that the master is no more considered valid

It is not valid in the sense that there is indeed no matching version; but there may still be a master (e.g. pointing to another version), we just don't know which one...

I think there is no easy fix, the proper way is to count only the versions [except for non versioned, version suspended...], deduce if there is a master from the list and do not try to substract the size/count of masters from versions at the end of the process as is done now. This shall be done separately though, in [ARTESCA-9329](https://scality.atlassian.net/browse/ARTESCA-9329)

@nicolas2bert what do you think we should do _just now_, as the least incorrect value : ignore all PHDs, or count them as master when they have a length?

nicolas2bert: Both solutions provided are incorrect, but the approach of disregarding all PHD flags appears to be more accurate and easier to comprehend. 
This approach is more accurate because the PHD flag is removed when no version is present, whereas the metadata associated with the PHD flag is consistently unreliable or incorrect.
Additionally, this approach is easier to understand since it focuses on a single property, namely 'isPHD' instead of two: `entry.value.isPHD && !entry.value.hasOwnProperty('content-length')`.

Again, I don't have a strong opinion since both solution return incorrect results.

tmacro (approved):

nicolas2bert (approved):

PR #296, opened 2023-08-21
impr(S3UTILS-145): Add OLDER_THAN and VERBOSE options to bucketVersionStats.js

tmacro:
OLDER_THAN:
Limit counted versions to those with a last-modified older than the given timestamp. Can be given as an ISO 8601 timestamp or as a number of seconds suffixed with `s`.

VERBOSE:
Causes the tool to output a log line for each version counted.

### DIFF
--- a/bucketVersionsStats.js
+++ b/bucketVersionsStats.js
@@ -18,6 +18,8 @@ const { KEY_MARKER } = process.env;
 const { VERSION_ID_MARKER } = process.env;
 const { HTTPS_CA_PATH } = process.env;
 const { HTTPS_NO_VERIFY } = process.env;
+const { OLDER_THAN } = process.env;
+const VERBOSE = !!process.env.VERBOSE;
 const AWS_SDK_REQUEST_RETRIES = 100;
 const AWS_SDK_REQUEST_INITIAL_DELAY_MS = 30;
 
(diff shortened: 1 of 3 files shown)


bucketVersionsStats.js
```diff
@@ -58,6 +62,21 @@ Optional environment variables:
 });
 const s3EndpointIsHttps = ENDPOINT.startsWith('https:');
 
+let _OLDER_THAN_TIMESTAMP;
+if (OLDER_THAN) {
```
jonathan-gramain, line 72: Instead of doing ad-hoc code, it could be more adequate to reuse the [parseOlderThan](https://github.com/scality/s3utils/blob/development/1.13/utils/parseOlderThan.js#L11) function, and extend it (if needed) to parse a number of seconds. This way, we both avoid surprises in users trying different tools with different expectations on OLDER_THAN as well as factorizing the code and making it more easily testable.

Personally I would prefer the spec to be an explicit `42 seconds` (with `42 second` accepted as well) instead of `42s`.

utils/parseOlderThan.js
```diff
@@ -16,6 +16,11 @@ function parseOlderThan(olderThan) {
         cutoff.setDate(cutoff.getDate() - numberOfDays);
         return cutoff;
     }
+    const numberOfSecsMatch = /^([0-9]+) seconds?$/.exec(olderThan);
```
jonathan-gramain, line 19: Can you also add a li'l test [here](https://github.com/scality/s3utils/blob/development/1.14/tests/unit/utils/parseOlderThan.js)? :wink:

jonathan-gramain (approved):

bucketVersionsStats.js
ben-scality: If too complicated, we can just use the timestamp ISO8601

tmacro: ended up with ISO8601 + `<n> days` + `<n> seconds`

rachedbenmustapha (approved):

PR #297, opened 2023-08-26
S3UTILS-145: Bump version

tmacro:


francoisferrand (dismissed):
Why do you want to bump?

The 1.13.24 release is already made, on the exact commit of the bump: so no point making a new release now...

And it is best to bump at the time of release, so that git commit history is consistent with tags (ideally tag is on the bump commit), to avoid confusion and possibly skipping a version.

tmacro:
So a few things seem to be going on. I'm trying to bump 1.14 and just bumped the last digit expecting the minor version to be "14".  I should have checked as we're terrible about updating those sorts of things.

rachedbenmustapha (approved):

jonathan-gramain (approved):

PR #300, opened 2023-10-11
Bugfix/s3 utils 148 mongodb improvements

ghost:
- bump mongodb driver by bumping arsenal version.
- rework logging to display the mongodb reasons along with the error code.
- add periodic logging to better track the progress of the job in case of problems.
- migrate all code (including tests) to the latest mongodb changes (i.e. callbacks are removed and replaced by promises, cursors are not readable anymore by default).

### DIFF
--- a/utils/S3UtilsMongoClient.js
+++ b/utils/S3UtilsMongoClient.js
@@ -14,51 +14,59 @@ const __COUNT_ITEMS = 'countitems';
 
 
 class S3UtilsMongoClient extends MongoClientInterface {
-    getObjectMDStats(bucketName, bucketInfo, isTransient, log, callback) {
-        const c = this.getCollection(bucketName);
-        const cursor = c.find({}, {
-            projection: {
-                '_id': 1,
-                'value.last-modified': 1,
-                'value.replicationInfo': 1,
-                'value.dataStoreName': 1,
-                'value.content-length': 1,
-                'value.versionId': 1,
-                'value.owner-id': 1,
-                'value.isDeleteMarker': 1,
-                'value.isNull': 1,
-                'value.archive': 1,
-                'value.x-amz-storage-class': 1,
-                'value.isPHD': 1,
-            },
-        });
-        const collRes = {
-            bucket: {}, // bucket level metrics
-            location: {}, // location level metrics
-            account: {}, // account level metrics
-        };
-        let stalledCount = 0;
-        const cmpDate = new Date();
-        cmpDate.setHours(cmpDate.getHours() - 1);
+    async getObjectMDStats(bucketName, bucketInfo, isTransient, log, callback) {
+        try {
+            const c = this.getCollection(bucketName);
+            const cursor = c.find({}, {
+                projection: {
+                    '_id': 1,
+                    'value.last-modified': 1,
+                    'value.replicationInfo': 1,
+                    'value.dataStoreName': 1,
+                    'value.content-length': 1,
+                    'value.versionId': 1,
+                    'value.owner-id': 1,
+                    'value.isDeleteMarker': 1,
+                    'value.isNull': 1,
+                    'value.archive': 1,
+                    'value.x-amz-storage-class': 1,
+                    'value.isPHD': 1,
+                },
+            });
+            const collRes = {
+                bucket: {}, // bucket level metrics
+                location: {}, // location level metrics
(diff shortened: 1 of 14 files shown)


francoisferrand:
* this change is incomplete : there are other modules which use mongo in this component (e.g. CrrExistingObjects, DataReport, ...) and they must be migrated as well.
* this is a significant/risky change: should be done on next branch only (which is not integrated yet, I think)

ghost:
@francoisferrand : there is no mongodb-related function in other files as far as I can see. Please share example if you saw some. Usage of already migrated functions does not require any change, as callbacks are kept; we only need to change direct use of the MongoDB driver.


We also need this change for the current branch if we want to be able to support this parameter for Artesca 1.7, if we cannot bump arsenal, then we need to create (and maintain) an hotfix branch in Arsenal, for s3utils 1.13.

francoisferrand:
Indeed, I missed the fact we did not use mongo driver directly, but through arsenal which has been migrated!

> We also need this change for the current branch if we want to be able to support this parameter for Artesca 1.7, if we cannot bump arsenal, then we need to create (and maintain) an hotfix branch in Arsenal, for s3utils 1.13.

It would be much better to merge on 1.14, and switch to s3utils 1.14 on Artesca : this way we keep the option to stick or revert to 1.13, and do not mix concerns.

As far as "product" is concerned, it should not matter that we make the change on 1.13 or 1.14 : they will get the same code either way.... but as far as "engineering" (and maintenance) is concerned, this is the safe way.

benzekrimaha (approved):
LGTM

KillianG (approved):

utils/S3UtilsMongoClient.js
```diff
@@ -14,51 +14,57 @@ const __COUNT_ITEMS = 'countitems';
...
             const usersBucketCreationDatesMap = usersBucketCreationDatesArray
                 .reduce((map, obj) => ({ ...map, [obj._id]: obj.value.creationDate }), {});
-            return cursor.forEach(
+            const startCursorDate = new Date();
+            let processed = 0;
+            await cursor.forEach(
                 res => {
+                    // Periodically display information about the cursor
+                    // if more than 30s elapsed
+                    if (Date.now() - startCursorDate > 30000) {
```
francoisferrand, line 61: this will call each time the callback is called, once the 30 second liimit has been reached: need to reset `startCursorDate` when we add a log ?

francoisferrand: this will call each time the callback is called, once the 30 second liimit has been reached: need to reset `startCursorDate` when we add a log ?

```suggestion
                    const currentDate = Date.now();
                    if (currentDate - startCursorDate >= 30000) {
                        startCursorDate = currentDate;
```

utils/S3UtilsMongoClient.js
```diff
@@ -377,32 +400,42 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
-            _id: bucketName,
-        }, {
-            $set: {
-                '_id': bucketName,
-                'value.capabilities.VeeamSOSApi.CapacityInfo': {
-                    Capacity: capacityInfo.Capacity,
-                    Available: capacityInfo.Available,
-                    Used: capacityInfo.Used,
-                    LastModified: (new Date()).toISOString(),
+    async updateBucketCapacityInfo(bucketName, capacityInfo, log, cb) {
```
francoisferrand, line 405: is this function is now `async`, may be cleaner to remove the `cb` and update caller sites with `.then(next).catch(next)` if needed...

(same for all these functions)

ghost: It will be definitely cleaner, but this is not the role of this PR, switching from callbacks to promises will require to change almost all our implementation...

utils/S3UtilsMongoClient.js
ghost: Done in https://github.com/scality/s3utils/commit/8f7df21ab49415c7a5bb523ac6ffb5a1beaa2084

ghost: I rebased and squashed the commits, see the updated link: https://github.com/scality/s3utils/pull/300/files#diff-3a778b361ddfba6106347ff0cb053bcbb1a84374fa0d7e6e6413e7c6fb6ca5cdR55

PR #302, opened 2023-10-20
S3UTILS-146: Recover from specific listing parsing errors

tmacro:
Adds a util function that will attempt to reparse a listing response to recover from specific errors.
If a value fails to parse (only timestamps and ints presently) then the original string value is returned.

### DIFF
--- a/tests/unit/utils/safeList.js
+++ b/tests/unit/utils/safeList.js
@@ -1,0 +1,311 @@
+const xml2js = require('xml2js');
+const { recoverListing } = require('../../../utils/safeList');
+
+
+describe('test safeListObjectVersions', () => {
+    const testCases = [
+        {
+            description: 'should parse valid empty response',
+            response: {
+                ListVersionsResult: {
+                    Name: 'my-bucket',
+                    IsTruncated: false,
+                    MaxKeys: 1000,
+                },
+            },
+            expectedData: {
+                Name: 'my-bucket',
+                IsTruncated: false,
+                MaxKeys: 1000,
+                Versions: [],
+                DeleteMarkers: [],
+                CommonPrefixes: [],
+            },
+        },
+        {
+            description: 'should return error for invalid empty response',
+            response: {},
+            expectedError: new Error('im an original error'),
+        },
+        {
+            description: 'should parse truncated response',
+            response: {
+                ListVersionsResult: {
+                    Name: 'my-bucket',
+                    IsTruncated: true,
+                    MaxKeys: 1000,
+                    NextKeyMarker: 'my-next-key-marker',
+                    NextVersionIdMarker: 'my-next-version-id-marker',
+                    Version: [
+                        {
+                            Key: 'my-key',
+                            VersionId: 'my-version-id',
+                            IsLatest: true,
+                            LastModified: '2021-01-01T00:00:00.000Z',
+                            ETag: 'my-etag',
+                            Size: 123,
+                            StorageClass: 'STANDARD',
+                            Owner: {
+                                ID: 'my-id',
+                                DisplayName: 'my-display-name',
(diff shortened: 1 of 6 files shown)


tests/unit/utils/safeList.js
```diff
@@ -0,0 +1,305 @@
...
+
+            recoverListing(resp, (err, data) => {
+                expect(err).toStrictEqual(testCase.expectedError || null);
+                expect(data).toStrictEqual(testCase.expectedData);
+                done();
+            });
+        });
+    });
+
+    it('should pass through original error if the XML fails to parse', () => {
```
jonathan-gramain: Don't you need a `done` callback to make this test async as well?

utils/safeList.js
```diff
@@ -0,0 +1,286 @@
...
+ * @param {string} key - key to extract
+ * @param {object} data - data to extract from
+ * @returns {ExtractResult} - extracted value
+ */
+
+/**
+ * extractKeys - extract keys from data using the provided extractors
+ *
+ * @param {object} data - data to extract keys from
+ * @param {object<string, Extractor>} keys - Mapping of keys to extractors
```
jonathan-gramain, line 26: Since `Extractor` is a function type, we could want to document here what the function should take as arguments, and what it should return.

jonathan-gramain (approved):

tests/unit/utils/safeList.js
tmacro: strangely it seems to work without, but I'll add one anyway for consistency.

utils/safeList.js
tmacro: `Extractor` and `ExtractResult` are defined above using JSDoc typdefs

utils/safeList.js
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/**
+ * safeExtract - extract a value from data
+ * @param {string} key - key to extract
+ * @param {any} data - data to extract from
+ * @returns {ExtractResult} - extracted value
+ */
+function safeExtract(key, data) {
+    if (data[key] && data[key].length > 0) {
```
ghost: As this is documented as `any`, maybe we should ensure that's an array first?

utils/safeList.js
```diff
@@ -0,0 +1,286 @@
...
+ * Values that fail to parse will be returned as the raw string value present in the xml.
+ *
+ * @param {AWS.S3} s3 - S3 client instance
+ * @param {object} params - listObjectVersions params
+ * @param {function} cb - callback
+ * @returns {undefined}
+ */
+function safeListObjectVersions(s3, params, cb) {
+    const req = s3.listObjectVersions(params);
+    req.on('complete', resp => {
```
ghost, line 271: Don't we also have 'error' events to consider here?

tests/unit/utils/safeList.js
```diff
@@ -0,0 +1,306 @@
...
+    const testCases = [
+        {
+            description: 'should parse empty response',
+            response: {
+                ListVersionsResult: {
+                    Name: 'my-bucket',
+                    IsTruncated: false,
+                    MaxKeys: 1000,
+                },
+            },
```
ghost: I think a test with an empty response would make sense here, to make sure we respond correctly

utils/safeList.js
tmacro: The xml parser only returns an object of arrays but I can swap it for a `Array.isArray` call instead

utils/safeList.js
tmacro: The "complete" event is fired when the request completes, regardless of whether there was an error.

ghost (approved):
LGTM (maybe you want to bump the version before merging?)

PR #303, opened 2023-11-14
remove UpdateBucketCapacityJobTakingTooLong alert

welansari:
UpdateBucketCapacityJobTakingTooLong creates noise and is not useful as what we truly want to monitor is having multiple jobs failing in series.

NoSuccessfulUpdateBucketCapacityJobRunIn10m already tests that as jobs are only run one at a time and at a 4 minute interval.

Issue: S3UTILS-150

### DIFF
--- a/tests/unit/CompareRaftMembers/DiffStreamOplogFilter.js
+++ b/tests/unit/CompareRaftMembers/DiffStreamOplogFilter.js
@@ -5,8 +5,6 @@ const { shuffle } = require('arsenal');
 
 const DiffStreamOplogFilter = require('../../../CompareRaftMembers/DiffStreamOplogFilter');
 
-const HTTP_TEST_PORT = 9090;
-
 class MockRaftOplogStream extends stream.Readable {
     constructor(entriesToEmit, refreshPeriodMs) {
         super({ objectMode: true });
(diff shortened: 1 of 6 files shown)


francoisferrand (approved):

tests/unit/CompareRaftMembers/DiffStreamOplogFilter.js
```diff
@@ -5,7 +5,7 @@ const { shuffle } = require('arsenal');
 
 const DiffStreamOplogFilter = require('../../../CompareRaftMembers/DiffStreamOplogFilter');
 
-const HTTP_TEST_PORT = 9090;
+const HTTP_TEST_PORT = 9091;
```
francoisferrand: a safer alternative may be to let the server:
1. bind to localhost only (avoid requiring extra permissions)
2. allocates a random port (i.e. specify port = 0)

this way, there will be no conflict, and the port to connect to can be retrieved by calling `server.address().port`, something like this:
```
  server = http.createServer();
  server.listen(':0');
  port = server.address().port
```

francoisferrand (approved):

KillianG (approved):

nicolas2bert (approved):

PR #304, opened 2023-11-21
impr(S3UTILS-147/S3UTILS-151):

fra-scality:
we found corner cases where the MD answer was:
- not a correct - invalid -  JSON object
- not empty but Null

this should fix these two corner cases, so the scan of the SproxyD keys is not interrupted with an exception but it logs a specific error


francoisferrand:
@fra-scality this does not seem to be related to S3UTILS-150, can you please check & fix the link to Jira ?
(in PR title & commit descriptions)

PR #309, opened 2024-03-25
Improvement/s3 utils 155 granular metrics

ghost:
Update the job to:

- Add granularity for objects in cold storage, being restored, or restored, in all the computed metrics. Code architecture is kept, this only adds linear complexity (that we have no choice but to compute).
- Add more unit and functionsl tests to cover cold objects (restoring, restored and in cold).
- Change the logic when an object is being restored, to count it also in the destination location (i.e., the hot one). This ensures the returned metric is consistent with the upcoming quota checks: we consider objects pending restoration as already part of the metrics of the destination (hot) location. This only applies to locations.

Main commit to review : https://github.com/scality/s3utils/pull/309/commits/4317c3ef84cc60103b57c6f457eab7d8a088deee


BourgoisMickael (approved):

KazToozs (approved):

ghost:
FYI @KazToozs @BourgoisMickael , I added a commit that also splits the `restoring` and `restored` into current & nonCurrent. It makes more sense in terms of logic, and it also ensures backward compatibility with the UI (where Pensieve-API will return the right aggregated metric transparently).

Edit: added another one to keep the original values for current/nonCurrent, making the code fully backward compatible, without altering existing projects, and ability to know the data not in cold.

PR #310, opened 2024-03-26
S3UTILS-157: explicit closing of mongodb cursors

ghost:
There is suspicion of cursor leaks with the CountItems job, especially when enumerating a large number of cursors. This PR adds explicit checks after the end of the processing. The cursors should already be closed as per the nature of the `forEach`, that closes the cursor when there is no resource to check anymore.

Issue: S3UTILS-157

### DIFF
--- a/utils/S3UtilsMongoClient.js
+++ b/utils/S3UtilsMongoClient.js
@@ -14,10 +14,40 @@ const __COUNT_ITEMS = 'countitems';
 
 
 class S3UtilsMongoClient extends MongoClientInterface {
+    /**
+     * Get the list of buckets and their location dates
+     * @param {object} log - Werelogs logger
+     * @returns {object} - Object with bucket names as keys
+     * and their creation dates as values
+     */
+    async _getUsersBucketCreationDates(log) {
+        let cursorUsersBucketCreationDates;
+        try {
+            cursorUsersBucketCreationDates = await this.getCollection(USERSBUCKET).find({}, {
+                projection: {
+                    'value.creationDate': 1,
+                },
+            });
+            const usersBucketCreationDatesArray = await cursorUsersBucketCreationDates.toArray();
+            return usersBucketCreationDatesArray
+                .reduce((map, obj) => ({ ...map, [obj._id]: obj.value.creationDate }), {});
+        } catch (err) {
+            log.error('Failed to read __usersbucket collection', {
+                method: 'getUsersBucketCreationDates',
+                errDetails: { ...err },
+                errorString: err.toString(),
+            });
+            return null;
+        } finally {
+            await cursorUsersBucketCreationDates.close();
+        }
+    }
+
     async getObjectMDStats(bucketName, bucketInfo, isTransient, log, callback) {
+        let cursor;
         try {
             const c = this.getCollection(bucketName);
-            const cursor = c.find({}, {
+            cursor = c.find({}, {
                 projection: {
                     '_id': 1,
                     'value.last-modified': 1,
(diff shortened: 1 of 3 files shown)


utils/S3UtilsMongoClient.js
```diff
@@ -44,13 +46,13 @@ class S3UtilsMongoClient extends MongoClientInterface {
 
             const locationConfig = getLocationConfig(log);
 
-            const usersBucketCreationDatesArray = await this.getCollection(USERSBUCKET).find({}, {
+            const cursorUsersBucketCreationDates = await this.getCollection(USERSBUCKET).find({}, {
```
francoisferrand: ~nit: `usersBucketCreationDatesCursor` to be slightly more "grammatically" correct~

why (re)name "this" cursor?
Since `toArray()` is called, this is not a cursor at all, but really an array... One could argue there is no need to make the `Array` explicit in the name indeed, but no point introducing `cursor` either...

utils/S3UtilsMongoClient.js
```diff
@@ -196,6 +198,21 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
             return callback(err);
+        } finally {
+            if (cursor && !cursor.closed) {
+                log.info('Finished processing cursor', {
+                    method: 'getObjectMDStats',
+                    bucketName,
+                });
+                cursor.close();
+            }
+            if (cursorUsersBucketCreationDates && !cursorUsersBucketCreationDates.closed) {
```
francoisferrand: this is not a cursor, but an array: so cannot be closed like this I guess?

would need to introduce an extra variable before converting to array... and ideally close the cursor ASAP once we have retrieved the array...

ghost:
Putting this work on hold as for now this does not look like the cause of the number of open connections to MongoDB, we need to reproduce first, otherwise these changes are not needed.

utils/S3UtilsMongoClient.js
ghost: I updated, even if this is indeed not stricly necessary, as we convert the cursor to an array (also applies to `forEach`s that complete); we now explicitly close the cursor. 
https://github.com/scality/s3utils/pull/310/files#diff-3a778b361ddfba6106347ff0cb053bcbb1a84374fa0d7e6e6413e7c6fb6ca5cdR54

utils/S3UtilsMongoClient.js
ghost: Removed the line as per my last comment, cursor is closed right after the `toArray()` call
The code for the other cursor is kept, as indeed in case of error durring the processing, that may lead an open cursor.

utils/S3UtilsMongoClient.js
```diff
@@ -44,11 +45,15 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+            const cursorUsersBucketCreationDates = await this.getCollection(USERSBUCKET).find({}, {
                 projection: {
                     'value.creationDate': 1,
                 },
-            }).toArray();
+            });
+
+            const usersBucketCreationDatesArray = await cursorUsersBucketCreationDates.toArray();
+
+            await cursorUsersBucketCreationDates.close();
```
francoisferrand: shouldn't we introduce a (small) `finally` block to close this cursor, to ensure we close the cursor in case `toArray()` fails with an exception?

```suggestion
	const usersBucketCreationDatesArray = await async() {
	   try {
          return await cursorUsersBucketCreationDates.toArray();
       } finally {
	      await cursorUsersBucketCreationDates.close();
	   }
	}();
```

(may be cleaner to extract this and `cursorUsersBucketCreationDates` creation to a separate method, though)

ghost: Done in https://github.com/scality/s3utils/pull/310/commits/5017e66391393dbb5a59114897abf68d59efdb21 , with a new function (unit tested)

utils/S3UtilsMongoClient.js
```diff
@@ -14,10 +14,40 @@ const __COUNT_ITEMS = 'countitems';
...
+            cursorUsersBucketCreationDates = await this.getCollection(USERSBUCKET).find({}, {
+                projection: {
+                    'value.creationDate': 1,
+                },
+            });
+            const usersBucketCreationDatesArray = await cursorUsersBucketCreationDates.toArray();
+            return usersBucketCreationDatesArray
+                .reduce((map, obj) => ({ ...map, [obj._id]: obj.value.creationDate }), {});
+        } catch (err) {
+            log.error('Failed to read __usersbucket collection', {
```
francoisferrand, line 35: nit: any reason to catch the exception instead of letting it propagate up, like the other exceptions which may be generated from within `getObjectMDStats()` and get caught at the end of that function?

(i am fine either way, just wondering)

francoisferrand (approved):

benzekrimaha (approved):

ghost:
(rebasing to remove a unused line, leftover from testing)

utils/S3UtilsMongoClient.js
ghost: To was mainly to ensure the right error messages are logged, so we can return an arsenal error

PR #312, opened 2024-05-13
Feature/s3 utils 158/storage consumption metrics

benzekrimaha:
This PR aims to provide monitoring metrics for the count items cron-job. 
Today the metrics that have been put in place are : 
- `bucketCount` : Total number of buckets
- `bucketProcessingDuration` : Duration of processing a bucket
- `consolidationDuration` : Duration of metrics consolidation in seconds
- `workersCount` : a metric that shows success / errors during the workers count 
- `objectsCount` : Number of objects based on their states 
- `metricsCount` : Number of buckets ( none empty) , number of accounts with buckets , and number of locations 

- It also puts in place a job stop mecanism resumed with : 
==> If the job completed and the scraping occured => **immediate** stop 
==> Otherwise we wait for a duration ( threshold x frequency x 4 x 1000 ) if a scraping occurs during it we stop otherwise we wait for the whole duration then we stop the job 

_Note for reviewers : this is a first iteration as the job has never been monitored ( only alerts were put in place ) please let me know if you see that some metrics might make more sens / or others are not really needed_ 

Linked issue : https://scality.atlassian.net/browse/S3UTILS-158

### DIFF
--- a/CountItems/CountManager.js
+++ b/CountItems/CountManager.js
@@ -2,6 +2,7 @@ const async = require('async');
 const { once } = require('arsenal').jsutil;
 const { validStorageMetricLevels } = require('./utils/constants');
 const { consolidateDataMetrics } = require('./utils/utils');
+const monitoring = require('../utils/monitoring');
 
 class CountManager {
     constructor(params) {
@@ -35,6 +36,7 @@ class CountManager {
                 return done(new Error('emptyWorkerList'));
             }
             const id = this.workerList.shift();
+            const processingStartTime = process.hrtime.bigint();
             return this.workers[id].count(bucketInfo, (err, res) => {
                 this.log.info('processing a bucket', {
                     method: 'CountManager::_setupQueue',
@@ -44,6 +46,8 @@ class CountManager {
                 if (err) {
                     return done(err);
                 }
+                const processingDuration = Number(process.hrtime.bigint() - processingStartTime) / 1e9;
+                monitoring.bucketProcessingDuration.observe(processingDuration);
                 this._consolidateData(res);
                 this.workerList.push(id);
                 return done();
@@ -53,6 +57,7 @@ class CountManager {
     }
 
     _consolidateData(results) {
+        const startTime = process.hrtime.bigint();
         if (!results) {
             return;
         }
@@ -111,6 +116,8 @@ class CountManager {
         } else {
             this.dataMetrics = results.dataMetrics;
         }
+        const consolidationDurationInS = Number(process.hrtime.bigint() - startTime) / 1e9;
+        monitoring.consolidationDuration.observe(consolidationDurationInS);
     }
 
     setup(callback) {
--- a/CountItems/CountMaster.js
+++ b/CountItems/CountMaster.js
@@ -58,7 +58,7 @@ class CountMaster {
                 });
                 return this.stop(null, () => callback(err));
             }
-            return this.stop(null, () => callback());
+            return callback();
         });
     }
 }
--- a/CountItems/CountWorker.js
+++ b/CountItems/CountWorker.js
@@ -1,6 +1,7 @@
 const assert = require('assert');
 const async = require('async');
 const { BucketInfo } = require('arsenal').models;
+const monitoring = require('../utils/monitoring');
 
 class CountWorker {
     constructor(params) {
@@ -35,7 +36,10 @@ class CountWorker {
         return async.waterfall([
             next => this.client._getIsTransient(bucketInfo, this.log, next),
             (isTransient, next) => this.client.getObjectMDStats(bucketName, bucketInfo, isTransient, this.log, next),
-        ], callback);
+        ], (err, results) => {
+            monitoring.bucketsCount.inc({ status: err ? 'error' : 'success' });
+            callback(err, results);
+        });
     }
 
     clientTeardown(callback) {
--- a/CountItems/masterProcess.js
+++ b/CountItems/masterProcess.js
@@ -1,4 +1,5 @@
 const werelogs = require('werelogs');
+const { network } = require('arsenal');
 const { reshapeExceptionError } = require('arsenal').errorUtils;
 const S3UtilsMongoClient = require('../utils/S3UtilsMongoClient');
 
@@ -6,6 +7,9 @@ const CountMaster = require('./CountMaster');
 const CountManager = require('./CountManager');
 const createMongoParams = require('../utils/createMongoParams');
 const createWorkers = require('./utils/createWorkers');
+
+const WebServer = network.http.server;
+const monitoring = require('../utils/monitoring');
 
 const logLevel = Number.parseInt(process.env.DEBUG, 10) === 1
     ? 'debug' : 'info';
@@ -15,8 +19,19 @@ const loggerConfig = {
     dump: 'error',
 };
 
+let waitingForPromScraping = false;
+
 werelogs.configure(loggerConfig);
 const log = new werelogs.Logger('S3Utils::CountItems::Master');
+
+function tryParseInt(s, defaultValue) {
+    const v = Number.parseInt(s, 10);
+    return v > 0 ? v : defaultValue;
+}
+
+const prometheusPollingPeriod = tryParseInt(process.env.PROMETHEUS_POLLING_PERIOD, 30);
+
+const prometheusPollingAttempts = tryParseInt(process.env.PROMETHEUS_POLLING_ATTEMPTS, 5);
 
 const numWorkers = process.env.NUM_WORKERS && !Number.isNaN(process.env.NUM_WORKERS)
     ? Number.parseInt(process.env.NUM_WORKERS, 10)
@@ -37,6 +52,17 @@ const countMaster = new CountMaster({
     client: new S3UtilsMongoClient(createMongoParams(log)),
 });
 
+const metricServer = new WebServer(8003, log).onRequest((req, res) => monitoring.metricsHandler(
+    () => {
+        if (waitingForPromScraping === true) {
+            countMaster.stop(null, () => process.exit(1));
+        }
+    },
+    req,
+    res,
+));
+metricServer.start();
+
 const handleSignal = sig => countMaster.stop(sig, () => process.exit(0));
 process.on('SIGINT', handleSignal);
 process.on('SIGHUP', handleSignal);
@@ -53,5 +79,8 @@ countMaster.start(err => {
     if (err) {
         process.exit(1);
     }
-    process.exit(0);
+    waitingForPromScraping = true;
+    setTimeout(() => {
+        countMaster.stop(null, () => process.exit(0));
+    }, prometheusPollingAttempts * prometheusPollingPeriod * 1000 * 4);
 });
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3utils",
-  "version": "1.14.6",
+  "version": "1.14.7",
   "engines": {
     "node": ">= 16"
   },
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "eslint-plugin-import": "^2.20.1",
     "eslint-plugin-jest": "^23.6.0",
     "jest": "^23.6.0",
-    "mongodb-memory-server": "^8.10.2"
+    "mongodb-memory-server": "^8.10.2",
+    "sinon": "^17.0.1"
   }
 }
--- a/utils/S3UtilsMongoClient.js
+++ b/utils/S3UtilsMongoClient.js
@@ -5,6 +5,7 @@ const { errors, constants } = require('arsenal');
 const async = require('async');
 const { validStorageMetricLevels } = require('../CountItems/utils/constants');
 const getLocationConfig = require('./locationConfig');
+const monitoring = require('./monitoring');
 
 const METASTORE = '__metastore';
 const INFOSTORE = '__infostore';
@@ -150,6 +151,7 @@ class S3UtilsMongoClient extends MongoClientInterface {
                             entry: res,
                             error,
                         });
+                        monitoring.objectsCount.inc({ status: 'error' });
                         return;
                     }
 
@@ -159,6 +161,7 @@ class S3UtilsMongoClient extends MongoClientInterface {
                             method: 'getObjectMDStats',
                             entry: res,
                         });
+                        monitoring.objectsCount.inc({ status: 'skipped' });
                         return;
                     }
 
@@ -231,6 +234,7 @@ class S3UtilsMongoClient extends MongoClientInterface {
                             collRes.account[account].locations[location].deleteMarkerCount += res.value.isDeleteMarker ? 1 : 0;
                         });
                     });
+                    monitoring.objectsCount.inc({ status: 'success' });
                     processed++;
                 },
                 err => {
--- a/utils/monitoring.js
+++ b/utils/monitoring.js
@@ -1,0 +1,88 @@
+const { errors } = require('arsenal');
+const promClient = require('prom-client');
+const { Registry } = require('prom-client');
+const { http } = require('httpagent');
+const CountMaster = require('../CountItems/CountMaster');
+
+const aggregatorRegistry = new promClient.AggregatorRegistry();
+const { collectDefaultMetrics } = promClient;
+
+
+// Histogram of the bucket processing duration, by the utilization service.
+const bucketProcessingDuration = new promClient.Histogram({
+    name: 's3_countitems_bucket_listing_duration_seconds',
+    help: 'Bucket processing duration',
+    buckets: [1, 10, 60, 600, 3600, 18000, 36000],
+});
+
+const consolidationDuration = new promClient.Histogram({
+    name: 's3_countitems_bucket_merge_duration_seconds',
+    help: 'Duration of metrics consolidation in seconds',
+    buckets: [0.01, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10],
+});
+
+const bucketsCount = new promClient.Counter({
+    name: 's3_countitems_total_buckets_count',
+    help: 'Total number of buckets processed',
+    labelNames: ['status'],
+});
+
+const objectsCount = new promClient.Counter({
+    name: 's3_countitems_total_objects_count',
+    help: 'Total number of objects processed',
+    labelNames: ['status'],
+});
+
+/**
+ * @param {http.ServerResponse} res - http response object
+ * @param {Error | errors.ArsenalError} error - Error
+ * @return {void}
+ */
+function _writeResponse(res, error) {
+    let statusCode = 500;
+    if (error instanceof errors.ArsenalError && Number.isInteger(error.code)) {
+        statusCode = error.code;
+    }
+    res.writeHead(statusCode, { 'Content-Type': 'application/json' });
+    res.end();
+}
+
+/**
+ * @param {function} onScraped - callback to call when metrics are scraped
+ * @param {http.IncomingMessage} req - http request object
+ * @param {http.ServerResponse} res - http response object
+ * @return {void}
+ */
+async function metricsHandler(onScraped, req, res) {
+    if (req.method !== 'GET' || req.url !== '/metrics') {
+        return _writeResponse(res, errors.MethodNotAllowed);
+    }
+    try {
+        const [registerMetrics, clusterMetrics] = await Promise.all([
+            promClient.register.metrics(),
+            aggregatorRegistry.clusterMetrics(),
+        ]);
+        const promMetrics = `${registerMetrics}\n${clusterMetrics}`;
+        const contentLen = Buffer.byteLength(promMetrics, 'utf8');
+        return res.writeHead(200, {
+            'content-length': contentLen,
+            'content-type': promClient.register.contentType,
+        }).end(promMetrics);
+    } catch (ex) {
+        return _writeResponse(res, ex);
+    } finally {
+        if (onScraped) {
+            onScraped();
+        }
+    }
+}
+
+module.exports = {
+    client: promClient,
+    collectDefaultMetrics,
+    metricsHandler,
+    bucketProcessingDuration,
+    consolidationDuration,
+    bucketsCount,
+    objectsCount,
+};
--- a/tests/functional/countItems.js
+++ b/tests/functional/countItems.js
@@ -3,8 +3,8 @@ const async = require('async');
 const werelogs = require('werelogs');
 const { BucketInfo, ObjectMD } = require('arsenal').models;
 const { constants } = require('arsenal');
+const sinon = require('sinon');
 const S3UtilsMongoClient = require('../../utils/S3UtilsMongoClient');
-
 const CountMaster = require('../../CountItems/CountMaster');
 const CountManager = require('../../CountItems/CountManager');
 const createMongoParams = require('../../utils/createMongoParams');
@@ -190,6 +190,7 @@ jest.setTimeout(120000);
 describe('CountItems', () => {
     const oldEnv = process.env;
     let client;
+    let setTimeoutSpy;
 
     beforeAll(done => {
         process.env = oldEnv;
@@ -209,9 +210,12 @@ describe('CountItems', () => {
             next => client.setup(next),
             next => populateMongo(client, next),
         ], done);
+        setTimeoutSpy = jest.spyOn(global, 'setTimeout');
+        setTimeoutSpy.mockImplementation((callback, delay) => callback());
     });
 
     afterAll(done => {
+        setTimeoutSpy.mockRestore();
         async.series([
             next => client.db.dropDatabase().then(() => next()).catch(() => next()),
             next => client.close(next),
--- a/yarn.lock
+++ b/yarn.lock
@@ -1037,12 +1037,47 @@
   dependencies:
     type-detect "4.0.8"
 
+"@sinonjs/commons@^2.0.0":
+  version "2.0.0"
+  resolved "https://registry.yarnpkg.com/@sinonjs/commons/-/commons-2.0.0.tgz#fd4ca5b063554307e8327b4564bd56d3b73924a3"
+  integrity sha512-uLa0j859mMrg2slwQYdO/AkrOfmH+X6LTVmNTS9CqexuE2IvVORIkSpJLqePAbEnKJ77aMmCwr1NUZ57120Xcg==
+  dependencies:
+    type-detect "4.0.8"
+
+"@sinonjs/commons@^3.0.0":
+  version "3.0.1"
+  resolved "https://registry.yarnpkg.com/@sinonjs/commons/-/commons-3.0.1.tgz#1029357e44ca901a615585f6d27738dbc89084cd"
+  integrity sha512-K3mCHKQ9sVh8o1C9cxkwxaOmXoAMlDxC1mYyHrjqOWEcBjYr76t96zL2zlj5dUGZ3HSw240X1qgH3Mjf1yJWpQ==
+  dependencies:
+    type-detect "4.0.8"
+
+"@sinonjs/fake-timers@^11.2.2":
+  version "11.2.2"
+  resolved "https://registry.yarnpkg.com/@sinonjs/fake-timers/-/fake-timers-11.2.2.tgz#50063cc3574f4a27bd8453180a04171c85cc9699"
+  integrity sha512-G2piCSxQ7oWOxwGSAyFHfPIsyeJGXYtc6mFbnFA+kRXkiEnTl8c/8jul2S329iFBnDI9HGoeWWAZvuvOkZccgw==
+  dependencies:
+    "@sinonjs/commons" "^3.0.0"
+
 "@sinonjs/fake-timers@^9.1.2":
   version "9.1.2"
   resolved "https://registry.yarnpkg.com/@sinonjs/fake-timers/-/fake-timers-9.1.2.tgz#4eaab737fab77332ab132d396a3c0d364bd0ea8c"
   integrity sha512-BPS4ynJW/o92PUR4wgriz2Ud5gpST5vz6GQfMixEDK0Z8ZCUv2M7SkBLykH56T++Xs+8ln9zTGbOvNGIe02/jw==
   dependencies:
     "@sinonjs/commons" "^1.7.0"
+
+"@sinonjs/samsam@^8.0.0":
+  version "8.0.0"
+  resolved "https://registry.yarnpkg.com/@sinonjs/samsam/-/samsam-8.0.0.tgz#0d488c91efb3fa1442e26abea81759dfc8b5ac60"
+  integrity sha512-Bp8KUVlLp8ibJZrnvq2foVhP0IVX2CIprMJPK0vqGqgrDa0OHVKeZyBykqskkrdxV6yKBPmGasO8LVjAKR3Gew==
+  dependencies:
+    "@sinonjs/commons" "^2.0.0"
+    lodash.get "^4.4.2"
+    type-detect "^4.0.8"
+
+"@sinonjs/text-encoding@^0.7.2":
+  version "0.7.2"
+  resolved "https://registry.yarnpkg.com/@sinonjs/text-encoding/-/text-encoding-0.7.2.tgz#5981a8db18b56ba38ef0efb7d995b12aa7b51918"
+  integrity sha512-sXXKG+uL9IrKqViTtao2Ws6dy0znu9sOaP1di/jKGW1M6VssO8vlpXCQcpZ+jisQ1tTFAC5Jo/EOzFbggBagFQ==
 
 "@socket.io/component-emitter@~3.1.0":
   version "3.1.0"
@@ -2744,6 +2779,11 @@ diff@^3.2.0:
   version "3.5.0"
   resolved "https://registry.yarnpkg.com/diff/-/diff-3.5.0.tgz#800c0dd1e0a8bfbc95835c202ad220fe317e5a12"
   integrity sha512-A46qtFgd+g7pDZinpnwiRJtxbC1hpgf0uzP3iG89scHk0AUC7A1TGxf5OiiOUv/JMZR8GOt8hL900hV0bOy5xA==
+
+diff@^5.1.0:
+  version "5.2.0"
+  resolved "https://registry.yarnpkg.com/diff/-/diff-5.2.0.tgz#26ded047cd1179b78b9537d5ef725503ce1ae531"
+  integrity sha512-uIFDxqpRZGZ6ThOk84hEfqWoHx2devRFvpTZcTHur85vImfaxUbTW9Ryh4CpCuDnToOP1CEtXKIgytHBPVff5A==
 
 diskusage@^1.1.1:
   version "1.1.3"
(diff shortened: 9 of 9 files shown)


utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        labelNames: ['service'],
+        buckets: [1, 10, 60, 600, 3600, 18000, 36000],
+    });
+
+const consolidationDuration = (promClient.register.getSingleMetric('count_items_consolidationDuration'))
+    || new promClient.Histogram({
+        name: 'count_items_consolidationDuration',
+        help: 'Duration of metrics consolidation in seconds',
+        labelNames: ['service'],
+        buckets: [0.1, 0.5, 1, 2, 5, 10],
```
ghost: I think we can have smaller values, as no I/O is involved here

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+    || new promClient.Counter({
+        name: 'count_items_workersCountErrors',
+        help: 'Number of errors in counting workers',
+        labelNames: ['state'],
+    });
+
+const objectsCount = (promClient.register.getSingleMetric('count_items_objectsCount'))
+    || new promClient.Counter({
+        name: 'count_items_objectsCount',
+        help: 'Number of objects',
```
ghost:
```suggestion
        help: 'Number of objects processed',
```

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        labelNames: ['state'],
+    });
+
+const metricsCount = (promClient.register.getSingleMetric('count_items_metricsCount'))
+    || new promClient.Counter({
+        name: 'count_items_metricsCount',
+        help: 'Number of metrics',
```
ghost:
```suggestion
        help: 'Number of entities with data usage',
```

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        help: 'Total number of buckets',
+        labelNames: ['bucketName'],
+    });
+
+// Histogram of the bucket processing duration, by the utilization service.
+
+const bucketProcessingDuration = (promClient.register.getSingleMetric('count_items_bucketProcessingDuration'))
+    || new promClient.Histogram({
+        name: 'count_items_bucketProcessingDuration',
+        help: 'Duration of processing a bucket',
```
ghost:
```suggestion
        help: 'Bucket processing duration',
```

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+const { Registry } = require('prom-client');
+const CountMaster = require('../CountItems/CountMaster');
+
+const aggregatorRegistry = new promClient.AggregatorRegistry();
+const { collectDefaultMetrics } = promClient;
+
+const bucketCount = (promClient.register.getSingleMetric('bucketCount_metrics'))
+    || new promClient.Gauge({
+        name: 'bucketCount_metrics',
+        help: 'Total number of buckets',
```
ghost:
```suggestion
        help: 'Total number of buckets processed',
```

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+function _writeResponse(res, error) {
+    let statusCode = 200;
+    if (error) {
+        if (Number.isInteger(error.code)) {
+            statusCode = error.code;
+        } else {
+            statusCode = 500;
+        }
+    }
```
ghost: It seems to be always `MethodNotAllowed`, even in the `catch` case,, either we remove this logic, or we pass the real error from the exception?

package.json
ghost, line 1: You can bump the project's version as we'll need to release it

CountItems/masterProcess.js
```diff
@@ -26,7 +30,7 @@ const concurrentCursors = (process.env.CONCURRENT_CURSORS
     && !Number.isNaN(process.env.CONCURRENT_CURSORS))
     ? Number.parseInt(process.env.CONCURRENT_CURSORS, 10)
     : 5;
-
+const metricServer = new WebServer(8003, log);
```
ghost:
```suggestion

const metricServer = new WebServer(8003, log);
```

package.json
benzekrimaha: done here : https://github.com/scality/s3utils/pull/312/commits/8ebe3b36b0f444dd16b7f38b8f5312a6852e8fcc#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R3

utils/monitoring.js
benzekrimaha: the real error is now returned , see here: https://github.com/scality/s3utils/pull/312/commits/5b7f15a750eb30d1dda2b2ed562563334cca9e57

utils/monitoring.js
benzekrimaha: changed here : https://github.com/scality/s3utils/pull/312/commits/845c8698405f12a5273fce139241140e7c4d42df#diff-612a4c10e405510cdaad879589d4a7d157df6e0ff42b4645d2a21d4d17e41fa9R31

CountItems/CountManager.js
```diff
@@ -26,6 +27,9 @@ class CountManager {
             account: {},
         };
         this.workerList = [];
+        this.uniqueAccounts = new Set();
+        this.uniqueLocations = new Set();
```
francoisferrand: the list of locations is actually known : we have it in the configuration, if needed and not already provided to `countItems` it may be added in defintion of the job (by zkop)

CountItems/CountManager.js
```diff
@@ -35,7 +39,12 @@ class CountManager {
                 return done(new Error('emptyWorkerList'));
             }
             const id = this.workerList.shift();
+            const processingStartTime = process.hrtime.bigint();
             return this.workers[id].count(bucketInfo, (err, res) => {
+                const processingDuration = Number(process.hrtime.bigint() - processingStartTime) / 1e9;
+                monitoring.bucketProcessingDuration.labels({
+                    service: 'countItems',
+                }).observe(processingDuration);
```
francoisferrand: * should be done at the same time as `monitoring.bucketCount.inc();`
* actually, that second metrics is redundant : historgram metric (`bucketProcessingDuration` in this case) already include a counter for the total number of time `observe()` is called

CountItems/CountManager.js
```diff
@@ -84,6 +95,18 @@ class CountManager {
...
+                            monitoring.metricsCount.inc({ metricLevel: 'account' });
+                        }
+                        if (metricLevel === 'location' && !this.uniqueLocations.has(resourceName)) {
+                            this.uniqueLocations.add(resourceName);
+                            monitoring.metricsCount.inc({ metricLevel: 'location' });
+                        }
+                        if (metricLevel === 'bucket' && !this.uniqueBuckets.has(resourceName)) {
+                            this.uniqueBuckets.add(resourceName);
+                            monitoring.metricsCount.inc({ metricLevel: 'bucket' });
+                        }
```
francoisferrand: no need for the extra variable or to do this here : you can simply update the metrics by looking at `this.dataMetrics` at the end of the loop: just look at `this.dataMetrics[metricLevel].size` (or the increase before/after the loop) for each metricLevel...

since there is no I/O or anything here, there is no benefit to having the "progress" of this operation, which will probably not be scrapped in between.

CountItems/CountManager.js
```diff
@@ -26,6 +27,9 @@ class CountManager {
             account: {},
         };
         this.workerList = [];
+        this.uniqueAccounts = new Set();
+        this.uniqueLocations = new Set();
+        this.uniqueBuckets = new Set();
```
francoisferrand: these variables are intrinsically not scalable : since they would eventually include one entry for each account/bucket (which we may have millions.....). True, there are already such a variable for the metrics, but we should avoid adding even more such variable...

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
+const { errors } = require('arsenal');
+const promClient = require('prom-client');
+const { Registry } = require('prom-client');
+const CountMaster = require('../CountItems/CountMaster');
+
+const aggregatorRegistry = new promClient.AggregatorRegistry();
+const { collectDefaultMetrics } = promClient;
+
+const bucketCount = (promClient.register.getSingleMetric('bucketCount_metrics'))
+    || new promClient.Gauge({
```
francoisferrand: what is the point of this pattern?

the metric should just be created once, it should not be already in the register: and if not, it means we should have already initialized the variable....

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        labelNames: ['bucketName'],
+    });
+
+// Histogram of the bucket processing duration, by the utilization service.
+
+const bucketProcessingDuration = (promClient.register.getSingleMetric('count_items_bucketProcessingDuration'))
+    || new promClient.Histogram({
+        name: 'count_items_bucketProcessingDuration',
+        help: 'Bucket processing duration',
+        labelNames: ['service'],
```
francoisferrand: what is the point of this `service` label?

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+const promClient = require('prom-client');
+const { Registry } = require('prom-client');
+const CountMaster = require('../CountItems/CountMaster');
+
+const aggregatorRegistry = new promClient.AggregatorRegistry();
+const { collectDefaultMetrics } = promClient;
+
+const bucketCount = (promClient.register.getSingleMetric('bucketCount_metrics'))
+    || new promClient.Gauge({
+        name: 'bucketCount_metrics',
```
francoisferrand: Need to follow naming conventions for metrics:
* s3 prefix
* name of component
* lowercase
* end with an indication of the unit

Also this metric is indeed the _"total number fo buckets processed"_, meaning:
- it is not a gauge, but really a counter (even increasing!)... and it is up to prometheus to "filter out" resets of the metric
- the name must clearly convey the fact this is not the current number of buckets...

```suggestion
        name: 's3_countitems_total_buckets_count',
```

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        help: 'Total number of buckets processed',
+        labelNames: ['bucketName'],
```
francoisferrand: `bucketName` is a high cardinality label, must not be label...
(and logically, this is the number of buckets, so no point have a bucketName label)

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+// Histogram of the bucket processing duration, by the utilization service.
+
+const bucketProcessingDuration = (promClient.register.getSingleMetric('count_items_bucketProcessingDuration'))
+    || new promClient.Histogram({
+        name: 'count_items_bucketProcessingDuration',
```
francoisferrand: naming convention, but also `processing` is overuse ("consolidation" is actually more processing than the first phase)...

```suggestion
        name: 's3_countitems_bucket_listing_duration_seconds',
```

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        help: 'Bucket processing duration',
+        labelNames: ['service'],
+        buckets: [1, 10, 60, 600, 3600, 18000, 36000],
+    });
+
+const consolidationDuration = (promClient.register.getSingleMetric('count_items_consolidationDuration'))
+    || new promClient.Histogram({
+        name: 'count_items_consolidationDuration',
```
francoisferrand:
```suggestion
        name: 's3_countitems_bucket_merge_duration_seconds',
```

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        help: 'Duration of metrics consolidation in seconds',
+        labelNames: ['service'],
+        buckets: [0.01, 0.05, 0.1, 0.2, 0.5, 1, 1.5, 2],
```
francoisferrand:
```suggestion
        buckets: [0.01, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10],
```

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+const workersCount = (promClient.register.getSingleMetric('count_items_workersCountErrors'))
+    || new promClient.Counter({
+        name: 'count_items_workersCountErrors',
+        help: 'Number of errors in counting workers',
+        labelNames: ['state'],
+    });
+
+const objectsCount = (promClient.register.getSingleMetric('count_items_objectsCount'))
+    || new promClient.Counter({
+        name: 'count_items_objectsCount',
```
francoisferrand:
```suggestion
        name: 's3_countitems_total_objects_count',
```

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        help: 'Number of objects processed',
+        labelNames: ['state'],
+    });
+
+const metricsCount = (promClient.register.getSingleMetric('count_items_metricsCount'))
+    || new promClient.Counter({
+        name: 'count_items_metricsCount',
+        help: 'Number of entities with data usage',
```
francoisferrand: i don't think `entities` is right here, since entities is already a vault concept...

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+const objectsCount = (promClient.register.getSingleMetric('count_items_objectsCount'))
+    || new promClient.Counter({
+        name: 'count_items_objectsCount',
+        help: 'Number of objects processed',
+        labelNames: ['state'],
+    });
+
+const metricsCount = (promClient.register.getSingleMetric('count_items_metricsCount'))
+    || new promClient.Counter({
+        name: 'count_items_metricsCount',
```
francoisferrand: what is the point of this metric?
the number of locations & buckets is already known ; and number of accounts should be reported by vault instead...

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        help: 'Number of entities with data usage',
+        labelNames: ['metricLevel'],
```
francoisferrand: should not use `metric` in metrics definition, this is too confusing (inception-like)

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+/**
+ * @param {http.ServerResponse} res - http response object
+ * @param {Error | errors.ArsenalError} error - Error
+ * @return {void}
+ */
+function _writeResponse(res, error) {
+    let statusCode = 200;
+    if (error) {
```
francoisferrand: what is the point?
this function is only ever called with an error, so the rest is dead code

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        name: 'count_items_metricsCount',
+        help: 'Number of entities with data usage',
+        labelNames: ['metricLevel'],
+    });
+/**
+ * @param {http.ServerResponse} res - http response object
+ * @param {Error | errors.ArsenalError} error - Error
+ * @return {void}
+ */
+function _writeResponse(res, error) {
```
francoisferrand, line 41: `res` is not used...

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+        return res.writeHead(200, {
+            'content-length': contentLen,
+            'content-type': promClient.register.contentType,
+        }).end(promMetrics);
+    } catch (ex) {
+        return _writeResponse(res, ex);
+    } finally {
+        if (CountMaster.waitingForPromScraping === true) {
+            countMasterInstance.stop(null, onScraped);
+        }
```
francoisferrand: to reduce coupling (and avoid circular deps between modules), would be best to move this test to the callback:

```suggestion
        if (onScraped) {
            onScraped();
        }
```

utils/S3UtilsMongoClient.js
```diff
@@ -150,6 +151,7 @@ class S3UtilsMongoClient extends MongoClientInterface {
                             entry: res,
                             error,
                         });
+                        monitoring.objectsCount.inc({ state: 'error' });
```
francoisferrand: seems like the label is not a state of the object : more like `result` or `status`

utils/constants.js
```diff
@@ -1,3 +1,5 @@
 module.exports = {
     MaxParallelLimit: 10,
+    frequency: 30,
+    threshold: 5,
```
francoisferrand: * these parameters are unintellegible → make them explicit (`prometheusPollingFrequency`...), or just make the computation once and have just a "final scrapping timeout" variable
* the values should not be hard-coded here, but set by ZKOP: i.e. passed via env variables (though we may have defaults obviously)

→ should be top-level const in `CountItems/masterProcess.js`

utils/monitoring.js
```diff
@@ -0,0 +1,113 @@
...
+    || new promClient.Histogram({
+        name: 'count_items_consolidationDuration',
+        help: 'Duration of metrics consolidation in seconds',
+        labelNames: ['service'],
+        buckets: [0.01, 0.05, 0.1, 0.2, 0.5, 1, 1.5, 2],
+    });
+
+const workersCount = (promClient.register.getSingleMetric('count_items_workersCountErrors'))
+    || new promClient.Counter({
+        name: 'count_items_workersCountErrors',
```
francoisferrand:
```suggestion
        name: 's3_countitems_worker_errors_count',
```

...though it does not seem to be an error in "worker" : the metric should not represent "where in the code in fails", but "what operation" did actually fail...
→ a "worker" error would be for exemple if we failed to create a worker, or if a worker crashed
→ in this case, this is used when `getObjectMDStats` returns an error, i.e. when we fail to process a bucket: so it may be more a case of `s3_countitems_total_buckets_count` with a label indicating success or failure?

CountItems/CountMaster.js
```diff
@@ -1,11 +1,14 @@
...
 const { reshapeExceptionError } = require('arsenal').errorUtils;
+const { threshold, frequency } = require('../utils/constants');
 
 class CountMaster {
     constructor(params) {
         this.log = params.log;
         this.manager = params.manager;
         this.client = params.client;
+        this.metrics = params.metrics;
+        CountMaster.waitingForPromScraping = false;
```
francoisferrand: why do we need to put this here?
this like all this would fit very nicey in masterProcess directly, without splitting the implementation in 3 places:
* start() is called by masterProcess, it can also start the metrics server (can even be done immediately)
* start's `callback` is provided by masterProcess, so it can extend it to handle the termination

CountItems/CountManager.js
benzekrimaha: this part was removed as the `metricsCount` was redundant with already existing metrics , based on comment here  : https://github.com/scality/s3utils/pull/312#discussion_r1601060012

CountItems/CountManager.js
benzekrimaha: this part was removed as the metricsCount was redundant with already existing metrics , based on comment here : https://github.com/scality/s3utils/pull/312#discussion_r1601060012

utils/monitoring.js
benzekrimaha: this was a leftover from when I was trying to only collect clusterMetrics

utils/monitoring.js
benzekrimaha: removed here as we indeed only have count-items: https://github.com/scality/s3utils/pull/312/commits/be0f3e16b518690540561a94a498ff07dc3860b0#diff-612a4c10e405510cdaad879589d4a7d157df6e0ff42b4645d2a21d4d17e41fa9R21

utils/monitoring.js
benzekrimaha: fixed here : https://github.com/scality/s3utils/pull/312/commits/be0f3e16b518690540561a94a498ff07dc3860b0#diff-612a4c10e405510cdaad879589d4a7d157df6e0ff42b4645d2a21d4d17e41fa9R21

utils/monitoring.js
benzekrimaha: it is used here : https://github.com/scality/s3utils/pull/312/commits/be0f3e16b518690540561a94a498ff07dc3860b0#diff-612a4c10e405510cdaad879589d4a7d157df6e0ff42b4645d2a21d4d17e41fa9R21

utils/monitoring.js
benzekrimaha: this metric was removed

utils/S3UtilsMongoClient.js
benzekrimaha: changed to `status` : https://github.com/scality/s3utils/pull/312/commits/be0f3e16b518690540561a94a498ff07dc3860b0#diff-612a4c10e405510cdaad879589d4a7d157df6e0ff42b4645d2a21d4d17e41fa9R30-R34

utils/monitoring.js
benzekrimaha: done here : https://github.com/scality/s3utils/pull/312/commits/be0f3e16b518690540561a94a498ff07dc3860b0#diff-612a4c10e405510cdaad879589d4a7d157df6e0ff42b4645d2a21d4d17e41fa9R74-R76

CountItems/CountMaster.js
benzekrimaha: done here : https://github.com/scality/s3utils/pull/312/commits/be0f3e16b518690540561a94a498ff07dc3860b0#diff-46cbf734fd7b9022c7bbf597a8a5deb96829b581ca817a861bbc3bf1e86274a9R52-R82

utils/monitoring.js
benzekrimaha: the metic was removed

utils/monitoring.js
benzekrimaha: removed as this metric can be retrieved based on the comment here : https://github.com/scality/s3utils/pull/312#discussion_r1601020512

utils/monitoring.js
benzekrimaha: removed this metric as redudant based on comment here : https://github.com/scality/s3utils/pull/312#discussion_r1601020512

utils/constants.js
benzekrimaha: done here : https://github.com/scality/s3utils/pull/312/commits/a50c483216bfb00c5fa7bc9b25c02e1cbb673bbb#diff-46cbf734fd7b9022c7bbf597a8a5deb96829b581ca817a861bbc3bf1e86274a9R25-R32

utils/monitoring.js
benzekrimaha: renamed here : https://github.com/scality/s3utils/pull/312/commits/a50c483216bfb00c5fa7bc9b25c02e1cbb673bbb#diff-612a4c10e405510cdaad879589d4a7d157df6e0ff42b4645d2a21d4d17e41fa9R24-R28

utils/monitoring.js
```diff
@@ -0,0 +1,88 @@
...
+        return res.writeHead(200, {
+            'content-length': contentLen,
+            'content-type': promClient.register.contentType,
+        }).end(promMetrics);
+    } catch (ex) {
+        return _writeResponse(res, ex);
+    } finally {
+        if (onScraped) {
+            onScraped();
+        }
```
ghost, line 76:
```suggestion
        return onScraped?.();
```

Maybe we also want to make sure it's a function

utils/monitoring.js
francoisferrand: it is passed to the function indeed, but the function does not use it (in the body of the function!)

francoisferrand: yes my bad

CountItems/masterProcess.js
```diff
@@ -40,17 +47,18 @@ const countMaster = new CountMaster({
...
     client: new S3UtilsMongoClient(createMongoParams(log)),
-    metrics: metricServer,
 });
 
-metricServer.onRequest((req, res) => monitoring.metricsHandler(
-    countMaster,
+const metricServer = new WebServer(8003, log).onRequest((req, res) => monitoring.metricsHandler(
     () => {
-        process.exit(1);
+        if (CountMaster.waitingForPromScraping === true) {
```
francoisferrand: `waitingForPromScraping` does not belong to CountMaster : it is part of the "metrics handling", which is in masterProcess.js : a variable from this file can be set to true around line 79, before setting the timeout.

CountItems/masterProcess.js
```diff
@@ -18,6 +22,14 @@ const loggerConfig = {
 werelogs.configure(loggerConfig);
 const log = new werelogs.Logger('S3Utils::CountItems::Master');
 
+const prometheusPollingFrequency = process.env.PROMETHEUS_POLLING_FREQUENCY && !Number.isNaN(process.env.PROMETHEUS_POLLING_FREQUENCY)
+    ? Number.parseInt(process.env.PROMETHEUS_POLLING_FREQUENCY, 10)
```
francoisferrand: should we not try to `parseInt()`, and default if it fails, so we never end up with an invalid prometheusPollingFrequency?
otherwise we need to assert/abort/fail if the variable is not valid (e.g. prometheusPollingFrequency is a number greater than 0).

CountItems/masterProcess.js
```diff
@@ -18,6 +22,14 @@ const loggerConfig = {
...
+    : 30;
+
+const promtheusPollingThreshold = process.env.PROMETHEUS_POLLING_THRESHOLD && !Number.isNaN(process.env.PROMETHEUS_POLLING_THRESHOLD)
```
francoisferrand: can't speak of a "threshold" here : polling is an action, not a measurable...

not sure how to call that variable though, maybe having a single variable `PROMETHEUS_MAX_POLLING_DELAY` (instead of period + max number of periods) would be easier...

CountItems/masterProcess.js
```diff
@@ -18,6 +22,14 @@ const loggerConfig = {
 werelogs.configure(loggerConfig);
 const log = new werelogs.Logger('S3Utils::CountItems::Master');
 
+const prometheusPollingFrequency = process.env.PROMETHEUS_POLLING_FREQUENCY && !Number.isNaN(process.env.PROMETHEUS_POLLING_FREQUENCY)
```
francoisferrand: nit: the value "30" is a duration (30s), so this is not the frequency but the period...

CountItems/masterProcess.js
```diff
@@ -18,6 +22,14 @@ const loggerConfig = {
...
+    ? Number.parseInt(process.env.PROMETHEUS_POLLING_FREQUENCY, 10)
+    : 30;
+
+const promtheusPollingThreshold = process.env.PROMETHEUS_POLLING_THRESHOLD && !Number.isNaN(process.env.PROMETHEUS_POLLING_THRESHOLD)
+    ? Number.parseInt(process.env.PROMETHEUS_POLLING_THRESHOLD, 10)
```
francoisferrand: same point about parsing

utils/monitoring.js
```diff
@@ -0,0 +1,88 @@
...
+    buckets: [1, 10, 60, 600, 3600, 18000, 36000],
+});
+
+const consolidationDuration = new promClient.Histogram({
+    name: 's3_countitems_bucket_merge_duration_seconds',
+    help: 'Duration of metrics consolidation in seconds',
+    buckets: [0.01, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10],
+});
+
+const workersCount = new promClient.Counter({
```
francoisferrand: is it `workers` or `buckets` ?

CountItems/CountWorker.js
```diff
@@ -35,7 +36,10 @@ class CountWorker {
         return async.waterfall([
             next => this.client._getIsTransient(bucketInfo, this.log, next),
             (isTransient, next) => this.client.getObjectMDStats(bucketName, bucketInfo, isTransient, this.log, next),
-        ], callback);
+        ], (err, results) => {
+            monitoring.workersCount.inc({ state: err ? 'error' : 'success' });
```
francoisferrand: * this is not a "state", it is a result or status...
* what is the benefit of this metrics vs the number of buckets metrics computed as part of the duration metrics, in CountMaster?

ghost (approved):
LGTM (after currently open comments are resolved); just a small comment

utils/monitoring.js
benzekrimaha: renamed to buckets indeed this was leftover when I updated the description : https://github.com/scality/s3utils/pull/312/commits/094b51c921935a4def339cbd48b61b2f8cf40a3d#diff-612a4c10e405510cdaad879589d4a7d157df6e0ff42b4645d2a21d4d17e41fa9R24

CountItems/CountWorker.js
benzekrimaha: it is now status -> https://github.com/scality/s3utils/pull/312/commits/094b51c921935a4def339cbd48b61b2f8cf40a3d#diff-612a4c10e405510cdaad879589d4a7d157df6e0ff42b4645d2a21d4d17e41fa9R24

this one gives insights in case of errors which is not really shown in the duration metric

CountItems/masterProcess.js
benzekrimaha: renamed prometheusPollingPeriod : https://github.com/scality/s3utils/pull/312/commits/7a546979810051418f9af9bc3ed2567a0bc69eb4#diff-46cbf734fd7b9022c7bbf597a8a5deb96829b581ca817a861bbc3bf1e86274a9R27-R28

CountItems/masterProcess.js
benzekrimaha: rename to `prometheusPollingAttempts` ( any better naming idea is welcome ... ) => https://github.com/scality/s3utils/pull/312/commits/7a546979810051418f9af9bc3ed2567a0bc69eb4#diff-46cbf734fd7b9022c7bbf597a8a5deb96829b581ca817a861bbc3bf1e86274a9R31-R32

CountItems/masterProcess.js
benzekrimaha: done here : https://github.com/scality/s3utils/pull/312/commits/7a546979810051418f9af9bc3ed2567a0bc69eb4#diff-46cbf734fd7b9022c7bbf597a8a5deb96829b581ca817a861bbc3bf1e86274a9R27-R29

CountItems/masterProcess.js
benzekrimaha: done here : https://github.com/scality/s3utils/pull/312/commits/7a546979810051418f9af9bc3ed2567a0bc69eb4#diff-46cbf734fd7b9022c7bbf597a8a5deb96829b581ca817a861bbc3bf1e86274a9R27-R29

CountItems/masterProcess.js
benzekrimaha: done here : https://github.com/scality/s3utils/pull/312/commits/7a546979810051418f9af9bc3ed2567a0bc69eb4#diff-46cbf734fd7b9022c7bbf597a8a5deb96829b581ca817a861bbc3bf1e86274a9R81

utils/monitoring.js
francoisferrand: why would it not be a function: if it's provided but not a function, this is out of the contract, and will not behave as expected anyway...
(similar to every other callback : we almost never check if they are indeed callbacks...)

ghost: It's more to be on the safe side: we are not using typescript, the argument seems to be optional, but first in the list, as we have a condition on its value... Given that it's called outside of the try/catch, it would throw an error to call a function if the object is not one, independently from any contract we might have (maybe we can reuse the function in other s3utils scripts...). And we already had bugs where callbacks were "not a function" in some cases...

francoisferrand: that is my point: i don't think it helps...
* if somebody passes a value instead of the callback, he expects something with it (either he though this was a callback, or he wanted to pass extra info): so silently ignoring the unexpected type will just bury the bug...
* if we throw an error, the process will crash (or at the very least display a log about the uncaught exception) which will be easier to notice while testing
* obviously this if the code is not tested at all (either manually or automated), then all odds are off...

yes typescript will be better ("mandatory" compile stage would immediately detect the mismatch), but since we don't have it we must be more thorough when _testing_ instead of trying to handle unrecoverable errors...

ghost (commented):
Adding a comment as discussed: we need to set the counters/gauges to 0 at the beginning of a count-items run, otherwise the values will only increase

utils/monitoring.js
ghost: To me, the function should not be dependent on the whole project implementation:
* First, because we don't test a lot s3utils
* Second, because s3utils acts as a repo of multiple independent projects: we don't want any kind of contract with the rest of the implementation, we just want to make sure the object is a function before calling it...
* And third, because our functions should be robust, introducing such contract is only known by the ones implementing the function... In JS we can have plenty of reason where an object can be undefined, and we don't want to throw because of that. If we still want to log something, we can log it in this case using  `error` level.

Anyway it's not blocking, as I approved the PR

CountItems/masterProcess.js
```diff
@@ -15,9 +19,19 @@ const loggerConfig = {
...
 werelogs.configure(loggerConfig);
 const log = new werelogs.Logger('S3Utils::CountItems::Master');
 
+const prometheusPollingPeriod = Number.parseInt(process.env.PROMETHEUS_POLLING_PERIOD, 10) > 0
+    ? Number.parseInt(process.env.PROMETHEUS_POLLING_PERIOD, 10)
+    : 30;
+
+const prometheusPollingAttempts = Number.parseInt(process.env.PROMETHEUS_POLLING_ATTEMPTS, 10) > 0
+    ? Number.parseInt(process.env.PROMETHEUS_POLLING_ATTEMPTS, 10)
+    : 5;
```
francoisferrand: should introduce a function to avoid duplicated parsing (i.e. opportunity for typo):

```suggestion
function tryParseInt(s, defaultValue) {
  const v = Number.parseInt(s, 10);
  return v > 0 ? v : defaultValue;
}

const prometheusPollingPeriod = tryParseInt(process.env.PROMETHEUS_POLLING_PERIOD, 30);
const prometheusPollingAttempts = tryParseInt(process.env.PROMETHEUS_POLLING_ATTEMPTS, 5);
```

CountItems/masterProcess.js
```diff
@@ -37,6 +51,26 @@ const countMaster = new CountMaster({
...
+metricServer.start();
+
+// Set bucketsCount to 0
+monitoring.bucketsCount.labels('error').set(0);
+monitoring.bucketsCount.labels('success').set(0);
+
+// Set objectsCount to 0
+monitoring.objectsCount.labels('error').set(0);
+monitoring.objectsCount.labels('success').set(0);
+monitoring.objectsCount.labels('skipped').set(0);
```
francoisferrand: why do you reset the metric to 0?
prometheus series should not be "pre-generated", but only created when the label value is used.

benzekrimaha: Like discussed the approach was to switch to a gauge , this has been dropped

CountItems/masterProcess.js
benzekrimaha: Indeed , done here : https://github.com/scality/s3utils/pull/312/commits/902a90af79f05379d2e978a489b3df5929c615e5 

_please note that all commits will be squashed once the approvals complete_

utils/monitoring.js
```diff
@@ -0,0 +1,88 @@
...
+
+const consolidationDuration = new promClient.Histogram({
+    name: 's3_countitems_bucket_merge_duration_seconds',
+    help: 'Duration of metrics consolidation in seconds',
+    buckets: [0.01, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10],
+});
+
+const bucketsCount = new promClient.Counter({
+    name: 's3_countitems_total_buckets_count',
+    help: 'Buckets processed by workers state',
```
francoisferrand: nit:
```suggestion
    help: 'Total number of buckets processed',
```

utils/monitoring.js
```diff
@@ -0,0 +1,88 @@
...
+    labelNames: ['status'],
+});
+
+const objectsCount = new promClient.Counter({
+    name: 's3_countitems_total_objects_count',
+    help: 'Number of objects processed',
```
francoisferrand:
```suggestion
    help: 'Total number of objects processed',
```

francoisferrand (approved):

benzekrimaha:
@bert-e create_pull_requests

PR #317, opened 2024-05-18
Improvement/s3 utils 163 fixup

ghost:
Small fixup to properly get the pre-scan inflights, otherwise all inflights are in the metrics at the end of the job

### DIFF
--- a/utils/S3UtilsMongoClient.js
+++ b/utils/S3UtilsMongoClient.js
@@ -197,9 +197,9 @@ class S3UtilsMongoClient extends MongoClientInterface {
                 return callback(errors.InternalError);
             }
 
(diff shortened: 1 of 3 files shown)


utils/S3UtilsMongoClient.js
```diff
@@ -199,7 +199,7 @@ class S3UtilsMongoClient extends MongoClientInterface {
 
             const bucketEntry = usersBucketCreationDatesMap[`${bucketInfo.getOwner()}${constants.splitter}${bucketName}`];
```
francoisferrand: Nit: not a bucket entry, more like bucketDate

francoisferrand (approved):

PR #320, opened 2024-06-12
Bugfix/s3 utils 167

ghost:
Fixes a bug that would cause a crash of the job when there is no prom metric to get from a worker, with `clusterMetrics`, causing a non-Arsenal error, and causing the bug fixed here.

The root cause of the issue was not reproduced. I suspect inconsistencies in the platform, as we also have to explicitly specify the `useUnifiedTopology` option, although it's here by default, in the image, and working within all other deployments.

### DIFF
--- a/utils/monitoring.js
+++ b/utils/monitoring.js
@@ -1,12 +1,9 @@
 const { errors } = require('arsenal');
 const promClient = require('prom-client');
-const { Registry } = require('prom-client');
 const { http } = require('httpagent');
-const CountMaster = require('../CountItems/CountMaster');
 
 const aggregatorRegistry = new promClient.AggregatorRegistry();
 const { collectDefaultMetrics } = promClient;
-
 
 // Histogram of the bucket processing duration, by the utilization service.
 const bucketProcessingDuration = new promClient.Histogram({
@@ -40,7 +37,7 @@ const objectsCount = new promClient.Counter({
  */
 function _writeResponse(res, error) {
     let statusCode = 500;
-    if (error instanceof errors.ArsenalError && Number.isInteger(error.code)) {
+    if (error && Number.isInteger(error.code)) {
         statusCode = error.code;
     }
     res.writeHead(statusCode, { 'Content-Type': 'application/json' });
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3utils",
-  "version": "1.14.10",
+  "version": "1.14.11",
   "engines": {
     "node": ">= 16"
   },


francoisferrand (dismissed):

utils/monitoring.js
```diff
@@ -40,7 +37,7 @@ const objectsCount = new promClient.Counter({
  */
 function _writeResponse(res, error) {
     let statusCode = 500;
-    if (error instanceof errors.ArsenalError && Number.isInteger(error.code)) {
+    if (error && Number.isInteger(error.code)) {
```
francoisferrand, line 40: why would it crash when error is not of type `ArsenalError` ?
this should just cause the `error instanceof errors.ArsenalError` to be false, and thus keep the  default statusCode 500...

ghost: Not sure exactly why, but I suspect there is something odd on the platform where we had the issue (not discussing it here - it's a public repo), we got:
```
{"name":"S3Utils::CountItems::Master","time":1717691607622,"error":{"message":"Right-hand side of 'instanceof' is not an object","stack":"TypeError: Right-hand side of 'instanceof' is not an object\n    at _writeResponse (/usr/src/app/utils/monitoring.js:43:15)\n    at Object.metricsHandler (/usr/src/app/utils/monitoring.js:72:16)","name":"TypeError"},"level":"error","message":"Uncaught Exception","hostname":"artesca-data-ops-count-items-111111-phm8q","pid":12}
```

I suspect this is the same issue that require us to put the `useUnifiedTopology` explicitly, even if this should be true by default in the cluster. There is something wrong somewhere in their setup.

Meanwhile, we do not need to ensure it's an arsenal error anyway, so just removing this code, it will unblock the client till we understand the issue better (if we want to)

benzekrimaha (approved):

utils/monitoring.js
ghost: And just to complete,a s said in the ticket, I was not able to reproduce the issue, and customers who did the upgrade are not complaining for now either.

francoisferrand: looking at the error : `Right-hand side of 'instanceof' is not an object`, it would seem that this is not a platform issue but really a coding bug

i.e. what is really `errors.ArsenalError` ? is it really defined (or a typo) ?
did you try (in some actually executed code path) to execute the same expression (`error instanceof errors.ArsenalError`) ?

ghost: This check is not something we want to have, and this is not something I could trigger even when simulating no metric from a worker. Hence its removal, is there is a bug? Maybe, but we do not need this check in any case, so not worth digging into it

But to answer, this is a class that exists. iDE finds it as well. I triggered this line with my tests with success - no error, same docker image. It's like the MongoDB option, which works on Artesca 1.7 without any change, yet the same driver version and MongoDB version.

francoisferrand: > This check is not something we want to have

if we don't need to have it, indeed it can and should be removed.

> but we do not need this check in any case, so not worth digging into it

sorry to disagree, but it is worth digging. We have some difference in production, supposedly with the same (immutable!) docker image: so there is really an issue worth investigating....

francoisferrand (approved):

utils/monitoring.js
ghost: Yes, not saying we shouldn't try to understand why we have that, but it should be investigated globally, not for this particular bugfix

ghost: I created https://scality.atlassian.net/browse/OS-832 to track this problem.

PR #322, opened 2024-06-26
Count items crashes if a bucket has a website configuration

ghost:
This bug is fixed by nullifying nay website configuration before creating the Arsenal object, as we do not need anything related to websites in count items.

Issue: S3UTILS-168

### DIFF
--- a/CountItems/CountWorker.js
+++ b/CountItems/CountWorker.js
@@ -30,6 +30,11 @@ class CountWorker {
         if (!this.client.client) {
             return callback(new Error('NotConnected'));
         }
+        // 'fromObj' expects that the website configuration is an instance of
+        // WebsiteConfiguration as it is not used in CountItems, we nullify it.
+        if (bucketInfoObj._websiteConfiguration) {
+            Object.assign(bucketInfoObj, { _websiteConfiguration: null });
+        }
         const bucketInfo = BucketInfo.fromObj(bucketInfoObj);
         const bucketName = bucketInfo.getName();
         this.log.info(`${process.pid} handling ${bucketName}`);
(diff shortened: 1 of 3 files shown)


CountItems/CountWorker.js
```diff
@@ -30,6 +30,11 @@ class CountWorker {
         if (!this.client.client) {
             return callback(new Error('NotConnected'));
         }
+        // 'fromObj' expects that the website configuration is an instance of
+        // WebsiteConfiguration As it it not used in CountItems, we nullify it.
```
francoisferrand: why would be not be a `WebsiteConfiguration` ? There is probably another root cause, would be best to fix it instead of masking it with a band-aid...

ghost: It's not band-aid: the function we use, `fromObj`, loads an object and expects the website configuration to be of the right instance, it is not the `deSerialize` method, where we accept a non-instanciated JSON object. The latter could be used, but we would need to serialize the bucket metadata object. I chose to keep the same function, but adapt the object to pass the class `assert`s... (less computations)

> why would be not be a WebsiteConfiguration ?

Because this object comes from MongoDB, and has no specific class inside, it's just an object with fields.

ghost: I will also definitely not update arsenal for the count-items job, that we really need to drop today...

benzekrimaha:
```suggestion
        // WebsiteConfiguration As it is not used in CountItems, we nullify it.
```

benzekrimaha (approved):
LGTM

francoisferrand (approved):

PR #325, opened 2024-09-19
bf: S3UTILS-180 unbounded number of sproxyd keys in Map

jonathan-gramain:
- Fix missing transmission of the environment variable that limits the number of keys kept in the in-memory maps
  (`OBJECT_REPAIR_DUPLICATE_KEYS_WINDOW_SIZE`)

- Add `Number.parseInt()` calls to environment variables where needed

- Sanitize a little bit the `env.js`:

  - reorder lexicographically

  - set default value of `OBJECT_REPAIR_RAFT_LOG_BEGIN_CSEQ` to `null` to make ObjectRepair use the computed default when fetched (instead of having it declared separately from `env.js`)

Manually tested with a mocked oplog server returning entries with sproxyd keys:

- without the fix: memory usage grows quickly (multiple GB after a few minutes)

- with the fix: memory stabilizes at a level proportional to the value of `OBJECT_REPAIR_DUPLICATE_KEYS_WINDOW_SIZE`

- not specifying `OBJECT_REPAIR_DUPLICATE_KEYS_WINDOW_SIZE` in the environment: stabilizes memory usage around the default value of 100000 entries.


jonathan-gramain:
ping

dvasilas (approved):

BourgoisMickael (approved):

PR #326, opened 2024-10-03
Fix `MongoClientInterface.getBucketInfos` to ignore views

francoisferrand:
- Ensure getBucketInfos acces only mongo collections
- add debug ssh runner
- Bump arsenal 8.1.137

Issue: S3UTILS-178


ghost (approved):
maybe we want to release a new version of s3utils with this fix (so, need to bump package version)?

welansari (approved):

PR #329, opened 2024-11-28
Fix handling of system.* reserved collections

francoisferrand:
- Bump arsenal 8.1.141
- Use `getBucketInfos` from arsenal
- Release 1.14.16

Issue: S3UTILS-182

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3utils",
-  "version": "1.14.15",
+  "version": "1.14.16",
   "engines": {
     "node": ">= 16"
   },
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "dependencies": {
     "@senx/warp10": "^1.1.2",
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/arsenal#8.1.137",
+    "arsenal": "git+https://github.com/scality/arsenal#8.1.142",
     "async": "^2.6.4",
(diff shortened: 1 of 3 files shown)


utils/S3UtilsMongoClient.js
```diff
@@ -838,62 +838,6 @@ class S3UtilsMongoClient extends MongoClientInterface {
         }
     }
 
-    /*
```
ghost, line 841: I don't remember why we had a re-implementation of this function, were we doing more logic here? Just making sure we are not deleting something we should not

francoisferrand: it looks like what is described:

> Overwrite the getBucketInfos method to specially handle the cases that bucket collection exists but bucket is not in metastore collection.
> For now, to make the count-items cronjob more robust, we ignore those "bad buckets"

But this is now the behavior in arsenal

package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@senx/warp10": "^1.1.2",
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/arsenal#8.1.137",
+    "arsenal": "git+https://github.com/scality/arsenal#8.1.141",
```
benzekrimaha: can't we bump to 8.1.142 as it's the latest release ?

KillianG (approved):

ghost (approved):

PR #331, opened 2024-12-19
Fix triggering replication when multiple destinations are set

welansari:
- Issue 1: In Artesca "SITE_NAME" is never used, so we always
trigger objects that are replicated to the first storageClass in
the replication rule.

- Issue 2: We check the global replication status when verifying
wether or not an object should be retriggered. This doesn't necessarily
work all the time, especially when replicating to multiple destinations.
As if one destination fails the global status becomes failed, which will
make it impossible to trigger objects with a completed status for example.

- Issue 3: replication info is completely overwritten when it does not contain
info about a specific site. This will cause an issue when replicating to multiple
destinations as the script can only be launched for one site at a time, so when
having a object with non initialized replication info, we won't be able to set
the replication info properly for all destinations.

Issue: S3UTILS-184

### DIFF
--- a/CRR/ReplicationStatusUpdater.js
+++ b/CRR/ReplicationStatusUpdater.js
@@ -114,16 +114,19 @@ class ReplicationStatusUpdater {
      * Determines if an object should be updated based on its replication metadata properties.
      * @private
      * @param {ObjectMD} objMD - The metadata of the object.
+     * @param {string} site - The destination site name.
      * @returns {boolean} True if the object should be updated.
      */
-    _objectShouldBeUpdated(objMD) {
+    _objectShouldBeUpdated(objMD, site) {
         return this.replicationStatusToProcess.some(filter => {
             if (filter === 'NEW') {
+                // Either site specific replication info is missing
+                // or are initialized with empty fields.
                 return (!objMD.getReplicationInfo()
-                    || objMD.getReplicationInfo().status === '');
+                    || !objMD.getReplicationSiteStatus(site));
             }
             return (objMD.getReplicationInfo()
-                && objMD.getReplicationInfo().status === filter);
+                && objMD.getReplicationSiteStatus(site) === filter);
         });
     }
 
@@ -172,35 +175,53 @@ class ReplicationStatusUpdater {
                 //   codebase easier to maintain and upgrade, as opposed to having multiple branches or versions of
                 //   the code for different schema versions.
                 objMD = new ObjectMD(JSON.parse(mdRes.Body));
-                if (!this._objectShouldBeUpdated(objMD)) {
+                if (!this._objectShouldBeUpdated(objMD, storageClass)) {
                     skip = true;
                     return process.nextTick(next);
                 }
                 // Initialize replication info, if missing
                 // This is particularly important if the object was created before
                 // enabling replication on the bucket.
-                if (!objMD.getReplicationInfo()
-                    || !objMD.getReplicationSiteStatus(storageClass)) {
+                let replicationInfo = objMD.getReplicationInfo();
+                if (!replicationInfo || !replicationInfo.status) {
                     const { Rules, Role } = repConfig;
                     const destination = Rules[0].Destination.Bucket;
                     // set replication properties
                     const ops = objMD.getContentLength() === 0 ? ['METADATA']
                         : ['METADATA', 'DATA'];
-                    const backends = [{
+                    replicationInfo = {
+                        status: 'PENDING',
+                        content: ops,
+                        backends: [],
+                        destination,
+                        storageClass: '',
+                        role: Role,
+                        storageType: '',
+                    };
+                    objMD.setReplicationInfo(replicationInfo);
+                }
+                // Update replication info with site specific info
+                if (!objMD.getReplicationSiteStatus(storageClass)) {
+                    // When replicating to multiple destinations,
+                    // the storageClass and storageType properties
+                    // become comma-separated lists of the storage
+                    // classes and types of the replication destinations.
+                    const storageClasses = objMD.getReplicationStorageClass()
+                        ? `${objMD.getReplicationStorageClass()},${storageClass}` : storageClass;
+                    objMD.setReplicationStorageClass(storageClasses);
+                    if (this.storageType) {
+                        const storageTypes = objMD.getReplicationStorageType()
+                            ? `${objMD.getReplicationStorageType()},${this.storageType}` : this.storageType;
+                        objMD.setReplicationStorageType(storageTypes);
+                    }
+                    // Add site to the list of replication backends
+                    const backends = objMD.getReplicationBackends();
+                    backends.push({
                         site: storageClass,
                         status: 'PENDING',
                         dataStoreVersionId: '',
-                    }];
-                    const replicationInfo = {
-                        status: 'PENDING',
-                        backends,
-                        content: ops,
-                        destination,
-                        storageClass,
-                        role: Role,
-                        storageType: this.storageType,
-                    };
-                    objMD.setReplicationInfo(replicationInfo);
+                    });
+                    objMD.setReplicationBackends(backends);
                 }
 
                 objMD.setReplicationSiteStatus(storageClass, 'PENDING');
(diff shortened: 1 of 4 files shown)


CRR/ReplicationStatusUpdater.js
```diff
@@ -114,16 +114,19 @@ class ReplicationStatusUpdater {
      * Determines if an object should be updated based on its replication metadata properties.
      * @private
      * @param {ObjectMD} objMD - The metadata of the object.
+     * @param {string} storageClass - The storage class for replication.
```
francoisferrand: `location` or `site` ?
the `storageClass` is really just where the object is "locally" stored, not where we will copy/send it...

CRR/ReplicationStatusUpdater.js
```diff
@@ -172,36 +175,54 @@ class ReplicationStatusUpdater {
...
                 }
                 // Initialize replication info, if missing
                 // This is particularly important if the object was created before
                 // enabling replication on the bucket.
-                if (!objMD.getReplicationInfo()
-                    || !objMD.getReplicationSiteStatus(storageClass)) {
+                let replicationInfo = objMD.getReplicationInfo();
+                if (!replicationInfo || !replicationInfo.status) {
                     const { Rules, Role } = repConfig;
                     const destination = Rules[0].Destination.Bucket;
```
francoisferrand, line 188: this may not be correct, if there is more than 1 desitination....

CRR/ReplicationStatusUpdater.js
```diff
@@ -172,36 +175,54 @@ class ReplicationStatusUpdater {
...
                         status: 'PENDING',
-                        backends,
                         content: ops,
+                        backends: [],
                         destination,
-                        storageClass,
+                        storageClass: '',
                         role: Role,
-                        storageType: this.storageType,
+                        storageType: '',
```
francoisferrand, line 199: what are these 2 fields (`storageClass` and `storageType`) used for?

CRR/ReplicationStatusUpdater.js
welansari: `crrExistingObjects` doesn't currently support triggering replication for multiple sites at once. So this would work as it will initialize the replication info when triggering the first site and then append the other sites' info to the fields when the other ones are triggered.

CRR/ReplicationStatusUpdater.js
welansari: `storageClass` is what each location's replication queue processor uses to check if it should replicate an object or not. 
`storageType` is used by Cloudserver in Backbeat routes to do some pre-checks (check if versioning is supported on the backend and that the location is valid)

KillianG (approved):

CRR/ReplicationStatusUpdater.js
francoisferrand: that is how these fields are used... my question is really what information do they actually store, what these fields represent (for example, the storageClass of the object is already known, and stored in `.location[]` and `dataStoreName` ; and this cannot be the storageClass of the 'remote' object, since there may be multiple destinations...)

same for `destination` btw, I don't understand this field initialize with a single bucket name :-/ Or maybe this is a left-over from the first replication (not supporting multi-targets)?

CRR/ReplicationStatusUpdater.js
francoisferrand: even without triggering multiple sites at once : if a user requests replicatoin to the second "site", this will initialize with the 1st destination (--> trigger replication!), then we will add the second one...

When there is no replicationInfo (i.e. typically when it is empty), should we not just initialize an empty replicationInfo, and let the next block ("Update replication info with site specific info") fill the details for the requested destination?

CRR/ReplicationStatusUpdater.js
francoisferrand: the code which initializes these in cloudserver:
```
function _getReplicationInfo(rule, replicationConfig, content, operationType,
    objectMD, bucketMD) {
    const storageTypes = [];
    const backends = [];
    const storageClasses = _getStorageClasses(rule);
    if (!storageClasses) {
        return undefined;
    }
    storageClasses.forEach(storageClass => {
        const storageClassName =
              storageClass.endsWith(':preferred_read') ?
              storageClass.split(':')[0] : storageClass;
        const location = s3config.locationConstraints[storageClassName];
        if (location && replicationBackends[location.type]) {
            storageTypes.push(location.type);
        }
        backends.push(_getBackend(objectMD, storageClassName));
    });
    if (storageTypes.length > 0 && operationType) {
        content.push(operationType);
    }
    return {
        status: 'PENDING',
        backends,
        content,
        destination: replicationConfig.destination,
        storageClass: storageClasses.join(','),
        role: replicationConfig.role,
        storageType: storageTypes.join(','),
        isNFS: bucketMD.isNFS(),
    };
}
```

--> it seems these fields (along with `role`) are not multi-destination/rule aware?
--> these should be mostly useless for Zenko's multibackend replication (otherwise we probably have bugs), but may still be needed for CRR (and may cause bugs if multiple rules/destinations were used in that case)

CRR/ReplicationStatusUpdater.js
francoisferrand: in cloudserver, it seems this is initialized to `bucketMD.replicationConfig.destination` : should we do the same?

welansari: This is a weird one... So, we currently only support having a single destination bucket for all replication rules of a bucket. When creating the rules/workflows via UI it's even worse, the destination bucket becomes the name of the bucket we are replicating from.
The value of this field is not used in Zenko. When creating a location, Cloudserver initializes a client class for the respective backend (aws/azure/...) that keeps the name of the destination bucket in memory, that's the value we use when replicating and not what's in the replication rule (only the storageClass is used to know which client to use).

CRR/ReplicationStatusUpdater.js
welansari: `storageClass` and `storageType` are multi-destination aware (although their name suggests otherwise), they contain the list of all storage classes and storage types we are replicating to.
Example:
```
storageClass: "aws-location,azure-blob"
storageType: "aws_s3,azure"
```

In Zenko, these fields are duplicates of information we already have, as backbeat/cloudserver have a list of all location information, we could just use the storage class stored in the rules to get the info we want. I think these are more of a relic from S3C that we can't really remove right now as S3C uses them.

In CRR, the role is also a list. I don't think this works in Zenko tho.
The `destination` field is a weird one, i explained how it works in the previous comment

francoisferrand: > these are more of a relic from S3C that we can't really remove right now as S3C uses them
> In CRR, the role is also a list. I don't think this works in Zenko tho.

"soon", we will have a single arsenal, cloudserver & backbeat : so we can start planning to clean this up, possibly removing (deprecating) redudant fields and aligning the way CRR and multi-backend replication store/manage the state?

can you please create a tech debt ticket, to track this (future) work and document the problem/situation/state of the analysis?

CRR/ReplicationStatusUpdater.js
```diff
@@ -172,36 +175,54 @@ class ReplicationStatusUpdater {
...
                     };
                     objMD.setReplicationInfo(replicationInfo);
                 }
+                // Update replication info with site specific info
+                if (objMD.getReplicationSiteStatus(storageClass) === undefined) {
```
francoisferrand:
```suggestion
                if (!objMD.getReplicationSiteStatus(storageClass)) {
```

CRR/ReplicationStatusUpdater.js
```diff
@@ -114,16 +114,19 @@ class ReplicationStatusUpdater {
...
             if (filter === 'NEW') {
+                // Either site specific replication info is missing
+                // or are initialized with empty fields.
                 return (!objMD.getReplicationInfo()
-                    || objMD.getReplicationInfo().status === '');
+                    || !objMD.getReplicationSiteStatus(site));
             }
             return (objMD.getReplicationInfo()
-                && objMD.getReplicationInfo().status === filter);
+                && objMD.getReplicationSiteStatus(site) === filter);
```
francoisferrand, line 129: is this true as well for CRR (in s3c) ?

esp. we should check that there is no "legacy" data format where we would not have the per-site status, and where we should rely only on the global state...

(e.g. maybe if CRR sometimes has objet which are marked with `replicationInfo.status === 'COMPLETE'` and `site` listed in `storageClass`, but no site status...)

francoisferrand: @nicolas2bert @jonathan-gramain what do you think?

francoisferrand (approved):

francoisferrand:
code-wise LGTM, but there may be some corner with S3C/CRR, to be confirmed...

CRR/ReplicationStatusUpdater.js
welansari: Replication metadata are managed the same in both S3C and Zenko so yes this should work fine.
I looked back to the initial format of the metadata, there was always a global status.

PR #333, opened 2024-12-23
CountItems to consider MPUs in storage metrics

ghost:
- Sum of all MPU parts as part of the `current` metric for the associated bucket
- Follow the same logic as the usual object to ensure we properly map the parts to the right accounts
- Use the overview keys to count the number of incomplete/pending objects

The logic to process the object was kept inline in the function as the number of accessed variables is high, to avoid unnecessary complexity: the function is already unit-tested and hopefully dropped soon or later...

Issue: S3UTILS-186

### DIFF
--- a/CountItems/utils/utils.js
+++ b/CountItems/utils/utils.js
@@ -33,7 +33,7 @@ function consolidateDataMetrics(target, source) {
                 _currentRestoring: 0,
                 _nonCurrentRestored: 0,
                 _nonCurrentRestoring: 0,
-                _incompleteMPUParts: 0,
+                _incompleteMPUUploads: 0,
                 deleteMarker: 0,
             },
         });
@@ -61,7 +61,7 @@ function consolidateDataMetrics(target, source) {
     resTarget.objectCount._currentRestored += objectCount && objectCount._currentRestored ? objectCount._currentRestored : 0;
     resTarget.objectCount._nonCurrentRestoring += objectCount && objectCount._nonCurrentRestoring ? objectCount._nonCurrentRestoring : 0;
     resTarget.objectCount._nonCurrentRestored += objectCount && objectCount._nonCurrentRestored ? objectCount._nonCurrentRestored : 0;
-    resTarget.objectCount._incompleteMPUParts += objectCount && objectCount._incompleteMPUParts ? objectCount._incompleteMPUParts : 0;
+    resTarget.objectCount._incompleteMPUUploads += objectCount && objectCount._incompleteMPUUploads ? objectCount._incompleteMPUUploads : 0;
 
     resTarget.usedCapacity._inflightsPreScan += usedCapacity && usedCapacity._inflightsPreScan ? usedCapacity._inflightsPreScan : 0;
     if (accountOwnerID) {
@@ -75,7 +75,7 @@ function consolidateDataMetrics(target, source) {
         ? usedCapacity._nonCurrentCold + usedCapacity._nonCurrentRestored + usedCapacity._nonCurrentRestoring : 0;
     resTarget.objectCount.current += objectCount
         ? objectCount._currentCold + objectCount._currentRestored + objectCount._currentRestoring
-        + objectCount._incompleteMPUParts : 0;
+        + objectCount._incompleteMPUUploads : 0;
     resTarget.objectCount.nonCurrent += objectCount
         ? objectCount._nonCurrentCold + objectCount._nonCurrentRestored + objectCount._nonCurrentRestoring : 0;
 
--- a/CountItems/utils/utils.js
+++ b/CountItems/utils/utils.js
@@ -18,6 +18,7 @@ function consolidateDataMetrics(target, source) {
                 _nonCurrentRestored: 0,
                 _nonCurrentRestoring: 0,
                 _inflightsPreScan: 0,
+                _incompleteMPUParts: 0,
             },
         });
     }
@@ -32,6 +33,7 @@ function consolidateDataMetrics(target, source) {
                 _currentRestoring: 0,
                 _nonCurrentRestored: 0,
                 _nonCurrentRestoring: 0,
+                _incompleteMPUParts: 0,
                 deleteMarker: 0,
             },
         });
@@ -48,6 +50,7 @@ function consolidateDataMetrics(target, source) {
     resTarget.usedCapacity._currentRestored += usedCapacity && usedCapacity._currentRestored ? usedCapacity._currentRestored : 0;
     resTarget.usedCapacity._nonCurrentRestoring += usedCapacity && usedCapacity._nonCurrentRestoring ? usedCapacity._nonCurrentRestoring : 0;
     resTarget.usedCapacity._nonCurrentRestored += usedCapacity && usedCapacity._nonCurrentRestored ? usedCapacity._nonCurrentRestored : 0;
+    resTarget.usedCapacity._incompleteMPUParts += usedCapacity && usedCapacity._incompleteMPUParts ? usedCapacity._incompleteMPUParts : 0;
 
     resTarget.objectCount.current += objectCount && objectCount.current ? objectCount.current : 0;
     resTarget.objectCount.nonCurrent += objectCount && objectCount.nonCurrent ? objectCount.nonCurrent : 0;
@@ -58,6 +61,7 @@ function consolidateDataMetrics(target, source) {
     resTarget.objectCount._currentRestored += objectCount && objectCount._currentRestored ? objectCount._currentRestored : 0;
     resTarget.objectCount._nonCurrentRestoring += objectCount && objectCount._nonCurrentRestoring ? objectCount._nonCurrentRestoring : 0;
     resTarget.objectCount._nonCurrentRestored += objectCount && objectCount._nonCurrentRestored ? objectCount._nonCurrentRestored : 0;
+    resTarget.objectCount._incompleteMPUParts += objectCount && objectCount._incompleteMPUParts ? objectCount._incompleteMPUParts : 0;
 
     resTarget.usedCapacity._inflightsPreScan += usedCapacity && usedCapacity._inflightsPreScan ? usedCapacity._inflightsPreScan : 0;
     if (accountOwnerID) {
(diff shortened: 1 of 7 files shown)


CountItems/utils/utils.js
```diff
@@ -48,6 +50,7 @@ function consolidateDataMetrics(target, source) {
     resTarget.usedCapacity._currentRestored += usedCapacity && usedCapacity._currentRestored ? usedCapacity._currentRestored : 0;
     resTarget.usedCapacity._nonCurrentRestoring += usedCapacity && usedCapacity._nonCurrentRestoring ? usedCapacity._nonCurrentRestoring : 0;
     resTarget.usedCapacity._nonCurrentRestored += usedCapacity && usedCapacity._nonCurrentRestored ? usedCapacity._nonCurrentRestored : 0;
+    resTarget.usedCapacity._incompleteMPUParts += usedCapacity && usedCapacity._incompleteMPUParts ? usedCapacity._incompleteMPUParts : 0;
```
francoisferrand, line 53: nit:

```suggestion
    resTarget.usedCapacity._incompleteMPUParts += usedCapacity?._incompleteMPUParts || 0;
```

utils/S3UtilsMongoClient.js
```diff
@@ -342,18 +362,75 @@ class S3UtilsMongoClient extends MongoClientInterface {
                         });
                         return callback(err);
```
francoisferrand: when `callback` is called with error, we should not process the MPUs anymore (which would call `callback` eventually as well)

utils/S3UtilsMongoClient.js
```diff
@@ -342,18 +362,75 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
                     }
-                    const bucketStatus = bucketInfo.getVersioningConfiguration();
-                    const isVer = (bucketStatus && (bucketStatus.Status === 'Enabled'
-                        || bucketStatus.Status === 'Suspended'));
                     const retResult = this._handleResults(collRes, isVer);
                     retResult.stalled = stalledCount;
                     return callback(null, retResult);
```
francoisferrand: should not be called here, we are not done (still need to process MPU bucket) ?

utils/S3UtilsMongoClient.js
```diff
@@ -342,18 +362,75 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                            method: 'getObjectMDStats',
+                            errDetails: { ...err },
+                            errorString: err.toString(),
+                        });
+                        return callback(err);
+                    }
+                    const retResult = this._handleResults(collRes, isVer);
+                    retResult.stalled = stalledCount;
+                    return callback(null, retResult);
+                },
```
francoisferrand: multiple calls to `callback` : after processing both the cursor and mpu cursor, and once more eventually after the inflight processing...

....which may actually not really be the issue: looking at the documentation of mongodb driver, I don't see any mention of a second callback: so I am thinking this may be a left-over from the upgrade to promises, and this callback should just be removed (errors will raise an exception, caught eventually; and `handleResults` is called for normal case at line 434)

utils/S3UtilsMongoClient.js
```diff
@@ -207,132 +214,145 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                            if (!collRes[metricLevel][resourceName]) {
+                                collRes[metricLevel][resourceName] = {
                                     ...baseMetricsObject,
                                 };
                             }
-                            collRes.account[account].locations[location][targetData] += data.location[location];
-                            collRes.account[account].locations[location][targetCount]++;
-                            collRes.account[account].locations[location].deleteMarkerCount += res.value.isDeleteMarker ? 1 : 0;
+                            collRes[metricLevel][resourceName][targetData] += data[metricLevel][resourceName];
+                            // Do not count the MPU parts as objects
```
francoisferrand, line 322: not an issue in this PR, but worth considering for the future: eventually, do we want to report MPU by number of uploads (i.e. "potential objects") or actual parts (which are stored internally as separate documents, and actually each reference some data)...

thinking about this,
* it may actually be better to count each part as an object : so we can also report on left-over parts even if the overview key is missing. Semantics may not be so good (part vs object), but I'd rather report something that more closely matches the storage than the user's business logic for now, and thus not mask any issue.
* as far as semantics go, `mpuPartsCount` should be a number of parts: if we count only the overview keys, should be something like `mpuUploadsCount` instead (here and in other field names)

A compromise may be to count (and store) `partsCount`, `uploadsCount` and `partsSize`, but only aggregate `uploadsCount` in object count and `partsSize` in object size. But not sure it is worth the extra effort: may be best to just keep it "simple", counting and measuring individual _parts_ and reporting them like objects...

utils/S3UtilsMongoClient.js
```diff
@@ -207,132 +214,145 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                } else if (isObjectRestoring) {
+                    targetCount += 'Restoring';
+                    targetData += 'Restoring';
+                } else if (isObjectRestored) {
+                    targetCount += 'Restored';
+                    targetData += 'Restored';
+                }
+
+                if (isMPUPart || isOverviewKey) {
+                    targetCount = 'mpuPartCount';
```
francoisferrand: typo:
```suggestion
                    targetCount = 'mpuPartsCount';
```

tests/unit/CountItems/utils/utils.js
```diff
@@ -23,6 +24,7 @@ describe('CountItems::utils::consolidateDataMetrics', () => {
             _currentRestoring: 0,
             _nonCurrentRestored: 0,
             _nonCurrentRestoring: 0,
+            _incompleteMPUParts: 0,
```
francoisferrand: as implemented _today_ (see my other comment), this is not the count of `parts` but the count of `uploads`, so the field should be named `_incompleteMPUUploads`. Or may be better instead to actually count parts.

utils/S3UtilsMongoClient.js
ghost: > not an issue in this PR, but worth considering for the future

I formally disagree that more effort should be put into this obsolete script. But it's worth discussing how to count MPUs for Scuba.

CountItems/utils/utils.js
ghost: Not possible with s3utils today

utils/S3UtilsMongoClient.js
ghost: I will remove the err callbacks, they are not used with mongodb driver v5 indeed, so no impact

tests/unit/CountItems/utils/utils.js
ghost: Counting parts does not make sense from a client point of view, as they cannot control, most of the time, how the MPUs are splited. What is important is knowing how many MPUs are incomplete and how much data is occupied by these MPUs. It's anyway not important here, because our only use case are the quotas (only on storage bytes) and reflecting the current usage in the UI (no count of objects here as well). We can however consider it for Scuba, in a separate work.

KillianG (approved):

utils/S3UtilsMongoClient.js
francoisferrand: > But it's worth discussing how to count MPUs for Scuba.

that is my point: this discussion is not about the component where it is implemented (utapi, s3utils, scuba, ...) but really about the semantics and data we want to measure.

tests/unit/CountItems/utils/utils.js
francoisferrand: as discussed above, i seems presomptuous to say what makes or does not make sense for a client point of view: anyway the APIs are limited, it is not presented to the user...

the request from product (as a proxy of customers) today is really just to count the "size" used by incomplete MPU (parts or uploads is the same here). As for the number of parts or uploads, it actually fits different uses, both for the "customers" (but different personas):
* number of parts may makes more sense for an admin, which wants to understand why mongo is overloaded;
* number of uploads may make more sense for the user which performed the upload, which wants to understand how many extra uploads he did... but maybe he does not care so much about number of uploads, and more about the amount of data he has to re-upload...

i also don't know what is expected, and it will need to be considered for Scuba : but as soon as you cristalize a behavior here and it gets shipped, customers may start to use it, and we will have to support it... So we must really refrain from adding something quickly just because we can, esp. if the semantics may be ambiguous.

utils/S3UtilsMongoClient.js
```diff
@@ -207,153 +214,197 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                            }
+                            collRes[metricLevel][resourceName].deleteMarkerCount += entry.value.isDeleteMarker ? 1 : 0;
                         });
+                    }
+                });
+                Object.keys(data.account).forEach(account => {
+                    if (!collRes.account[account].locations) {
+                        collRes.account[account].locations = {};
+                    }
+
```
francoisferrand, line 334: nit: introduce variable
```suggestion
const locations = collRes.account[account].locations;
```

francoisferrand (approved):

tests/unit/CountItems/utils/utils.js
ghost: What I mean is that in the S3 world we would typically use ListMultipartUploads to get the number of in-progress/incomplete MPUs: this is not returning all the part. If we need the parts we can use ListParts.

S3 doesn't expose the number of parts either [in their metrics](https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage_lens_metrics_glossary.html?trk=el_a134p000006vpQjAAI&trkCampaign=AWSInsights_Website_Docs_AmazonS3-storage_lens_metrics_glossary&sc_channel=el&sc_campaign=AWSInsights_Blog_discovering-and-deleting-incomplete-multipart-uploads-to-lower-&sc_outcome=Product_Marketing), but:

- Incomplete Multipart Upload Storage Bytes – The total bytes in scope with incomplete multipart uploads
- Incomplete Multipart Upload Object Count – The number of objects in scope that are incomplete multipart uploads

So for sure the number of upload will be required for us to be standard. We cannot reflect more storage utilization if we have 0 object reported (for example, reporting 1TB of data with 0 object if everything is only incomplete MPUs): there needs to be a correlation between the two. And the number of upload is the natural (and standard) information to have.

Then maybe we can consider the number of parts, not supporting them here is actually what I suggested and what you seem to align on: if we ship it we'll need it in SUR, yet it may not be needed, or hard to track within scuba...

ghost:
ping

PR #334, opened 2025-01-17
Bugfix/s3 utils 188 support metrics above max safe

benzekrimaha:


ghost:
The PR was updated to :
- Compute all metrics with bigints
- Store the storage metrics as Long in MongoDB
- Store the global metrics (in the document with `_id` `countitems`) as numbers, to not change anything. The values here will be imprecise but much more work is required if we want to support it properly everywhere.

package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@senx/warp10": "^1.1.2",
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/arsenal#8.1.142",
+    "arsenal": "git+https://github.com/scality/arsenal#d59d00d28b96748bd0099b50f6cfb84827a49dc6",
```
ghost: To be changed with a proper tag before merging

ghost:
PR updated with a unit test to ensure we compute the values properly when we have bigints. Too hard to test as functional test: we won't be uploading large objects in the CI.

CountItems/CountWorker.js
```diff
@@ -17,6 +17,35 @@ class CountWorker {
         this.handleMessage = this.handleMessage.bind(this);
     }
 
+    _serializeBigInts(obj) {
```
ghost: Note for reviewers:

The idea is that in s3utils we have a master-worker approach, where workers send their results to the master. This process serializes the results; yet the master has a need to know what values are bigint, and what are strings.

As such, these functions (and the equivalent ones in the master process class) will automatically use a special field, `__bigint`, to dynamically detect that an object is in fact a bigint.

ghost: @benzekrimaha we might want to dedup this, with some static functions in a util file

Edit: it's done

DataReport/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -74,22 +78,28 @@ function collectBucketMetricsAndUpdateBucketCapacityInfo(mongoClient, log, callb
...
+                        let capacity = -1n;
+                        if (isValidCapacityValue(Capacity)) {
+                            // is Capacity value is valid
                             capacity = Capacity;
                             // if bucket storage used is valid and capacity is bigger than used
-                            if (bucketStorageUsed !== -1 && (capacity - bucketStorageUsed) >= 0) {
+                            if (bucketStorageUsed !== -1n && (capacity - bucketStorageUsed) >= 0n) {
                                 available = capacity - bucketStorageUsed;
+                            } else if (bucketStorageUsed !== -1n && (capacity - bucketStorageUsed) < 0n) {
+                                available = 0n;
```
ghost, line 102: The reason of this change is because the sos api spec defines that -1 is for "ongoing computations" or metrics that are unavailable. Having -1 in this case means that SOSAPI will eventually be disabled, or more data will come, depending on Veeam implementation. Having 0 makes it explicit that we reached the maximum capacity for the bucket.

CountItems/utils/utils.js
```diff
@@ -9,79 +9,111 @@ function consolidateDataMetrics(target, source) {
...
-    resTarget.usedCapacity.nonCurrent += usedCapacity && usedCapacity.nonCurrent ? usedCapacity.nonCurrent : 0;
-    resTarget.usedCapacity._currentCold += usedCapacity && usedCapacity._currentCold ? usedCapacity._currentCold : 0;
-    resTarget.usedCapacity._nonCurrentCold += usedCapacity && usedCapacity._nonCurrentCold ? usedCapacity._nonCurrentCold : 0;
-    resTarget.usedCapacity._currentRestoring += usedCapacity && usedCapacity._currentRestoring ? usedCapacity._currentRestoring : 0;
-    resTarget.usedCapacity._currentRestored += usedCapacity && usedCapacity._currentRestored ? usedCapacity._currentRestored : 0;
-    resTarget.usedCapacity._nonCurrentRestoring += usedCapacity && usedCapacity._nonCurrentRestoring ? usedCapacity._nonCurrentRestoring : 0;
-    resTarget.usedCapacity._nonCurrentRestored += usedCapacity && usedCapacity._nonCurrentRestored ? usedCapacity._nonCurrentRestored : 0;
-    resTarget.usedCapacity._incompleteMPUParts += usedCapacity && usedCapacity._incompleteMPUParts ? usedCapacity._incompleteMPUParts : 0;
+    resTarget.usedCapacity.current += BigInt(usedCapacity?.current || 0n);
```
francoisferrand, line 45: not sure we should take the chance and remove these, but it seems to me these `BigInt` conversions should not be useful : since these fields are only managed by `_processEntryData()` (which already handles bigint) and this function...

CountItems/utils/utils.js
```diff
@@ -9,79 +9,111 @@ function consolidateDataMetrics(target, source) {
...
+        ? BigInt(objectCount._currentCold) + BigInt(objectCount._currentRestored) + BigInt(objectCount._currentRestoring)
+        + BigInt(objectCount._incompleteMPUUploads) : 0n;
     resTarget.objectCount.nonCurrent += objectCount
-        ? objectCount._nonCurrentCold + objectCount._nonCurrentRestored + objectCount._nonCurrentRestoring : 0;
+        ? BigInt(objectCount._nonCurrentCold) + BigInt(objectCount._nonCurrentRestored) + BigInt(objectCount._nonCurrentRestoring) : 0n;
 
     return resTarget;
 }
 
+
```
francoisferrand: nit: extra empty line

CountItems/utils/utils.js
```diff
@@ -9,79 +9,111 @@ function consolidateDataMetrics(target, source) {
...
+function serializeBigInts(obj) {
```
francoisferrand, line 85: nit: performance-wise, it is better to have such generic functions or custom functions for the specific objects we convert? (since we it is actually used for either `BucketInfo` or `BucketInfo[]`) ?

tests/functional/utils/S3UtilsMongoClient.js
```diff
@@ -1581,17 +1613,17 @@ describe('S3UtilsMongoClient::getObjectMDStats', () => {
     });
 });
 
-describe('S3UtilsMongoClient::updateBucketCapacityInfo', () => {
+describe.skip('S3UtilsMongoClient::updateBucketCapacityInfo', () => {
```
francoisferrand: test is skipped

tests/functional/utils/S3UtilsMongoClient.js
```diff
@@ -1644,17 +1676,17 @@ describe('S3UtilsMongoClient::updateBucketCapacityInfo', () => {
     });
 });
 
-describe('S3UtilsMongoClient::getBucketInfos', () => {
+describe.skip('S3UtilsMongoClient::getBucketInfos', () => {
```
francoisferrand: test is skipped

tests/functional/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -94,8 +94,8 @@ describe('collectBucketMetricsAndUpdateBucketCapacityInfo', () => {
                 next();
             }),
             next => client.getBucketAttributes(testBucketName, logger, (err, bucketInfo) => {
-                assert.equal(err, null);
-                assert.equal(bucketInfo.getCapabilities(), null);
+                assert.equal(err && err.message, 'InternalError');
+                assert.equal(bucketInfo, undefined);
                 next();
```
francoisferrand: seems weird to expect an `InternalError` on `getBucketAttributes()`...
(and it does not really to the test description)

utils/S3UtilsMongoClient.js
```diff
@@ -104,20 +114,19 @@ class S3UtilsMongoClient extends MongoClientInterface {
                 const id = entry._id;
                 if (id.startsWith('bucket_')) {
                     const inflightDocument = inflightsMap[id];
-                    const inflight = Long.fromNumber(Number(inflightDocument ? Math.max(0, inflightDocument - entry.usedCapacity._inflightsPreScan) : 0));
+                    const inflight = inflightDocument ? BigIntMax(0, inflightDocument - entry.usedCapacity._inflightsPreScan) : 0n;
```
```suggestion
                    const inflight = inflightDocument ? BigIntMax(0n, inflightDocument - entry.usedCapacity._inflightsPreScan) : 0n;
```

utils/S3UtilsMongoClient.js
```diff
@@ -13,37 +13,43 @@ const USERSBUCKET = '__usersbucket';
 const INFOSTORE_TMP = `${INFOSTORE}_tmp`;
 const __COUNT_ITEMS = 'countitems';
 
+const BigIntMax = (...args) => args.reduce((max, current) => {
+    const maxAsBigInt = typeof max === 'bigint' ? max : BigInt(max);
+    const currentAsBigInt = typeof current === 'bigint' ? current : BigInt(current);
+    return maxAsBigInt > currentAsBigInt ? max : current;
```
francoisferrand: * nit: same as earlier comment: it seems to be this type-check & BigInt conversion should not be needed...
* if in doubt, should probably just call BigInt constructor anyway, to be consistent (if there a perf impact we try to mitigate, then we should do it elsewhere as well...)

```suggestion
    const maxAsBigInt = BigInt(max);
    const currentAsBigInt = BigInt(current);
    return maxAsBigInt > currentAsBigInt ? max : current;
```

utils/S3UtilsMongoClient.js
```diff
@@ -754,11 +766,23 @@ class S3UtilsMongoClient extends MongoClientInterface {
         });
 
         return {
-            versions: Math.max(0, totalNonCurrentCount + totalNonCurrentColdCount + totalVersionRestoringCount + totalVerionsRestoredCount),
-            objects: totalCurrentCount + totalCurrentColdCount + totalRestoringCount + totalRestoredCount,
+            // We do not support bigint for the global count items document
```
francoisferrand, line 769: what limitation are we adressing?
* here we perform the computation, and it will be written to `Long` anyway in mongo : so should not affect countItems
* if we convert to Number eventually, might as well do all calculus in number...
* practically, there is no immediate risk: we are not close to the limit on the number of objects (our DB will probably break first)

e.g. I am fine to either do the computation in bigint or in number, but it's a waste to do it in BigInt to eventually convert to number while a Long is stored in Mongo...

ghost: It's because we store these values as numbers today, and migrating this is not urgent here. Contrary to what I previously said, the driver does not enforce a type automatically, so even if we exceed the max safe size, we re still storing a number. The `Long` are coming from a previous change we made.

So the issue here is that we also store the sum of all the managed data by location, and this may be bigger than a number -> supporting long will require changes in pensieve-api, cloudserver, and possibly more? So it's not about avoiding the change, but reduce as much as possible the risks here.

Example:
```
  {
    _id: 'countitems',
    value: {
      objects: 105027,
      versions: 0,
      buckets: 1,
      bucketList: [
        ...
      ],
      dataManaged: {
        total: { curr: 115465213591495740, prev: 0 },
        byLocation: {
          '9a0c45b5-759a-11ef-9efd-52d84ee9ecd5': { curr: 115465213591495740, prev: 0 }
        }
      },
      stalled: 0
    }
  },
```

tests/functional/collectBucketMetricsAndUpdateBucketCapacityInfo.js
ghost: I updated the test, it was not the right check

ghost: Interesting point though, it was passing for the wrong reasons: the assert in the callback, when failing, would never reach the `done`, but instead, throw n error that will be catched by whatever can do it; and here, it was in the mongoclientinterface class... so we end up calling this test callback a second time, and with the right error.

Note that this is not the first time i see this and we have this problem in almost all our node codebase.

CountItems/utils/utils.js
ghost: As discussed I believe this is minor as ran once per bucket, and eventually it allows us to be flexible in the future, if we were to add more fields.

CountItems/utils/utils.js
ghost: I tried, but we end up with functional tests failures, maybe @benzekrimaha you will have a bit more info on this?

ghost (approved):

CountItems/utils/utils.js
francoisferrand: is it worth investigating more, to catch issue or just leave the code cleaner?
(if you are confident it works well enough, but not happy with the code, we could merge "as is" and investigate after the release)

utils/S3UtilsMongoClient.js
francoisferrand: so we should add this as well in the list of "followup works"

tests/functional/utils/S3UtilsMongoClient.js
```diff
@@ -451,7 +464,26 @@ describe('S3UtilsMongoClient::getObjectMDStats', () => {
                             logger,
                             (err, data) => {
                                 assert.deepStrictEqual(err, null);
-                                assert.deepStrictEqual(data, expected);
+                                // assert.deepStrictEqual(data, expected);\
```
francoisferrand: commented-out code

CountItems/utils/utils.js
ghost: I think it's safe and it's more like we don't pass the right values to this function in the test, so not in the actual program, having passing tests would be rethink them. Better be extra-safe here, as this is the most sensitive part of our changes.

tests/functional/utils/S3UtilsMongoClient.js
ghost: It was meant to be removed in favor of the below lines, so I removed it.

tests/functional/collectBucketMetricsAndUpdateBucketCapacityInfo.js
```diff
@@ -160,16 +160,16 @@ describe('collectBucketMetricsAndUpdateBucketCapacityInfo', () => {
...
+                assert.strictEqual(Used, 20n);
                 next();
             }),
         ], done);
     });
 
-    test('should update bucket Capacity and Available -1 if Capacity value is not valid', done => {
-        testBucketCapacities.VeeamSOSApi.CapacityInfo.Capacity = 'not-a-number';
+    test('should update bucket Capacity and Available -1 if Capacity value is not set', done => {
+        testBucketCapacities.VeeamSOSApi.CapacityInfo.Capacity = -1;
```
francoisferrand: should it be `-1n` ?
or is it the case of an "old" bucket with capacity pre-dating BigInt change?

ghost: -1n, but yes it works with the previous values as well, as we also accept numbers.

francoisferrand (approved):

williamlardier:
ping

ghost:
ping

PR #335, opened 2025-03-25
Support per-location metrics in bucket

ghost:
We had the account->locations mapping, but not for buckets. This PR updates the logic so we have the per-location info for buckets as well.

Issue: S3UTILS-170


utils/S3UtilsMongoClient.js
```diff
@@ -765,6 +784,67 @@ class S3UtilsMongoClient extends MongoClientInterface {
             });
         });
 
+        // parse all location and reflect the data in the bucket
+        Object.keys((res.bucket || {})).forEach(bucket => {
```
anurag4DSB, line 790: Can there be a scenario where `res.bucket[bucket]` is undefined but dataMetrics.bucket[bucket] exists?

utils/S3UtilsMongoClient.js
```diff
@@ -765,6 +784,67 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+            if (!dataMetrics.bucket[bucket].locations) {
```
anurag4DSB, line 791: This also assumes dataMetrics.bucket[bucket] is always defined.

utils/S3UtilsMongoClient.js
```diff
@@ -765,6 +784,67 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                        current: 0n,
+                        nonCurrent: 0n,
+                        _currentCold: 0n,
+                        _nonCurrentCold: 0n,
+                        _currentRestored: 0n,
+                        _currentRestoring: 0n,
+                        _nonCurrentRestored: 0n,
+                        _nonCurrentRestoring: 0n,
+                        _incompleteMPUUploads: 0n,
+                        deleteMarker: 0n,
```
anurag4DSB, line 823: a helper function might help here to `createEmptyUsedCapacity()` to avoid repetition.

anurag4DSB (approved):
This code assumes that all the fields in dataMetrics.location[location].usedCapacity and objectCount exist and are BigInts. If that's okay, LGTM, left some comments.

utils/S3UtilsMongoClient.js
ghost: No, as dataMetrics is computed from the `res` object. What is possible is to have no bucket, but the commented line handles it.

utils/S3UtilsMongoClient.js
ghost: Yes, but here we also rely on the dataMetrics that we computed from `res` . We are missing types here (TS) to make it clearer... Anyway the same approach is used since years for the accounts, without any problem.

utils/S3UtilsMongoClient.js
ghost: Yes and no, the issue with a function is that we would need to deep copy the actual value dynamically (object + bigint requires it), to avoid shallow copy. While it works and reduce code duplication, it's way less optimized for the v8 engine running node, especially here where we do it a lot of time...

anurag4DSB: Thanks for the detailed context — totally agree that in this hot path, performance takes precedence and explicit allocation avoids the overhead of deep copies and potential deopts in V8.

That said, since this pattern is duplicated multiple times and tightly coupled to the shape of the metrics object, it might be worth leaving a code comment explaining why we’re avoiding a helper. This would help future readers (or linters/automated tools) understand that this isn’t just accidental duplication but a deliberate performance optimization.

Totally good with leaving it as-is, but calling this out explicitly makes the decision durable across time and reviewers 👌

utils/S3UtilsMongoClient.js
anurag4DSB: 👍 Got it — since dataMetrics is always derived from res, that guarantees the alignment here.

utils/S3UtilsMongoClient.js
ghost: I added a commit.
(I also rebased, we'll use the 1.15 branch from now on)

francoisferrand: certainly not worth digging further in this code that we want to get rid off ASAP, but just want to point out that V8 also includes very powerful JIT and optimization engine... So the same usual guideline should be applied: best to focus on code correctness & readability, and then _measure_ to see where optimization is needed and what is the most optimized solution...

francoisferrand (approved):

francoisferrand (approved):

PR #337, opened 2025-05-06
Improvement/s3 utils 191

benzekrimaha:
Issue: S3UTILS-191


package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@senx/warp10": "^2.0.3",
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/arsenal#8.2.8",
+    "arsenal": "git+https://github.com/scality/arsenal#22a5462d7d9bdec0d21a0e9fb484594af35c1cee",
```
benzekrimaha: Please note that this will be switched to the arsenal adequate version once the current in review PR with a fix is merged: https://github.com/scality/Arsenal/pull/2371

CountItems/CountWorker.js
```diff
@@ -27,6 +28,40 @@ class CountWorker {
...
+        async.waterfall([
+            next => this.client.getObject(PENSIEVE, overlayVersionId, null, this.log, next),
+            (version, next) => {
+                const overlayConfigId = `configuration/overlay/${version}`;
+                return this.client.getObject(PENSIEVE, overlayConfigId, null, this.log, next);
+            },
+        ], (err, res) => {
+            if (err) {
+                this.log.error('error getting configuration overlay', {
+                    method: '_getIsTransient',
```
ghost:
```suggestion
                    method: 'pensieveLocationIsTransient',
```

ghost (approved):

tests/unit/CountItems/CountWorker.js
```diff
@@ -180,13 +174,11 @@ describe('CountItems::CountWorker', () => {
             sendFn: testSendFn,
             client: mongoMock,
         });
+        w.getIsTransient = jest.fn((bucketInfo, cb) => cb(null, true));
```
francoisferrand: just mocking `getIsTransient` means we basically do not execute this code...

- Best to have some test exercising the "whole" code
- If needed, we may use a mock (in some tests, ideally not all) to simulate specific conditions/results/...
- In any case, need extra tests cases to verify the new functions behave somewhat correctly (esp. do not crash) and the information ("is transient") is used appropriately

benzekrimaha: done here : https://github.com/scality/s3utils/pull/337/commits/8f20ab69c13634aeafa55bab8f7c4f69bb846cc5#diff-c0f7a781647ac82835f751f3f2acc5767cca6b22e6f41de1539ec22c29ffcb32R220

tests/unit/CountItems/CountWorker.js
```diff
@@ -213,15 +205,52 @@ describe('CountItems::CountWorker', () => {
...
-        mongoMock._getIsTransient.mockImplementationOnce((_a, _b, cb) => cb(null, true));
         mongoMock.getObjectMDStats.mockImplementationOnce((_a, _b, _c, _d, cb) => cb(null, { value: 42 }));
         w.countItems(bucketInfo, (err, results) => {
             expect(err).toBeNull();
             expect(results).toEqual({ value: 42 });
             done();
         });
     });
+
+    test('should use real getIsTransient and not crash', done => {
```
francoisferrand: should have at least 2 scenarios, depending on `this.client.isLocationTransient` (since there are multiple code path).

the point is not just to see that `countItems` is calling `getIsTransient`, but really to execute the code in `getIsTransient` to ensure it will not fail in production... (it may be done though other tests, calling the function directly!)

(alternatively, since we use a wrapper of MongoClient in here: if you add the function there they can be tested in mongoclient tests, and you don't need to change the countWorker tests)

tests/unit/CountItems/CountWorker.js
francoisferrand: Looking at the tests, it does not cover all branches of the code.

In addition, these tests are overly complicated to tests 2 very simple functions : would seem much easier (both to write and to review) to have "simpler" tests to exercise all code path in `getIsTransient()` (and the nested `pensieveLocationIsTransient`) by just calling this function.

tests/unit/CountItems/CountWorker.js
```diff
@@ -213,15 +205,89 @@ describe('CountItems::CountWorker', () => {
...
+        mongoMock.getObjectMDStats.mockImplementationOnce((_a, _b, _c, _d, cb) => cb(null, { value: 42 }));
+        w.countItems(bucketInfo, (err, results) => {
+            expect(err).toBeNull();
+            expect(results).toEqual({ value: 42 });
+            done();
+        });
+    });
+
+    test('should use real getIsTransient and not crash', done => {
+        const testSendFn = jest.fn();
```
francoisferrand: test name feels weird : there is no "feature" of the code to use a "real getIsTransient", it is really just that `getIsTransient` is mocked in all tests...
--> should find a name that better matches what "feature" of the code this test exercises (or take this as a hint that indeed we are mixing approaches here, and we shoudl either not use a mock in the tests -focus on e2e function, covering all cases in data-, or keep the mock to test the "counting" part and create other tests to verify `getIsTransient` works as expected)

tests/unit/CountItems/CountWorker.js
```diff
@@ -22,7 +21,6 @@ describe('CountItems::CountWorker', () => {
                     setup: [null],
                     close: [],
                     isConnected: false,
-                    getIsTransient: [],
```
francoisferrand: in the previous code, we used to have different `getIsTransient()` mock results depending on the test case : i.e. either `[]` (i.e. no error, not transient) or `[null, true]` (i.e. no error, transient)

However, with your change the mock always returns `true` : and we don't actually test the "usual" scenario we are in...

francoisferrand (approved):

PR #340, opened 2025-07-16
S3UTILS-195 verifyBucketSproxydKeys.js crashes on empty object without location

nicolas2bert:
Context:

There is a known bug in Metadata v0 with null-compatible mode that can result in the creation of invalid or broken versions such as:

`"value": "{\"versionId\":\"99999999999999999999RG001  \",\"isNull\":true}"`

These entries appear as versions but are missing critical metadata like a valid locations array. They are likely causing the script to break or produce incorrect results. This PR adds stricter checks and error logging to show and handle those cases explicitly.

This PR:
  - Ensures that objects with non-zero content-length have a valid, non-empty `locations` array.
  - Logs an error and updates status counters for objects with broken metadata.

Goals:
- Catch and log objects with inconsistent or broken location metadata early in the scan.
- Improve robustness of the bucket consistency verification tool.

### DIFF
--- a/VerifyBucketSproxydKeys/validateLocations.js
+++ b/VerifyBucketSproxydKeys/validateLocations.js
@@ -1,0 +1,31 @@
+const { Logger } = require('werelogs');
+
+const log = new Logger('s3utils:validateLocations');
+
+/**
+ * Validates that locations array is valid for objects with non-zero content-length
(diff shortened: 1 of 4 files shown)


BourgoisMickael (approved):

anurag4DSB (approved):

nicolas2bert:
@bert-e approve

PR #341, opened 2025-07-25
crrExistingObjects: support only triggering current versions

welansari:
This is needed when using multiple-backend replication, as triggering
all versions may result in an ordering missmatch between versions on
both sites. This is due to the parallel execution of PUTs to the destination,
there is no guarantee the order of versions in source will be respected,
the latest version PUT in destination becomes the master.

More context [available here](https://scality.atlassian.net/browse/S3UTILS-183?focusedCommentId=423762).

Issue: S3UTILS-183

### DIFF
--- a/CRR/ReplicationStatusUpdater.js
+++ b/CRR/ReplicationStatusUpdater.js
@@ -1,7 +1,6 @@
 const {
     doWhilst, eachSeries, eachLimit, waterfall, series,
 } = require('async');
-const werelogs = require('werelogs');
 const { ObjectMD } = require('arsenal').models;
 
 const { setupClients } = require('./clients');
@@ -44,6 +43,7 @@ class ReplicationStatusUpdater {
             maxScanned,
             keyMarker,
             versionIdMarker,
+            currentVersionOnly,
         } = params;
 
         // inputs
@@ -61,6 +61,7 @@ class ReplicationStatusUpdater {
         this.maxScanned = maxScanned;
         this.inputKeyMarker = keyMarker;
         this.inputVersionIdMarker = versionIdMarker;
+        this.currentVersionOnly = currentVersionOnly;
         this.log = log;
 
         this._setupClients();
@@ -305,7 +306,12 @@ class ReplicationStatusUpdater {
                     this.log.warn(`missing SITE_NAME environment variable, triggering replication to the ${storageClass} storage class`);
                 }
                 return eachLimit(versions, this.workers, (i, apply) => {
-                    const { Key, VersionId } = i;
+                    const { Key, VersionId, IsLatest } = i;
+                    if (this.currentVersionOnly && !IsLatest) {
+                        ++this._nSkipped;
+                        apply();
+                        return;
+                    }
                     this._markObjectPending(bucket, Key, VersionId, storageClass, repConfig, apply);
                 }, next);
             },
--- a/crrExistingObjects.js
+++ b/crrExistingObjects.js
@@ -23,6 +23,7 @@ const MAX_SCANNED = (process.env.MAX_SCANNED
     && Number.parseInt(process.env.MAX_SCANNED, 10));
 const { KEY_MARKER } = process.env;
 const { VERSION_ID_MARKER } = process.env;
+const CURRENT_VERSION_ONLY = process.env.CURRENT_VERSION_ONLY === 'true' || process.env.CURRENT_VERSION_ONLY === '1' || false;
 
 const {
     ACCESS_KEY,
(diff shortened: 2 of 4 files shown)


CRR/ReplicationStatusUpdater.js
```diff
@@ -305,7 +306,18 @@ class ReplicationStatusUpdater {
                     this.log.warn(`missing SITE_NAME environment variable, triggering replication to the ${storageClass} storage class`);
                 }
                 return eachLimit(versions, this.workers, (i, apply) => {
-                    const { Key, VersionId } = i;
+                    const { Key, VersionId, IsLatest } = i;
+                    // Skip non-latest versions when CURRENT_VERSION_ONLY is enabled.
```
DarkIsDude: I'm not sure to understand the goal of the comment. The first part for me is not useful as you describe the if condition.
The note I don't understand why it's here as the call to the function is not done here

tests/unit/CRR/ReplicationStatusUpdater.js
DarkIsDude, line 1: Same for this file, comment just describe the code and no value are added with them

DarkIsDude (approved):

CRR/ReplicationStatusUpdater.js
welansari: fair enough, i removed the comments

tests/unit/CRR/crrExistingObjects.js
```diff
@@ -75,6 +75,7 @@ describe('crrExistingObjects', () => {
         process.env.KEY_MARKER = 'testKeyMarker';
         process.env.VERSION_ID_MARKER = 'testVersionIdMarker';
         process.env.DEBUG = '0';
+        process.env.CURRENT_VERSION_ONLY = 'true';
```
francoisferrand, line 78: this changes the behavior for CRR as well, which will affect S3C unexpectedly (and will be an issue when CRR is used in Zenko as well)

Can we make this behavior apply only for non-CRR replication(s) ?

francoisferrand: If we can't, and since we will now document that Zenko-replication does not guarantee ordering: we may keep the default behavior (replicate all versions), which matches the existing guarantees of Zenko replication (just replicate data) and may be fine for some backends (like AWS or Azure, which are not versioned) where we really just want to copy the data.

crrExistingObjects.js
```diff
@@ -23,6 +23,7 @@ const MAX_SCANNED = (process.env.MAX_SCANNED
     && Number.parseInt(process.env.MAX_SCANNED, 10));
 const { KEY_MARKER } = process.env;
 const { VERSION_ID_MARKER } = process.env;
+const CURRENT_VERSION_ONLY = process.env.CURRENT_VERSION_ONLY === 'true' || process.env.CURRENT_VERSION_ONLY === '1' || false;
```
francoisferrand, line 26:
```suggestion
const CURRENT_VERSION_ONLY = process.env.CURRENT_VERSION_ONLY === 'true' || process.env.CURRENT_VERSION_ONLY === '1';
```

tests/unit/CRR/crrExistingObjects.js
```diff
@@ -183,4 +186,28 @@ describe('crrExistingObjects', () => {
...
+            expect.objectContaining({
+                currentVersionOnly: true,
+            }),
+            expect.anything(),
+        );
+        expect(mockFatal).not.toHaveBeenCalled();
+        expect(mockError).not.toHaveBeenCalled();
+        expect(process.exit).not.toHaveBeenCalled();
+        expect(mockCrrRun).toHaveBeenCalled();
+    });
```
francoisferrand, line 212: shouldn't we have another test without specifying `process.env.CURRENT_VERSION_ONLY`, to verify that the (expected) default behavior is used?

tests/unit/CRR/crrExistingObjects.js
francoisferrand: should the "default" test not use the default behavior (replicate all versions), without setting `process.env.CURRENT_VERSION_ONLY` ?
(and a separate test to verify behavior when process.env.CURRENT_VERSION_ONLY is set to true)

francoisferrand (approved):

tests/unit/CRR/crrExistingObjects.js
welansari: This is tested in the test `should set the default parameter when unspecified`

tests/unit/CRR/crrExistingObjects.js
welansari: > this changes the behavior for CRR as well, which will affect S3C unexpectedly (and will be an issue when CRR is used in Zenko as well)

The default behaviour didn't change, so i don't see the problem this creates with CRR

> Can we make this behavior apply only for non-CRR replication(s) ?

A bit tricky to do in when using multiple destinations, doable if we only configured a single replication destination, however i'm not really sure we want to do this ? The param is fairly explicit, for me, it's up to the user of the script to decide on what mode they want to run it.

> should the "default" test not use the default behavior (replicate all versions), without setting process.env.CURRENT_VERSION_ONLY ?

If you're talking about the test `should run successfully` then i don't really think so as we don't run it with default values for the other params anyways. The test `should set the default parameter when unspecified` tests the behaviour with default params.

PR #346, opened 2025-08-25
Migration aws sdk v2 to v3 : crrExistingObject

SylvainSenechal:
Issue: S3UTILS-199

### DIFF
--- a/CRR/ReplicationStatusUpdater.js
+++ b/CRR/ReplicationStatusUpdater.js
@@ -1,9 +1,12 @@
 const {
-    doWhilst, eachSeries, eachLimit, waterfall, series,
+    doWhilst, eachSeries, eachLimit, waterfall,
 } = require('async');
 const { ObjectMD } = require('arsenal').models;
-
 const { setupClients } = require('./clients');
+const { 
+    ListObjectVersionsCommand, 
+    GetBucketReplicationCommand 
+} = require('@aws-sdk/client-s3');
 
 const LOG_PROGRESS_INTERVAL_MS = 10000;
 
@@ -26,6 +29,7 @@ class ReplicationStatusUpdater {
      * @param {number} [params.maxScanned] - (Optional) Maximum number of items to scan.
      * @param {string} [params.keyMarker] - (Optional) Key marker for resuming object listing.
      * @param {string} [params.versionIdMarker] - (Optional) Version ID marker for resuming object listing.
+     * @param {boolean} [params.currentVersionOnly] - (Optional) Whether to process only the current version of objects.
      */
     constructor(params, log) {
         const {
@@ -266,13 +270,15 @@ class ReplicationStatusUpdater {
      * @returns {void}
      */
     _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
-        return this.s3.listObjectVersions({
+        this.s3.send(new ListObjectVersionsCommand({
             Bucket: bucket,
             MaxKeys: this.listingLimit,
             Prefix: this.targetPrefix,
             VersionIdMarker,
             KeyMarker,
-        }, cb);
+        }))
+            .then(data => cb(null, data))
+            .catch(cb);
     }
 
     /**
@@ -284,15 +290,16 @@ class ReplicationStatusUpdater {
      * @returns {void}
      */
     _markPending(bucket, versions, cb) {
-        const options = { Bucket: bucket };
         waterfall([
-            next => this.s3.getBucketReplication(options, (err, res) => {
-                if (err) {
+            async () => {
+                try {
+                    const res = await this.s3.send(new GetBucketReplicationCommand({ Bucket: bucket }));
+                    return res.ReplicationConfiguration;
+                } catch (err) {
                     this.log.error('error getting bucket replication', { error: err });
-                    return next(err);
+                    throw err;
                 }
-                return next(null, res.ReplicationConfiguration);
-            }),
+            },
             (repConfig, next) => {
                 const { Rules } = repConfig;
                 const storageClass = this.siteName || Rules[0].Destination.StorageClass;
@@ -348,7 +355,8 @@ class ReplicationStatusUpdater {
                         this.log.error('error listing object versions', { error: err });
                         return done(err);
                     }
-                    return this._markPending(bucket, data.Versions.concat(data.DeleteMarkers), err => {
+                    const versions = (data.Versions || []).concat(data.DeleteMarkers || []);
+                    return this._markPending(bucket, versions, err => {
                         if (err) {
                             return done(err);
                         }
--- a/CRR/clients.js
+++ b/CRR/clients.js
@@ -1,10 +1,12 @@
-const AWS = require('aws-sdk');
+const { S3Client } = require('@aws-sdk/client-s3');
+const { NodeHttpHandler } = require('@aws-sdk/node-http-handler');
+const { ConfiguredRetryStrategy } = require('@smithy/util-retry');
 const http = require('http');
+const https = require('https');
 const BackbeatClient = require('../BackbeatClient');
 
 const AWS_SDK_REQUEST_DELAY_MS = 30;
 const AWS_SDK_REQUEST_RETRIES = 100;
-const LOG_PROGRESS_INTERVAL_MS = 10000;
 
 /**
  * Sets up and configures AWS S3 and Backbeat clients.
@@ -27,7 +29,37 @@ function setupClients({
     secretKey,
     endpoint,
 }, log) {
-    const awsConfig = {
+    const awsv3Config = {
+        region: 'us-east-1',
+        credentials: {
+            accessKeyId: accessKey,
+            secretAccessKey: secretKey,
+        },
+        endpoint,
+        forcePathStyle: true,
+        requestHandler: new NodeHttpHandler({
+            httpAgent: new http.Agent({ keepAlive: true }),
+            httpsAgent: new https.Agent({ 
+                keepAlive: true,
+                // With SDK v2, SSL was disabled. Here we added this
+                // to be able to run the script locally on a lab for testing,
+                // so https will work but there will be no security enforced
+                rejectUnauthorized: false
+            }),
+            requestTimeout: 60000,
+        }),
+        retryStrategy: new ConfiguredRetryStrategy(
+            AWS_SDK_REQUEST_RETRIES, // maxAttempts
+            attempt => {
+                log.error('aws sdk request error', { retryCount: attempt });
+                // The delay is not truly exponential; it resets to the minimum after every 10 calls,
+                // with a maximum delay of 15 seconds.
+                return AWS_SDK_REQUEST_DELAY_MS * (2 ** (attempt % 10));
+            }
+        )
+    };
+
+    const awsv2Config = {
         accessKeyId: accessKey,
         secretAccessKey: secretKey,
         endpoint,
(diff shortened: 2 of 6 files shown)


CRR/clients.js
```diff
@@ -43,32 +67,10 @@ function setupClients({
...
-        return AWS_SDK_REQUEST_DELAY_MS * (2 ** (retryCount % 10));
-    }
-
-    // Specific options for S3 requests
-    const s3SpecificOptions = {
-        maxRetries: AWS_SDK_REQUEST_RETRIES,
-        customBackoff: customBackoffStrategy,
+    return {
+        s3: new S3Client(awsv3Config),
+        bb: new BackbeatClient(awsv2Config),
```
SylvainSenechal, line 80: Let's do the BB client in a separate PR, this pr is already big enough and BackbeatClient is working differently

CRR/ReplicationStatusUpdater.js
```diff
@@ -266,13 +270,15 @@ class ReplicationStatusUpdater {
...
+        this.s3.send(new ListObjectVersionsCommand({
             Bucket: bucket,
             MaxKeys: this.listingLimit,
             Prefix: this.targetPrefix,
             VersionIdMarker,
             KeyMarker,
-        }, cb);
+        }))
+            .then(data => cb(null, data))
+            .catch(err => cb(err));
```
ghost: In general we try to avoid this pattern. The pattern is when we call callbacks but use promises with .then and .catch (in that order)

For instance today in S3, if an S3 API throws after calling the mongoclientinterface, because that call was a `.then(cb).catch(log then cb)`, any error in the callback would throw and be catched by the .catch. This is a bit misleading because we would think the .catch would work only for the first promise, but in fact it's not: we chain the results so the .then returns a promise and the .catch will also catch it...

It's a problem because then it (1) usually hides the right error with something totally unrelated and (2) we are actually calling the callback twice! That translates into wrong states in s3. 

Here, I believe its the same, if you call the `db` but then for whatever reason we throw later, we end up in the .catch, and call `cb` again. If that call comes from an `async.waterfall` or similar, we would then rethrow because of a `callback already called` error.

As we have callbacks I however cannot see an elegant solution (the solution would be to use properly async/await or return a promise directly). So I'm just putting that here... There is also a solution to have try/catch in the then and catch, but that is not good either...

Is it a big effort to switch this project from callback based to async/await (probably not in the same PR)?

.github/workflows/test.yml
```diff
@@ -69,9 +69,9 @@ jobs:
           wget -nc https://ftp.debian.org/debian/pool/main/o/openssl/libssl1.1_1.1.1w-0+deb11u1_amd64.deb
           sudo dpkg -i libssl1.1_1.1.1w-0+deb11u1_amd64.deb
       - run: yarn --network-concurrency 1
-      - run: yarn --silent lint -- --max-warnings 0
-      - run: yarn --silent test:unit
-      - run: yarn --silent test:functional
+      # - run: yarn --silent lint -- --max-warnings 0
```
benzekrimaha: Were these commented on purpose or just a left out from debug ?

CRR/ReplicationStatusUpdater.js
```diff
@@ -266,13 +270,20 @@ class ReplicationStatusUpdater {
      * @returns {void}
      */
     _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
-        return this.s3.listObjectVersions({
+        this.log.error(`AAAAA 11 ${bucket}, ${VersionIdMarker}, ${KeyMarker}`);
```
benzekrimaha: Left out from debug , need to be removed before merging

.github/workflows/test.yml
SylvainSenechal: yeah hold on i was testing something

CRR/ReplicationStatusUpdater.js
SylvainSenechal: Yeah I did it for one of the scripts, for this one its more complicated, we are using a multiple functions from the async lib, with workers, so I'll leave it this way

tests/utils/crr.js
```diff
@@ -159,25 +159,40 @@ const putMetadataRes = { versionId: '98295539708053999999RG001  ' };
...
+
+    // Override default responses with custom responses if provided
+    const responses = { ...defaultResponses, ...customS3Responses };
+    const sendMock = jest.fn(command => {
+        const commandName = command.constructor.name;
+        if (responses[commandName]) {
+            return Promise.resolve(responses[commandName]);
+        }
+        return Promise.reject(new Error(`Unknown command: ${commandName}`));
+    });
```
DarkIsDude:
```suggestion
    const defaultResponses = {
        ListObjectVersionsCommand: listVersionRes,
        GetBucketReplicationCommand: getBucketReplicationRes,
    };

    const responses = { ...defaultResponses, ...customS3Responses };
    const sendMock = jest.fn(command => {
        const commandName = command.constructor.name;
        if (responses[commandName]) {
            return Promise.resolve(responses[commandName]);
        }
        return Promise.reject(new Error(`Unknown command: ${commandName}`));
    });
```

If you want to keep your comment, you update your var name `defaultResponsesOverwrittenWithCustom` but for me it's not needed

CRR/ReplicationStatusUpdater.js
```diff
@@ -26,6 +29,7 @@ class ReplicationStatusUpdater {
      * @param {number} [params.maxScanned] - (Optional) Maximum number of items to scan.
      * @param {string} [params.keyMarker] - (Optional) Key marker for resuming object listing.
      * @param {string} [params.versionIdMarker] - (Optional) Version ID marker for resuming object listing.
+     * @param {boolean} [params.currentVersionOnly] - (Optional) Whether to process only the current version of objects.
```
DarkIsDude, line 32: Where do we use it ?

CRR/ReplicationStatusUpdater.js
```diff
@@ -266,13 +270,17 @@ class ReplicationStatusUpdater {
...
             MaxKeys: this.listingLimit,
             Prefix: this.targetPrefix,
             VersionIdMarker,
             KeyMarker,
-        }, cb);
+        }))
+            .then(data => {
+                cb(null, data);
+            })
+            .catch(err => cb(err));
```
DarkIsDude:
```suggestion
            .catch(cb);
```

CRR/ReplicationStatusUpdater.js
```diff
@@ -284,15 +292,17 @@ class ReplicationStatusUpdater {
...
+            next => {
+                this.s3.send(new GetBucketReplicationCommand(
+                    { Bucket: bucket },
+                ))
+                    .then(res => next(null, res.ReplicationConfiguration))
+                    .catch(err => {
+                        this.log.error('error getting bucket replication', { error: err });
+                        return next(err);
+                    });
+            },
```
DarkIsDude:
```suggestion
           async() => {
                const res = await this.s3.send(new GetBucketReplicationCommand(
                    { Bucket: bucket },
                ));
                
                return res.ReplicationConfiguration
            },
```

You can keep the `this.log.error('error getting bucket replication', { error: err });` if you want with a try catch

CRR/clients.js
```diff
@@ -27,7 +29,34 @@ function setupClients({
...
+            accessKeyId: accessKey,
+            secretAccessKey: secretKey,
+        },
+        endpoint,
+        forcePathStyle: true,
+        requestHandler: new NodeHttpHandler({
+            httpAgent: new http.Agent({ keepAlive: true }),
+            httpsAgent: new https.Agent({ 
+                keepAlive: true,
+                rejectUnauthorized: false
```
DarkIsDude, line 47: Why this one ?

tests/unit/CRR/ReplicationStatusUpdater.js
```diff
@@ -29,19 +29,24 @@ describe('ReplicationStatusUpdater', () => {
...
-                MaxKeys: 10,
-                Prefix: 'toto',
-                VersionIdMarker: null,
-            }, expect.any(Function));
-
-            expect(crr.s3.getBucketReplication).toHaveBeenCalledTimes(1);
-            expect(crr.s3.getBucketReplication).toHaveBeenCalledWith({
-                Bucket: 'bucket0',
-            }, expect.any(Function));
+            expect(crr.s3.send).toHaveBeenCalledTimes(2); // One for list, one for getBucketReplication
```
DarkIsDude:
```suggestion
            expect(crr.s3.send).toHaveBeenCalledTimes(2);
```

It's the next check

tests/unit/CRR/ReplicationStatusUpdater.js
```diff
@@ -83,24 +88,38 @@ describe('ReplicationStatusUpdater', () => {
...
-        crr.s3.listObjectVersions = jest.fn((params, cb) => cb(null, listVersionsRes));
+        crr = initializeCrrWithMocks({
+            buckets: ['bucket0'],
+            workers: 10,
+            replicationStatusToProcess: ['NEW'],
+            targetPrefix: 'toto',
+            listingLimit: 10,
+            siteName: 'aws-location',
+        }, logger, {
+            ListObjectVersionsCommand: listVersionsRes, // Use custom response for list object versions
```
DarkIsDude:
```suggestion
            ListObjectVersionsCommand: listVersionsRes,
```

That's the standard of the function and the signature of the function

DarkIsDude (approved):
LGTM.  Minor fixes. I can re-review if you want 🙏 and you can merge directly

CRR/ReplicationStatusUpdater.js
SylvainSenechal: Oh it's just some documentation for a parameter that Walid added a few weeks ago with another pr

tests/utils/crr.js
SylvainSenechal: yeah no need I removed it

CRR/ReplicationStatusUpdater.js
SylvainSenechal: good catch I'm gonna use this pattern instead

CRR/clients.js
SylvainSenechal: I tested the script against my artesca lab locally, which uses https, so I needed this to make it work. 
In prod, users use it from inside their artesca with http so they don't need it, but I think it's reasonable to keep this here

tests/unit/CRR/ReplicationStatusUpdater.js
SylvainSenechal: I mean the test here will call 'send' twice, one for listObjectVersion, another one for getBuckerReplicationConfig

CRR/ReplicationStatusUpdater.js
```diff
@@ -266,13 +270,17 @@ class ReplicationStatusUpdater {
...
             MaxKeys: this.listingLimit,
             Prefix: this.targetPrefix,
             VersionIdMarker,
             KeyMarker,
-        }, cb);
+        }))
+            .then(data => {
+                cb(null, data);
+            })
+            .catch(cb);
```
ghost:
```suggestion
            .then(data => cb(null, data))
            .catch(cb);
```

CRR/clients.js
```diff
@@ -27,7 +29,34 @@ function setupClients({
...
+        region: 'us-east-1',
+        credentials: {
+            accessKeyId: accessKey,
+            secretAccessKey: secretKey,
+        },
+        endpoint,
+        forcePathStyle: true,
+        requestHandler: new NodeHttpHandler({
+            httpAgent: new http.Agent({ keepAlive: true }),
+            httpsAgent: new https.Agent({
```
ghost, line 42: FYI there is this project that we try to use when we use custom http agents, to auto enforce consistent keepalive & other rules: https://github.com/scality/httpagent. That might be helpful here.

CRR/clients.js
```diff
@@ -27,7 +29,34 @@ function setupClients({
...
+        },
+        endpoint,
+        forcePathStyle: true,
+        requestHandler: new NodeHttpHandler({
+            httpAgent: new http.Agent({ keepAlive: true }),
+            httpsAgent: new https.Agent({ 
+                keepAlive: true,
+                rejectUnauthorized: false
+            }),
+            requestTimeout: 0,
```
ghost: I do not think we should disable the timeout of the request. Usually we have 60s of keepalive anyway. Should we enforce a maximum so we don't remain stuck due to any even causing the request to never be fulfilled?

CRR/clients.js
ghost: Could we add a comment to say, in v2 ssl was anyway disabled, and so here we start accepting https but there will be no security checks enforced? Indeed today, we do not have TLS between internal services, so it cannot work. But that might change with the arrival of talos/cilium (if confirmed), where we may start enforcing internal security.

CRR/clients.js
SylvainSenechal: I kept it the way it was setup in V2 at first, but yeah let's use 60s its more reasonable

ghost (approved):

PR #347, opened 2025-08-26
Migration aws sdk v2 to v3 : verifyReplication

SylvainSenechal:
Issue : S3UTILS-201

### DIFF
--- a/VerifyReplication/storage/s3.js
+++ b/VerifyReplication/storage/s3.js
@@ -1,4 +1,6 @@
-const AWS = require('aws-sdk');
+const { S3Client, HeadObjectCommand, ListObjectsV2Command } = require('@aws-sdk/client-s3');
+const { NodeHttpHandler } = require('@aws-sdk/node-http-handler');
+const { ConfiguredRetryStrategy } = require('@smithy/util-retry');
 const fs = require('fs');
 const http = require('http');
 const https = require('https');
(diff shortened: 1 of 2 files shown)


VerifyReplication/storage/s3.js
```diff
@@ -98,31 +85,29 @@ function getObjMd(params, cb) {
...
-        VersionId: versionId,
-    }, (err, data) => {
-        if (err) {
-            return cb(err);
+    try {
+        const commandParams = {
+            Bucket: bucket,
+            Key: key,
+        };
+        if (versionId) {
```
francoisferrand: do we need this `if` ?
is it not the same as `VersionId: versionId || undefined`, or even just `VersionId: versionId` as before?

VerifyReplication/storage/s3.js
```diff
@@ -17,75 +19,60 @@ function getClient(params) {
...
-                        retryCount,
-                        backoff,
-                    });
-                }
-                return backoff;
-            },
-        },
-    };
+    if (showClientLogsIfAvailable) {
+        clientConfig.logger = console;
```
francoisferrand, line 70: keep the `// TODO: may be use werelogs` or switch to werelogs

francoisferrand (approved):

VerifyReplication/storage/s3.js
SylvainSenechal: Yeah I was overly defensive but I rechecked, we can directly add the versionId in the params, and if its null/undefined/empty, the sdk will act like its not setup

ghost (approved):

PR #349, opened 2025-09-01
Migration aws sdk v2 to v3 : listFailedObjects

SylvainSenechal:
Issue : S3UTILS-202

### DIFF
--- a/VerifyReplication/verifyReplication.js
+++ b/VerifyReplication/verifyReplication.js
@@ -43,15 +43,16 @@ function verifyObjects(objectList, cb) {
         return destinationStorage.getObjMd(params, (err, dstMd) => {
             ++statusObj.dstProcessedCount;
             if (err) {
-                ++statusObj.dstFailedMdRetrievalsCount;
-                logger.error('error getting metadata', {
-                    error: err,
-                    bucket: statusObj.dstBucket,
-                    key: dstKey,
-                    srcLastModified,
-                });
                 if (err instanceof NotFound) {
                     ++statusObj.missingInDstCount;
+                } else {
+                    ++statusObj.dstFailedMdRetrievalsCount;
+                    logger.error('error getting metadata', {
+                        error: err,
+                        bucket: statusObj.dstBucket,
+                        key: dstKey,
+                        srcLastModified,
+                    });
                 }
                 // log the error and continue processing objects
                 return done();
--- a/listFailedObjects.js
+++ b/listFailedObjects.js
@@ -1,6 +1,7 @@
 const async = require('async');
-const AWS = require('aws-sdk');
-const { http } = require('httpagent');
+const { S3Client, ListObjectVersionsCommand, HeadObjectCommand } = require('@aws-sdk/client-s3');
+const { NodeHttpHandler } = require('@aws-sdk/node-http-handler');
+const http = require('http');
 
 const { Logger } = require('werelogs');
 
@@ -32,31 +33,29 @@ if (!SECRET_KEY) {
     process.exit(1);
 }
 
-const s3 = new AWS.S3({
-    accessKeyId: ACCESS_KEY,
-    secretAccessKey: SECRET_KEY,
+const s3 = new S3Client({
     region: 'us-east-1',
-    sslEnabled: false,
-    endpoint: ENDPOINT,
-    s3ForcePathStyle: true,
-    apiVersions: { s3: '2006-03-01' },
-    signatureVersion: 'v4',
-    signatureCache: false,
-    httpOptions: {
-        maxRetries: 0,
-        timeout: 0,
-        agent: new http.Agent({ keepAlive: true }),
+    credentials: {
+        accessKeyId: ACCESS_KEY,
+        secretAccessKey: SECRET_KEY,
     },
+    endpoint: ENDPOINT,
+    forcePathStyle: true,
+    tls: false,
+    requestHandler: new NodeHttpHandler({
+        httpAgent: new http.Agent({ keepAlive: true }),
+        requestTimeout: 60000,
+    }),
 });
 
 // list object versions
 function _listObjectVersions(bucket, VersionIdMarker, KeyMarker, cb) {
-    s3.listObjectVersions({
+    s3.send(new ListObjectVersionsCommand({
         Bucket: bucket,
         MaxKeys: LISTING_LIMIT,
         VersionIdMarker,
         KeyMarker,
-    }, cb);
+    })).then(data => cb(null, data)).catch(cb);
 }
 
 // return object with key and version_id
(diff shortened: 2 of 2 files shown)


CRR/clients.js
```diff
@@ -1,10 +1,12 @@
-const AWS = require('aws-sdk');
+const { S3Client } = require('@aws-sdk/client-s3');
+const { NodeHttpHandler } = require('@aws-sdk/node-http-handler');
+const { ConfiguredRetryStrategy } = require('@smithy/util-retry');
 const http = require('http');
+const https = require('https');
 const BackbeatClient = require('../BackbeatClient');
```
francoisferrand, line 6: 😭
We have backbeat client here, so we will need to handle it just like we will do in backbeat....
- Please create a ticket to fix this eventually (no point to do it twice, so we will do it once we have done the upgrade in backbeat)
- Need to consider how we will avoid duplicating `BackbeatClient` code between s3utils and backbeat. Maybe ok if we just generate it, or may need to export it from Cloudserver / a "sub package" of cloudserver / a dedicated package / ...

cc @benzekrimaha @SylvainSenechal

CRR/clients.js
```diff
@@ -27,7 +29,34 @@ function setupClients({
     secretKey,
     endpoint,
 }, log) {
-    const awsConfig = {
+    const awsv3Config = {
```
francoisferrand, line 32: seems like we have a lot of these aws client setup (in every script).
should we factorize this in the common/utils folder?

CRR/ReplicationStatusUpdater.js
```diff
@@ -266,13 +270,15 @@ class ReplicationStatusUpdater {
...
+        this.s3.send(new ListObjectVersionsCommand({
             Bucket: bucket,
             MaxKeys: this.listingLimit,
             Prefix: this.targetPrefix,
             VersionIdMarker,
             KeyMarker,
-        }, cb);
+        }))
+            .then(data => cb(null, data))
+            .catch(err => cb(err));
```
francoisferrand: nit:
```suggestion
        })).then(data => cb(null, data), cb);
```
or 
```suggestion
        })).then(data => cb(null, data)).catch(cb);
```

VerifyReplication/verifyReplication.js
```diff
@@ -41,26 +42,20 @@ function verifyObjects(objectList, cb) {
 
         return destinationStorage.getObjMd(params, (err, dstMd) => {
             ++statusObj.dstProcessedCount;
-            if (err && err.code !== 'NotFound') {
+            if (err) {
                 ++statusObj.dstFailedMdRetrievalsCount;
```
francoisferrand: should not be increased if `notfound`:

```suggestion
if (err instanceof NotFound) {
    ++statusObj.missingInDstCount;
} else {
    ++statusObj.dstFailedMdRetrievalsCount;
}
```

listFailedObjects.js
```diff
@@ -32,31 +33,35 @@ if (!SECRET_KEY) {
...
         VersionIdMarker,
         KeyMarker,
-    }, cb);
+    }))
+        .then(data => {
+            cb(null, data);
+        })
+        .catch(err => {
+            cb(err);
+        });
```
francoisferrand:
```suggestion
    })).then(data => cb(null, data), cb);
```

listFailedObjects.js
```diff
@@ -82,21 +87,21 @@ function listBucket(bucket, cb) {
...
+                    }))
+                        .then(res => {
                             if (res.ReplicationStatus === 'FAILED') {
                                 console.log({ Key, ...res });
                             }
                             return next();
-                        },
-                    );
+                        })
+                        .catch(err => next(err));
```
francoisferrand:
```suggestion
                        .catch(next);
```

listFailedObjects.js
```diff
@@ -82,21 +87,21 @@ function listBucket(bucket, cb) {
...
-                        (err, res) => {
-                            if (err) {
-                                return next(err);
-                            }
+                    s3.send(new HeadObjectCommand({
+                        Bucket: bucketName,
+                        Key,
+                        VersionId,
+                    }))
+                        .then(res => {
```
francoisferrand:
```suggestion
                    })).then(res => {
                       ...
                    }, next);
```

CRR/clients.js
francoisferrand: (should not be fixed in this PR)

SylvainSenechal: Yes, its on this ticket https://scality.atlassian.net/jira/software/c/projects/OS/boards/268?selectedIssue=S3UTILS-205

CRR/clients.js
SylvainSenechal: Its possible to refactor around it, although some of the clients have some configuration differences, I prefer to focus on the migration now, and potentially check at the end

listFailedObjects.js
```diff
@@ -32,31 +33,31 @@ if (!SECRET_KEY) {
...
-    s3.listObjectVersions({
+    s3.send(new ListObjectVersionsCommand({
         Bucket: bucket,
         MaxKeys: LISTING_LIMIT,
         VersionIdMarker,
         KeyMarker,
-    }, cb);
+    }))
+        .then(data => cb(null, data))
+        .catch(err => cb(err));
```
francoisferrand:
```suggestion
        .catch(cb);
```

listFailedObjects.js
```diff
@@ -82,21 +83,20 @@ function listBucket(bucket, cb) {
...
+                        Bucket: bucketName,
+                        Key,
+                        VersionId,
+                    })).then(res => {
+                        if (res.ReplicationStatus === 'FAILED') {
+                            console.log({ Key, ...res });
+                        }
+                        return next();
+                    })
+                        .catch(next);
```
francoisferrand: stay consistent with `.then` :
```suggestion
                    }).catch(next);
```

listFailedObjects.js
francoisferrand: nit: maybe just me, but I think the "gap" very confusing to read, breaking the flow of the chained call...
(which there was a strict linter, to enforce such "style" decision)

```suggestion
    })).then(data => cb(null, data)).catch(cb);
```

francoisferrand (approved):

listFailedObjects.js
```diff
@@ -82,21 +83,20 @@ function listBucket(bucket, cb) {
...
-                            return next();
-                        },
-                    );
+                    s3.send(new HeadObjectCommand({
+                        Bucket: bucketName,
+                        Key,
+                        VersionId,
+                    })).then(res => {
+                        if (res.ReplicationStatus === 'FAILED') {
+                            console.log({ Key, ...res });
```
ghost: Could we seize this opportunity to use werelogs and not console log here?

ghost (approved):

listFailedObjects.js
SylvainSenechal: yes let's go

PR #350, opened 2025-09-01
Migration aws sdk v2 to v3 : removeDeleteMarkers

SylvainSenechal:
Issue : S3UTILS-204

### DIFF
--- a/VerifyReplication/storage/s3.js
+++ b/VerifyReplication/storage/s3.js
@@ -73,7 +73,7 @@ function getClient(params) {
     return new S3Client(clientConfig);
 }
 
-async function getObjMd(params, cb) {
+function getObjMd(params, cb) {
     const {
         client,
         bucket,
@@ -86,27 +86,26 @@ async function getObjMd(params, cb) {
         return cb(new Error(errMsg));
     }
 
-    try {
-        const commandParams = {
-            Bucket: bucket,
-            Key: key,
-            VersionId: versionId
-        };
-        
-        const data = await client.send(new HeadObjectCommand(commandParams));
-        const resp = {
-            size: data.ContentLength,
-            lastModified: data.LastModified,
-            versionId: data.VersionId,
-            md: data.Metadata,
-        };
-        return cb(null, resp);
-    } catch (err) {
-        return cb(err);
-    }
+    const commandParams = {
+        Bucket: bucket,
+        Key: key,
+        VersionId: versionId
+    };
+    
+    return client.send(new HeadObjectCommand(commandParams))
+        .then(data => {
+            const resp = {
+                size: data.ContentLength,
+                lastModified: data.LastModified,
+                versionId: data.VersionId,
+                md: data.Metadata,
+            };
+            return cb(null, resp);
+        })
+        .catch(cb);
 }
 
-async function listObjects(params, cb) {
+function listObjects(params, cb) {
     const {
         client,
         bucket,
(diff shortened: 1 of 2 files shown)


VerifyReplication/storage/s3.js
```diff
@@ -98,31 +87,29 @@ function getObjMd(params, cb) {
...
-            versionId: VersionId,
-            md: Metadata,
+            size: data.ContentLength,
+            lastModified: data.LastModified,
+            versionId: data.VersionId,
+            md: data.Metadata,
         };
         return cb(null, resp);
-    });
+    } catch (err) {
```
francoisferrand: catch must not wrap the callback: otherwise, if the callback throws, we will call it again....

----

best way is to stick to a single approach:
* if the function being implemented (i.e. `getObjMd`) is a callback API, probably best to use the promise directly : `client.send().then(...)`
* if the function is async/promise-based (or updated this), then we can and should indeed use await

and/or we can make a trampoline to ease transition (in the same or different functions):

```
async function getObjMd(params, cb) {
   if (cb) {
       return getObjMd(params).then(result -> cb(null, result), cb);
   }

   // no need for catch actually, since that is the way to return the error!
   return await client.send(...);
} 
```

francoisferrand: note: we need to make this PR *correct*, but to avoid too much changes we can also choose to be less ambitious (not moving to "proper" async yet, just use callbacks and promises), and upgrade to async later (in separate PR)

listFailedObjects.js
```diff
@@ -32,31 +33,35 @@ if (!SECRET_KEY) {
...
         VersionIdMarker,
         KeyMarker,
-    }, cb);
+    }))
+        .then(data => {
+            cb(null, data);
+        })
+        .catch(err => {
+            cb(err);
+        });
```
francoisferrand:
```suggestion
        .catch(cb);
```

listFailedObjects.js
```diff
@@ -82,21 +87,21 @@ function listBucket(bucket, cb) {
...
+                    }))
+                        .then(res => {
                             if (res.ReplicationStatus === 'FAILED') {
                                 console.log({ Key, ...res });
                             }
                             return next();
-                        },
-                    );
+                        })
+                        .catch(err => next(err));
```
francoisferrand:
```suggestion
                        .catch(next);
```

VerifyReplication/storage/s3.js
```diff
@@ -98,31 +87,29 @@ function getObjMd(params, cb) {
...
+        return cb(err);
+    }
 }
 
-function listObjects(params, cb) {
+async function listObjects(params, cb) {
```
francoisferrand: same here: catch must not wrap the callback: otherwise, if the callback throws, we will call it again....

VerifyReplication/storage/s3.js
```diff
@@ -140,14 +127,19 @@ function listObjects(params, cb) {
...
+    try {
+        // TODO: support listing all versions
+        const data = await client.send(new ListObjectsV2Command({
+            Bucket: bucket,
+            MaxKeys: listingLimit,
+            Prefix: prefix,
+            Delimiter: delimiter,
+            ContinuationToken: nextContinuationToken,
+        }));
+        return cb(null, data);
```
francoisferrand: same here: catch must not wrap the callback: otherwise, if the callback throws, we will call it again....

VerifyReplication/storage/s3.js
SylvainSenechal: I see, I reverted it

VerifyReplication/storage/s3.js
SylvainSenechal, line 1: Not fully related to this pr whose goal is to migrate removeDeleteMarkers, but I made a mistake in one of the previous ones, mixing callback and catch, so I'm fixing it here

benzekrimaha (approved):
lgtm

francoisferrand (approved):

PR #348, opened 2025-09-01
Migration aws sdk v2 to v3 : cleanupBuckets

SylvainSenechal:
ISSUE : S3UTILS-203

PR #351, opened 2025-09-23
Started migration for buckerVersionStats

SylvainSenechal:
Issue : S3UTILS-200

### DIFF
--- a/bucketVersionsStats.js
+++ b/bucketVersionsStats.js
@@ -1,13 +1,13 @@
 const fs = require('fs');
 const { http, https } = require('httpagent');
 
-const AWS = require('aws-sdk');
-const { doWhilst } = require('async');
+const { S3Client, ListObjectVersionsCommand } = require('@aws-sdk/client-s3');
+const { NodeHttpHandler } = require('@aws-sdk/node-http-handler');
+const { ConfiguredRetryStrategy } = require('@smithy/util-retry');
 
 const { Logger } = require('werelogs');
 
 const parseOlderThan = require('./utils/parseOlderThan');
-const { safeListObjectVersions } = require('./utils/safeList');
 
 const log = new Logger('s3utils::bucketVersionsStats');
 const { ENDPOINT } = process.env;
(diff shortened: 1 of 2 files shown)


bucketVersionsStats.js
```diff
@@ -147,10 +142,17 @@ let VersionIdMarker;
 function _logProgress(message) {
     const loggedStats = {
         total: {
-            count: BigInt(stats.current.count + stats.noncurrent.count),
-            size: BigInt(stats.current.size + stats.noncurrent.size),
+            count: String(stats.current.count + stats.noncurrent.count),
```
SylvainSenechal: there was a weird error with the logger when using bigInt

cleanupNoncurrentVersions.js
```diff
@@ -297,13 +329,37 @@ function _doBatchDelete(bucket) {
...
+    const command = new DeleteObjectsCommand({
         Bucket: bucket,
         Delete: { Objects: batchDeleteObjects },
-    };
-    s3.deleteObjects(params, err => {
-        if (err) {
-            log.error('batch delete error', { error: err });
+    });
+
+    command.middlewareStack.add(
```
SylvainSenechal: We get a badDigest error without this

bucketVersionsStats.js
francoisferrand: nit: not sure if `String()` function shoud be used (as you did), or if the `toString()` method should be preferred. I did not find any mention of BigInt support with the `String()` function, though running in a node.js shell it seems to work just as well... 🤷‍♂️

cleanupNoncurrentVersions.js
francoisferrand: same as https://github.com/scality/s3utils/pull/348#discussion_r2387203260

francoisferrand (approved):

bucketVersionsStats.js
```diff
@@ -97,38 +97,32 @@ if (s3EndpointIsHttps) {
...
+    forcePathStyle: true,
+    tls: s3EndpointIsHttps,
+    requestHandler: new NodeHttpHandler({
+        httpAgent: agent,
+        httpsAgent: agent,
+        requestTimeout: 60000,
+    }),
+    retryStrategy: new ConfiguredRetryStrategy(
+        AWS_SDK_REQUEST_RETRIES,
+        // eslint-disable-next-line arrow-body-style
```
DarkIsDude: Disabling a linter rule should not be used. If you need to use it, I would suggest to add a comment to explain the why `// eslint-disable-next-line arrow-body-style YOUR COMMENT HERE`

cleanupNoncurrentVersions.js
```diff
@@ -173,6 +174,33 @@ if (s3EndpointIsHttps) {
...
+    forcePathStyle: true,
+    tls: s3EndpointIsHttps,
+    requestHandler: new NodeHttpHandler({
+        httpAgent: agent,
+        httpsAgent: agent,
+        requestTimeout: 60000,
+    }),
+    retryStrategy: new ConfiguredRetryStrategy(
+        AWS_SDK_REQUEST_RETRIES,
+        // eslint-disable-next-line arrow-body-style
```
DarkIsDude: Same here 🙏. Also this is duplicated code ? Can we factorise it somethere ?

DarkIsDude (approved):

bucketVersionsStats.js
SylvainSenechal: yeah i fixed it, removed the lint disable and changed the syntax

cleanupNoncurrentVersions.js
SylvainSenechal: We could factorise it, but for now all the scripts have their own independent client, I don't want to change this here for now

PR #353, opened 2025-10-24
S3UTILS-208 Reset object replication configuration

nicolas2bert:
Adds a new `FORCE_USING_CONFIGURATION` environment variable to the CRR existing objects processing tool. 
When enabled, this option allows custumers to force-update existing object replication configurations to match the current bucket replication configuration.

**Context**

When an object is created, its replication destination bucket and the IAM role used for replication are hardcoded in its object metadata under the `replicationInfo` field.
This design makes sure that each object is replicated according to the exact bucket replication configuration in effect at the time of its creation. 

Later changes of the bucket replication configuration only affect newly created objects.

However, in the case of CRR existing objects processing (where objects are re-queued for replication or re-replication) the current bucket replication configuration might have changed since the objects were originally created. Customers may want these reprocessed objects to follow the updated bucket replication configuration rather than the old one stored in their metadata.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "@senx/warp10": "^2.0.3",
     "@smithy/util-retry": "^4.0.7",
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/arsenal#8.2.26",
+    "arsenal": "git+https://github.com/scality/arsenal#8.2.36",
     "async": "^3.2.6",
     "aws-sdk": "^2.1691.0",
     "bucketclient": "scality/bucketclient#8.2.6",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3utils",
-  "version": "1.16.1",
+  "version": "1.16.2",
   "engines": {
     "node": ">= 22"
   },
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
     "@senx/warp10": "^2.0.3",
     "@smithy/util-retry": "^4.0.7",
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/arsenal#8.2.26",
+    "arsenal": "git+https://github.com/scality/arsenal#improvement/ARSN-536/setters",
```
dvasilas: Reminder to update before merge.

dvasilas: Also reminder to bump the s3utils package version

dvasilas (approved):

fredmnl (approved):
LGTM

Nitpick: `FORCE_USING_CONFIGURATION` is a bit confusing IMO but I'm not sure I have a much better option to provide. Here are some ideas `OVERWRITE_WITH_CURRENT_REPLICATION_CONFIGURATION`, `OVERRIDE_OLD_CONFIGURATION`, `OVERWRITE_OLD_CONFIGURATION`, `REWRITE_REPLICATION_INFO_FIELD`, `REEVALUATE_REPLICATION_FOR_OLD_OBJECTS`? Feel free to change or keep it as is.

nicolas2bert:
@bert-e approve

nicolas2bert:
ping

PR #354, opened 2025-11-06
S3UTILS-209 Prevent metadata ObjectMD field loss in CRRExistingObject

nicolas2bert:
This change adds a safety check to make sure we never rewrite object metadata using an older ObjectMD version than the one already stored. 

If the new version is lower than the original one, the update is stopped.

This is a preventive safeguard, we have not seen this issue happening on customer environments so far.


BourgoisMickael (approved):

dvasilas (approved):

nicolas2bert:
ping

nicolas2bert:
@bert-e approve

PR #355, opened 2025-11-12
Replace Backbeat client with Cloudserver client

SylvainSenechal:
ISSUE: S3UTILS-205


CRR/ReplicationStatusUpdater.js
```diff
@@ -86,19 +87,17 @@ class ReplicationStatusUpdater {
...
-        const { s3, bb } = setupClients({
+        this.s3 = createS3Client({
             accessKey: this.accessKey,
             secretKey: this.secretKey,
             endpoint: this.endpoint,
         }, this.log);
-
-        this.s3 = s3;
-        this.bb = bb;
+        this.cloudserverclient = new CloudserverClient(this.endpoint, this.accessKey, this.secretKey);
```
SylvainSenechal, line 100: Should I use an object instead of 3 parameters 🤔

package.json
```diff
@@ -47,7 +46,8 @@
     "vaultclient": "scality/vaultclient#8.5.1",
     "werelogs": "scality/werelogs#8.2.1",
     "xml2js": "^0.6.2",
-    "zenkoclient": "scality/zenkoclient#1.2.1"
+    "zenkoclient": "scality/zenkoclient#1.2.1",
+    "@scality/cloudserverclient": "git+https://github.com/scality/cloudserverclient#improvement/CLDSRVCLT-1"
```
SylvainSenechal: TODO: update after cloudserver client released

Clients/CloudserverClient.js
```diff
@@ -0,0 +1,68 @@
...
+
+        this.client = new CloudserverClient(config);
+    }
+
+    getMetadata(params, callback) {
+        const command = new GetMetadataCommand(params);
+        this.client.send(command)
+            .then(data => callback(null, data))
+            .catch(err => callback(err));
+    }
```
francoisferrand, line 58: since the rest of the code was migrated to fully "adopt" the new AWS sdk API (i.e. new Command().send() and promises), should we not do the same here, esp. since it seems there are not many calls?

(and here we would only have a `createCloudserverClient` function, similar to what we have s3Client.js - or possibly keep the existing "setupCllients" code, which would setup both clients mostly in the same way, and avoid some duplication [same http clients...])

francoisferrand (approved):

CRR/ReplicationStatusUpdater.js
```diff
@@ -161,7 +160,7 @@ class ReplicationStatusUpdater {
         let skip = false;
         return waterfall([
             // get object blob
-            next => this.bb.getMetadata({
+            next => this.cloudserverclient.getMetadata({
```
francoisferrand, line 163: not an issue to rename this variable, but it would also be acceptable to keep the old name IMHO (as this client is still used to access the "backbeat routes" of cloudserver...)

tests/utils/crr.js
```diff
@@ -171,6 +171,10 @@ const putMetadataRes = { versionId: '98295539708053999999RG001  ' };
  * @returns {ReplicationStatusUpdater} An instance of ReplicationStatusUpdater with mocked methods.
  */
 function initializeCrrWithMocks(config, log, customS3Responses = {}) {
+    // eslint-disable-next-line no-param-reassign
+    if (config.endpoint === undefined) {
+        config.endpoint = 'http://dummyEndpoint:8000';
```
francoisferrand, line 176: why is this needed?

SylvainSenechal: The cloudserver client initialization will fail if no endpoint is provided. Since we are in a test and the client is mocked, initializeCrrWithMocks is often called with config containing almost nothing (and missing the endpoint)

Clients/CloudserverClient.js
SylvainSenechal: Yeah I need to review the code see which one is best, the tradeoff here is with the current solution we don't even need to change the way the client is calling the methods, but at the cost of some helper function and old syntax. Else I remove it but I need to change the way the methods are called in the code, and that code is not async/await ready so it won't be 100% clean either

DarkIsDude (approved):

PR #359, opened 2025-12-19
S3UTILS-211: list replication status is now configurable

tcarmet:
Easier to review commit per commit, most of the diff is existing code because I renamed a file. 

We have two scripts now that achieve similar things:
- `listFailedObject.js` with its original behavior of listing failed replication objects.
- `listObjectsByReplicationStatus.js`, with the capability of listing objects with a given `REPLICATION_STATUS`.

Functional tests have been added to ensure all changes are now covered by tests.


listObjectsByReplicationStatus.js
```diff
@@ -31,6 +32,21 @@ if (!SECRET_KEY) {
     log.error('SECRET_KEY not defined');
     process.exit(1);
 }
+if (!REPLICATION_STATUS) {
+    REPLICATION_STATUS = 'FAILED';
+}
```
jonathan-gramain: I may be nitpicking and definitely not blocking for me, but instead of a default value, we could force the user to provide an explicit `REPLICATION_STATUS` env var when invoking the script as `listObjectsByReplicationStatus`, and set it automatically to `FAILED` when invoked as `listFailedObjects`. I feel like it makes more sense to avoid surprises when invoking as `listObjectsByReplicationStatus`.

So you could probably do this instead (untested):

```

    if (path.basename(process.argv[1]) === 'listFailedObjects.js') {
        // ensure REPLICATION_STATUS env var is not set first, also to avoid surprises and encourage people to use the new script name
        REPLICATION_STATUS = 'FAILED';
    } else if (!REPLICATION_STATUS) {
        // error out
    }
}
```

tests/functional/listObjectsByReplicationStatus.js
```diff
@@ -0,0 +1,312 @@
...
+    adminAccessKeyId,
+    adminSecretAccessKey,
+    createTestAccount,
+    deleteTestAccount,
+} = require('../utils/S3Setup');
+
+const log = new Logger('listObjectsByReplicationStatus:test');
+
+async function configureCrr(accountSource, accountDest) {
+    // activate bucket versionning on source and destination buckets
```
jonathan-gramain:
```suggestion
    // activate bucket versioning on source and destination buckets
```

tests/functional/listObjectsByReplicationStatus.js
```diff
@@ -0,0 +1,312 @@
...
+    log.info('Removing bucket crr configuration', { bucket: account.bucketName });
+    try {
+        await account.s3Client.deleteBucketReplication({ Bucket: account.bucketName }).promise();
+    } catch (err) {
+        log.error('Error removing bucket replication configuration', {
+            bucket: account.bucketName,
+            error: err.message,
+        });
+    }
+    // remove versioning
```
jonathan-gramain: Setting suspended-versioning status is not really removing versioning, it's another versioning state. We can't fully remove the versioning state of a bucket once it has been activated via S3, we need to use a lower-level API like bucketd.

If versioning-suspended is what is wanted, suggesting to just change or remove the comment.

tests/functional/listObjectsByReplicationStatus.js
```diff
@@ -0,0 +1,312 @@
...
+        }
+
+        // Verify objects have replication status
+        log.info('Verifying objects have replication status');
+        for (const obj of testObjects) {
+            const headResult = await accountSource.s3Client.headObject({
+                Bucket: accountSource.bucketName,
+                Key: obj.Key,
+            }).promise();
+            log.info('Object metadata', {
```
jonathan-gramain, line 251: Do you want to add an `expect` clause there to check the replication status (`PENDING`?), in addition to the logging? If not needed, you could change the comment as it implies the code verifies something (but it just logs).

tests/unit/listObjectsByReplicationStatus.js
```diff
@@ -0,0 +1,85 @@
...
+
+    it('should reject when buckets is only whitespace', async () => {
+        const options = { ...validOptions, buckets: '   ' };
+        await expect(listObjectsByReplicationStatus(options))
+            .rejects
+            .toThrow('No buckets given as input! Please provide a comma-separated list of buckets');
+    });
+
+    it('should reject when endpoint is missing', async () => {
+        const options = { ...validOptions, endpoint: null };
```
jonathan-gramain: nitpicking: I'd rather test when the `endpoint` attribute is completely missing (probably a more common error). Same for the rest of the attributes below.

listObjectsByReplicationStatus.js
tcarmet: That's fair, I thought about it as well. I'm thinking instead of using a path check, to remove the symlink of `listFailedObjects.js` and since it's now a callable function, I just call it from there and hardcode the `REPLICATION_STATUS` parameter as `FAILED` there.

We overall achieve the same thing here, but remove the symlink (I kind of don't like it for some reason).

tests/unit/listObjectsByReplicationStatus.js
tcarmet: Applied a `delete options.endpoint;` instead and did the same for the remaining similar tests to test them all the same way.

tests/functional/listObjectsByReplicationStatus.js
tcarmet: Indeed, ended up removing this part, not needed. My goal was being able to delete the bucket and was able to by deleting all versions and delete markers before deleting the bucket.

tests/functional/listObjectsByReplicationStatus.js
tcarmet: indeed forgot to check it. Added a `expect(headResult.ReplicationStatus).toBeDefined();`. Not verifying the status, as at this point I just want to verify it exists.

listObjectsByReplicationStatus.js
tcarmet: done.

jonathan-gramain (approved):

listFailedObjects.js
```diff
@@ -1,130 +1,33 @@
...
+        accessKey: ACCESS_KEY,
+        secretKey: SECRET_KEY,
+        endpoint: ENDPOINT,
+        replicationStatus: 'FAILED',
+        logger: log,
+    }).then(() => {
+        log.info('Completed successfully');
+        process.exit(0);
+    }).catch(err => {
+        log.error('Failed with error', { error: err });
```
BourgoisMickael: can we make sure this prints the error and not an empty object, maybe we need to stringify the error

listFailedObjects.js
```diff
@@ -1,130 +1,33 @@
...
-    (bucket, done) => listBucket(bucket, done),
-    err => {
-        if (err) {
-            log.error('error occured while listing failed objects', {
-                error: err,
-            });
-        }
-    },
-);
+module.exports = { listObjectsByReplicationStatus };
```
BourgoisMickael: you don't need to export in this file ?

listObjectsByReplicationStatus.js
```diff
@@ -0,0 +1,216 @@
...
+            return;
+        }
+        if (!secretKey) {
+            const error = new Error('SECRET_KEY not defined');
+            log.error(error.message);
+            reject(error);
+            return;
+        }
+        if (!replicationStatus) {
+            const error = new Error('REPLICATION_STATUS not defined! Please provide a comma-separated list of replication statuses: NEW,PENDING,COMPLETED,FAILED,REPLICA.');
```
BourgoisMickael: this list is duplicated multiple times. Can you have 1 array of status on top, and then `join(',')` when necessary to avoid duplicating the list of statuses

listObjectsByReplicationStatus.js
```diff
@@ -0,0 +1,216 @@
...
+ * @param {string} options.buckets - Comma-separated list of buckets
+ * @param {string} options.accessKey - AWS access key
+ * @param {string} options.secretKey - AWS secret key
+ * @param {string} options.endpoint - S3 endpoint
+ * @param {string} options.replicationStatus - Comma-separated replication statuses (required)
+ * @param {Object} [options.logger] - Logger instance
+ * @returns {Promise<void>}
+ */
+function listObjectsByReplicationStatus(options) {
+    return new Promise((resolve, reject) => {
```
BourgoisMickael: could you create the promise as late as possible.

Most code could be outside that promise.

And for error handling your could use `return Promise.reject(new Error('...'))`

listObjectsByReplicationStatus.js
```diff
@@ -0,0 +1,216 @@
...
+            replicationStatus,
+            logger,
+        } = options;
+
+        const log = logger || new Logger('s3utils:listObjectsByReplicationStatus');
+
+        // Validate inputs
+        if (!buckets || buckets.trim().length === 0) {
+            const error = new Error('No buckets given as input! Please provide a comma-separated list of buckets');
+            log.error(error.message);
```
BourgoisMickael: Does it duplicate the log as it's also logged in mapSeries callback + the catch at the end ?

tests/utils/S3Setup.js
```diff
@@ -0,0 +1,124 @@
+
+const AWS = require('aws-sdk');
```
BourgoisMickael: could we use sdk v3 as its already used by source code ?

tests/utils/S3Setup.js
```diff
@@ -0,0 +1,124 @@
+
```
BourgoisMickael:
```suggestion
```
missing linter, there are other empty blank lines

tests/unit/listObjectsByReplicationStatus.js
```diff
@@ -0,0 +1,89 @@
...
+        endpoint: 'http://localhost:8000',
+        replicationStatus: 'PENDING',
+    };
+
+    it('should reject when buckets is missing', async () => {
+        const options = { ...validOptions };
+        delete options.buckets;
+        await expect(listObjectsByReplicationStatus(options))
+            .rejects
+            .toThrow('No buckets given as input! Please provide a comma-separated list of buckets');
```
BourgoisMickael: I don't like that much to duplicate hardcoded string messages in tests.

Maybe those messages could be constants exported by that `listObjectsByReplicationStatus`, so we can modify the message more easily without having to update all tests.

Constants likes:
- `ERR_NO_BUCKET`
- `ERR_NO_ENDPOINT`
- `ERR_NO_ACCESS_KEY`
- ..

tests/unit/listObjectsByReplicationStatus.js
```diff
@@ -0,0 +1,89 @@
...
+            .not.toThrow('invalid REPLICATION_STATUS');
+    });
+
+    it('should reject when replicationStatus is not provided', async () => {
+        const options = { ...validOptions };
+        delete options.replicationStatus;
+        await expect(listObjectsByReplicationStatus(options))
+            .rejects
+            .toThrow('REPLICATION_STATUS not defined! Please provide a comma-separated list of replication statuses: NEW,PENDING,COMPLETED,FAILED,REPLICA.');
+    });
```
BourgoisMickael: All those validation rejection test can be dynamic in a for loop
```suggestion
    [
        { name: 'should reject when...', options: { ...validOptions, replicationStatus: null }, msg: '...' }
    ].forEach(t => 
    it(t.name, async () => {
        await expect(listObjectsByReplicationStatus(t.options)).rejects.toThrow(t.msg);
    });
```

listObjectsByReplicationStatus.js
tcarmet: it does, removed it.

listObjectsByReplicationStatus.js
tcarmet: done

listObjectsByReplicationStatus.js
tcarmet: done

listFailedObjects.js
tcarmet: set as print of `err.message` now.

tests/unit/listObjectsByReplicationStatus.js
tcarmet: fair enough!

tests/unit/listObjectsByReplicationStatus.js
tcarmet: done

tests/utils/S3Setup.js
tcarmet: done, added client-iam as a dev dep for now though, since it wasn't declared as a dep before.

BourgoisMickael (approved):

PR #360, opened 2025-12-29
S3UTILS-211: Install and configure workbench for tests

tcarmet:
Splitting up the work to ease review of #359. 

Setting up workbench in the CI environment for later use the in the functional tests of s3utils.

### DIFF
--- a/workbench/env/default/values.yaml
+++ b/workbench/env/default/values.yaml
@@ -1,0 +1,56 @@
+global:
+  log_level: info
+
+features:
+  scuba:
+    enabled: false
+    enable_service_user: true
+
+  bucket_notifications:
+    enabled: false
+    destinationAuth:
+      type: none
+      username: admin
+      password: admin123
+
+  cross_region_replication:
+    enabled: true
+
+  utapi:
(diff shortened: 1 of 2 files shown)


workbench/env/default/values.yaml
jonathan-gramain, line 1: It may be better to create a top-level `workbench` directory, under which to put `env`?

tcarmet: Do you mean `workbench/env/default/values.yaml` or `workbench/default/values.yaml`. We can also make it a hidden directory (like `.workbench/default/values.yaml`) if it's exposure you're worrying about.

jonathan-gramain: I mean, put everything workbench-related under the `workbench` directory, so it would be `workbench/env/default/values.yaml`, just to keep things tidy as this file is specific to workbench.

tcarmet: done

jonathan-gramain (approved):

dvasilas (approved):

PR #365, opened 2026-01-19
S3UTILS-216 Tool to detects missing permissions on CRR roles

nicolas2bert:


replicationAudit/check-replication-permissions.js
```diff
@@ -0,0 +1,480 @@
...
+            path: policyPart.substring(0, lastSlash + 1),
+            name: policyPart.substring(lastSlash + 1),
+        };
+    },
+};
+
+// ===========================================================================
+// Repd Client
+// ===========================================================================
+class RepdClient {
```
SylvainSenechal, line 131: nit: consider the folder Clients/RepdClient.js if you think another script could later use this client

nicolas2bert: I wanted to have a a single self-contained file with no dependencies.
Some reasons behind it:
- we can copy it easily 
 - it runs anywhere Node.js exists
 - No npm install needed on target

replicationAudit/check-replication-permissions.js
```diff
@@ -0,0 +1,480 @@
...
+
+    const jsonLength = buffer.readUInt32BE(0);
+    const totalLength = PROTOCOL.HEADER_SIZE + jsonLength;
+
+    if (buffer.length < totalLength) {
+        return null;
+    }
+
+    return {
+        data: JSON.parse(buffer.slice(PROTOCOL.HEADER_SIZE, totalLength).toString()),
```
anurag4DSB, line 81: Is it worth adding a try catch for JSON.parse?
When called from handleData() event handler, parsing errors will crash the script and leave TCP connections open

replicationAudit/check-replication-permissions.js
```diff
@@ -0,0 +1,480 @@
...
+const CONFIG = {
+    inputFile: process.argv[2] || '/root/buckets-with-replication.json',
+    leaderIp: process.argv[3] || '127.0.0.1',
+    outputFile: process.argv[4] || '/root/missing-replication-permissions.json',
+    repdPort: 4300,
+    dbName: 'vaultdb',
+    includePolicies: process.argv.includes('--include-policies'),
+    requestTimeoutMs: 10000,
+};
+
```
anurag4DSB, line 33: Is it intentional that the script uses process.argv for configuration (inputFile, leaderIp, outputFile, flags) instead of following the established environment variable pattern used by other scripts like crrExistingObjects.js and cleanupNoncurrentVersions.js.

anurag4DSB (commented):
Do we need functional tests for this? Also, critical code paths like RepdClient class, VaultKeys methods, checkBucketPermissions orchestration, and protocol handling are completely untested(unit tests)
Also, please do update the CLAUDE.md, the new replicationAudit/ directory introduces a significant new operational category but is not reflected in CLAUDE.md's script categories section.

replicationAudit/README.md
```diff
@@ -0,0 +1,436 @@
...
+      "ownerDisplayName": "testaccount",
+      "sourceRole": "arn:aws:iam::267390090509:role/crr-source-role"
+    }
+  ]
+}
+```
+
+### How It Works
+
+1. **Discovers buckets**: Queries the metastore via `/default/metastore` which contains all bucket metadata in a single request (more efficient than iterating raft sessions)
```
scality-gdoumergue: `/default/metastore` has been introduced in 9.3.0.0 (MD-897). This means the script is not compatible with versions below 9.3.0.0.
Instead the `users..bucket` database also contains the full list of buckets.

replicationAudit/README.md
```diff
@@ -0,0 +1,436 @@
...
+   (see above)
+
+2. Copy the script to the S3C supervisor:
+
+   ```bash
+   scp -i ~/.ssh/cloud replicationAudit/check-replication-permissions.js \
+       root@<supervisor-ip>:/root/
+   ```
+
+3. Copy the script to a store node:
```
scality-gdoumergue: A storage node is not necessarily a stateless S3 connector.
Instead of using saltstack in this procedure, use Ansible, and target one of the runners_s3 host (for example runners_s3[0]).

replicationAudit/README.md
```diff
@@ -0,0 +1,436 @@
...
+
+   This returns JSON like `{"ip":"10.160.116.162","port":4300}` - use the `ip` value.
+
+   **Note:** Vault metadata uses port 5300 for admin.
+
+5. Find the vault container ID:
+
+   ```bash
+   ssh -i ~/.ssh/cloud root@<supervisor-ip> \
+       "salt 'store-1' cmd.run 'crictl ps | grep scality-vault'"
```
scality-gdoumergue: `crictl ps | awk '/scality-vault/ {print $1}'`

replicationAudit/README.md
```diff
@@ -0,0 +1,436 @@
...
+
+**Script timeout**
+- Increase the salt timeout: `--timeout=300`
+- For large deployments, consider running directly on the store node via interactive SSH
+
+---
+
+## check-replication-permissions.js
+
+Checks if replication roles have `s3:ReplicateObject` permission by directly
```
scality-gdoumergue, line 138:
```suggestion
Checks if replication roles have `s3:ReplicateObject` permission for both
source and target buckets by directly
```

replicationAudit/README.md
```diff
@@ -0,0 +1,436 @@
...
+```bash
+# Variables
+SUPERVISOR=13.50.166.21
+STORE=store-1
+
+# Step 1: Copy scripts to supervisor
+scp -i ~/.ssh/cloud replicationAudit/list-buckets-with-replication.sh root@$SUPERVISOR:/root/
+scp -i ~/.ssh/cloud replicationAudit/check-replication-permissions.js root@$SUPERVISOR:/root/
+
+# Step 2: Copy scripts to store node
```
scality-gdoumergue: Instead of SSHing to the supervisor at each step, add an intermediate step here to connect to the sup's shell.

replicationAudit/list-buckets-with-replication.sh
```diff
@@ -0,0 +1,334 @@
...
+    fi
+    log "Connected to bucketd"
+}
+
+# Fetch all bucket names from the metastore
+# Streams directly to stdout to avoid holding full JSON in memory
+# Exits script on failure
+fetch_bucket_list() {
+    local response
+    response=$(curl -sf "$BUCKETD_URL/default/metastore") || {
```
scality-gdoumergue: I'm afraid this doesn't scale.
The `response` bash variable will need to store the name of thousands of buckets. Better directly pipe curl's stdout to the next jq command.

scality-gdoumergue: As commented in the README, this will only work with S3 >=9.3.0.0.

scality-gdoumergue (changes requested):

replicationAudit/check-replication-permissions.js
nicolas2bert: The risk is very low. Repd is a well-tested internal service that always returns valid JSON. A parse error would means there is a serious bug in repd itself or network corruption.

I would leave it as-is for this "diagnostic" script. The try-catch would be defensive overkill for a tool that runs occasionally and targets a trusted internal service.

replicationAudit/check-replication-permissions.js
nicolas2bert: It felt simpler for crictl exec workflow: 
`crictl exec $CONTAINER node script.js input.json $IP output.json`
instead of
`crictl exec -e VAR=value ...`

replicationAudit/list-buckets-with-replication.sh
nicolas2bert: good catch, I was sure we could get the buckets list paginated but ended up using the wrong API call `/default/metastore`.
Now, since `/default/bucket/users..bucket` is paginated, we should not have scalability issues.

anurag4DSB (approved):

replicationAudit/README.md
nicolas2bert: of course, thanks

replicationAudit/README.md
nicolas2bert: We are only checking for the source bucket replication roles policy.

replicationAudit/README.md
nicolas2bert: I like when each "steps" are quite independent/isolated.

replicationAudit/README.md
```diff
@@ -0,0 +1,450 @@
...
+       ansible -i env/<env>/inventory runners_s3[0] -m shell \
+       -a 'bash /root/list-buckets-with-replication.sh'"
+   ```
+
+4. Retrieve the output file:
+   ```bash
+   ssh root@<supervisor-ip> "cd /srv/scality/s3/s3-offline/federation && \
+       ansible -i env/<env>/inventory runners_s3[0] -m shell \
+       -a 'cat /root/buckets-with-replication.json'"
+   ```
```
scality-gdoumergue, line 44: 80% of our customers don't have `ssh` installed on their laptop.
add an intermediate step that thells the reader to connect through SSH to the supervisor, as root.
and then remove the unnecessary "ssh root@<supervisor-ip>"

replicationAudit/README.md
```diff
@@ -0,0 +1,450 @@
...
+       -a 'cp /root/check-replication-permissions.js /var/tmp/ && cp /root/buckets-with-replication.json /var/tmp/ && crictl exec $VAULT_CONTAINER node /var/tmp/check-replication-permissions.js /var/tmp/buckets-with-replication.json $LEADER_IP /var/tmp/missing.json'"
+   ```
+
+7. Retrieve the output:
+
+   ```bash
+   ssh root@<supervisor-ip> "cd /srv/scality/s3/s3-offline/federation && \
+       ansible -i env/<env>/inventory runners_s3[0] -m shell \
+       -a 'cat /var/tmp/missing.json'"
+   ```
```
scality-gdoumergue: same here: add an intermediate step to connect to the supervisor.

replicationAudit/README.md
```diff
@@ -0,0 +1,450 @@
...
+# Step 6: Copy files to /var/tmp and run the permission check script
+ssh root@$SUPERVISOR "cd $FEDERATION_DIR && \
+    ansible -i env/$ENV/inventory runners_s3[0] -m shell \
+    -a 'cp /root/check-replication-permissions.js /var/tmp/ && cp /root/buckets-with-replication.json /var/tmp/ && crictl exec $VAULT_CONTAINER node /var/tmp/check-replication-permissions.js /var/tmp/buckets-with-replication.json $LEADER_IP /var/tmp/missing.json'"
+
+# Step 7: Retrieve results
+ssh root@$SUPERVISOR "cd $FEDERATION_DIR && \
+    ansible -i env/$ENV/inventory runners_s3[0] -m shell \
+    -a 'cat /var/tmp/missing.json'"
+```
```
scality-gdoumergue: same here - add an intermediate step to connect to the sup

replicationAudit/list-buckets-with-replication.sh
```diff
@@ -0,0 +1,343 @@
...
+        batch_start=$((batch_start + BATCH_SIZE))
+        log "  Processed $processed/$bucket_count buckets (batch $batch_num)"
+    done
+
+    # Close results JSON array
+    echo "]" >> "$TMP_DIR/results.json"
+
+    # Calculate final stats
+    local repl_count
+    repl_count=$(jq 'length' "$TMP_DIR/results.json")
```
scality-gdoumergue: It works but it's ugly and over-complicating the code.
No need to enclose the results between square brackets and use `append_result` to have a list of json objects. You could have a non-comma-separated list of json objects in `$TMP_DIR/results.json` and use `jq --slurp` to make it an array when needed.

scality-gdoumergue (changes requested):

replicationAudit/list-buckets-with-replication.sh
nicolas2bert: I did not know about jq --slurp , thanks

scality-gdoumergue (approved):
thx!

nicolas2bert:
@bert-e reset

nicolas2bert:
@bert-e approve

PR #366, opened 2026-01-22
optimize count items with cache and clean for loop

SylvainSenechal:
do not merge


utils/S3UtilsMongoClient.js
maeldonn, line 116:
```suggestion
            const inflightsMap = {};
            for (const obj of inflights) {
                inflightsMap[obj._id] = obj.usedCapacity?._inflight || 0n;
            }

```

PR #367, opened 2026-01-23
S3UTILS-217: improve listObjectsByReplicationStatus.js

scality-gdoumergue:
- Show details of the errors when they happen
- Display the bucket's name
- Display the IsLatest field

This script has been tested in a lab:
- When the bucket doesn't exist: The script logs a 404 HTTP code error
- When the bucket is not versioned nor replicated: the script doesn't break
- When the ak/sk pair is wrong: The script logs a 403 HTTP code error
- When the IAM user doesn't have the right to list the bucket: The script logs a 403 HTTP code error
- When the IAM user doesn't have the right to HEAD objects: The script logs a 403 HTTP code error

### DIFF
--- a/listObjectsByReplicationStatus.js
+++ b/listObjectsByReplicationStatus.js
@@ -37,6 +37,7 @@ function _getKeys(list) {
     return list.map(v => ({
         Key: v.Key,
         VersionId: v.VersionId,
+        IsLatest: v.IsLatest,
     }));
 }
 
@@ -65,7 +66,7 @@ function _listBucket(s3, log, replicationStatusToProcess, bucket, cb) {
                 }
                 const keys = _getKeys(data.Versions || []);
                 return async.mapLimit(keys, 10, (k, next) => {
-                    const { Key, VersionId } = k;
+                    const { Key, VersionId, IsLatest } = k;
                     s3.send(new HeadObjectCommand({
                         Bucket: bucketName,
                         Key,
@@ -74,14 +75,30 @@ function _listBucket(s3, log, replicationStatusToProcess, bucket, cb) {
                         if (replicationStatusToProcess.includes(res.ReplicationStatus)) {
                             log.info('object with matching replication status found', {
                                 Key,
+                                VersionId,
+                                IsLatest,
                                 ReplicationStatus: res.ReplicationStatus,
-                                ...res
+                                ...res,
+                                bucketName
                             });
                         }
                         return next();
-                    }).catch(next);
+                    }).catch(err => {
+                        log.error('error getting object metadata', {
+                            bucketName,
+                            Key,
+                            VersionId,
+                            IsLatest,
+                            error: err
+                        });
+                        return next();
+                    });
                 }, err => {
                     if (err) {
+                        log.error('error processing batch of objects', {
+                            error: err,
+                            bucketName
+                        });
                         return done(err);
                     }
                     VersionIdMarker = data.NextVersionIdMarker;
--- a/tests/functional/listObjectsByReplicationStatus.js
+++ b/tests/functional/listObjectsByReplicationStatus.js
@@ -1,7 +1,7 @@
 const vaultclient = require('vaultclient');
 const { Logger } = require('werelogs');
-const { PutBucketVersioningCommand, PutBucketReplicationCommand, DeleteBucketReplicationCommand, PutObjectCommand, HeadObjectCommand } = require('@aws-sdk/client-s3');
-const { CreatePolicyCommand, CreateRoleCommand, AttachRolePolicyCommand, DetachRolePolicyCommand, DeleteRoleCommand, DeletePolicyCommand, DeleteUserCommand } = require('@aws-sdk/client-iam');
+const { PutBucketVersioningCommand, PutBucketReplicationCommand, DeleteBucketReplicationCommand, PutObjectCommand, HeadObjectCommand, DeleteObjectCommand, CreateBucketCommand, DeleteBucketCommand, DeleteObjectsCommand, ListObjectVersionsCommand } = require('@aws-sdk/client-s3');
+const { CreatePolicyCommand, CreateRoleCommand, AttachRolePolicyCommand, DetachRolePolicyCommand, DeleteRoleCommand, DeletePolicyCommand, DeleteUserCommand, CreateUserCommand, CreateAccessKeyCommand, AttachUserPolicyCommand } = require('@aws-sdk/client-iam');
 const { listObjectsByReplicationStatus } = require('../../listObjectsByReplicationStatus');
 
 const {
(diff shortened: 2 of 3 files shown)


tcarmet (approved):
Code itself LGTM, nothing to say here.

But, recently I spent some time to make sure we can test this script in the CI, previously there was none. So if you have the time in hand, and can have a look at [this file](https://github.com/scality/s3utils/blob/development/1/tests/functional/listObjectsByReplicationStatus.js) and if you are able to add some tests that would be great. If not, no worries, I'll take note and create a ticket on my end to do so.

Some notes:

Those tests do require a tool we created to simulate a small s3c env called [workbench](https://github.com/scality/workbench), to install it:

```shell
curl -L -o /tmp/workbench.tar.gz https://github.com/scality/workbench/releases/download/v0.8.0/workbench_Linux_x86_64.tar.gz
tar -xzf /tmp/workbench.tar.gz -C /tmp
sudo install /tmp/workbench /usr/local/bin/workbench
```

Run it inside s3utils repo, like so:

```shell
workbench up --env-dir ./workbench/env -d
```

You'll then be able to run those particular test like so:
```shell
yarn jest --verbose --logHeapUsage --projects jest.config.js  --coverage --testPathPattern='tests/functional/listObjectsByReplicationStatus.js'
```

Again, only if you have the time in hand, just let me know.

BourgoisMickael (approved):

listObjectsByReplicationStatus.js
```diff
@@ -74,14 +75,30 @@ function _listBucket(s3, log, replicationStatusToProcess, bucket, cb) {
...
                         }
                         return next();
-                    }).catch(next);
+                    }).catch(err => {
+                        log.error('error getting object metadata', {
+                            bucketName,
+                            Key,
+                            VersionId,
+                            IsLatest,
+                            error: err
```
jonathan-gramain, line 92: This can help logging error messages when the `message` field is dynamically computed:

```suggestion
                            error: err.message
```

listObjectsByReplicationStatus.js
```diff
@@ -74,14 +75,30 @@ function _listBucket(s3, log, replicationStatusToProcess, bucket, cb) {
...
+                        });
+                        return next();
+                    });
                 }, err => {
                     if (err) {
+                        log.error('error processing batch of objects', {
+                            error: err,
```
jonathan-gramain, line 99: same here

jonathan-gramain (approved):

scality-gdoumergue:
Thanks @tcarmet , I will use workbench to improve my knowledge!

tests/functional/listObjectsByReplicationStatus.js
```diff
@@ -214,7 +211,7 @@ describe('listObjectsByReplicationStatus', () => {
         await configureCrr(accountSource, accountDest);
     });
 
-    afterEach(async () => {
+    afterAll(async () => {
```
tcarmet: I'd like to leave those as afterEach, for further tests not be impacted by each other and really keep them isolated.

tests/functional/listObjectsByReplicationStatus.js
```diff
@@ -223,6 +220,46 @@ describe('listObjectsByReplicationStatus', () => {
         log.info('Test accounts deleted');
     });
 
+    it('should log an error when using credentials without access to the bucket', async () => {
+        // Create a test user with no bucket permissions
+        const noPermAccount = await createTestAccount(vaultClient);
```
tcarmet, line 227: Let me know what you think, but I don't think we have to bother into creating an account here, we can try to reach `accountSource.bucketName` with a some kind of fake accessKey and secretKey hardcoded string and expect a authorization denied and the script to fail accordingly with the right message.

tests/functional/listObjectsByReplicationStatus.js
```diff
@@ -223,6 +220,46 @@ describe('listObjectsByReplicationStatus', () => {
...
+            await deleteTestAccount(vaultClient, noPermAccount);
+        }
+
+        // The logs should contain at least one authorization error
+        const authErrorLog = capturedLogs.filter(entry => entry.name === 'AccessDenied');
+        expect(authErrorLog).toBeDefined();
+        expect(authErrorLog.length).toBe(1);
+    });
+
     it('should list objects by replication status', async () => {
```
tcarmet, line 454: Something that can be updated in this test is to ensure the bucket name is defined when listing an object, since this is something you have updated.

tests/functional/listObjectsByReplicationStatus.js
scality-gdoumergue: I prefer having a test that uses an existing key pair, which passes authentication, but not authorization

tcarmet: Seems fair 👌

tests/functional/listObjectsByReplicationStatus.js
```diff
@@ -223,6 +220,262 @@ describe('listObjectsByReplicationStatus', () => {
...
+        // Build a custom logger to capture errors
+        const capturedLogs = [];
+        const captureLogger = new Logger('s3utils:listObjectsByReplicationStatus:test');
+        const originalError = captureLogger.error.bind(captureLogger);
+        captureLogger.error = function (message, data) {
+            capturedLogs.push({ level: 'error', message, ...(data || {}) });
+            return originalError(message, data);
+        };
+
+        const endpoint = `http://${s3Host}:${s3Port}`;
```
tcarmet: Nit: I'm assuming I made this mistake in the first existing test but we might as well add it as a `const` in the `beforeAll` (or `Each`) to avoid redeclaring it.

tests/functional/listObjectsByReplicationStatus.js
```diff
@@ -223,6 +220,262 @@ describe('listObjectsByReplicationStatus', () => {
...
+                VersionId: v.VersionId,
+            }));
+            await accountSource.s3Client.send(new DeleteObjectsCommand({
+                Bucket: nonRepBucketName,
+                Delete: {
+                    Objects: objectsToDelete
+                }
+            }));
+        }
+        await accountSource.s3Client.send(new DeleteBucketCommand({ Bucket: nonRepBucketName }));
```
tcarmet: Should belong in a afterEach, can be part of of the deleteAccount function (I think there's already something that deletes objects that can be improved to support those deletes).

tests/functional/listObjectsByReplicationStatus.js
```diff
@@ -282,17 +535,21 @@ describe('listObjectsByReplicationStatus', () => {
...
         expect(foundObjectLogs.length).toBe(testObjects.length);
 
-        // Verify that all test objects were found
+        // Verify that all test objects were found and bucketName is present for each
         const foundKeys = foundObjectLogs.map(entry => entry.data.Key);
         for (const testObj of testObjects) {
             expect(foundKeys).toContain(testObj.Key);
+            // Expect the bucketName to be included in the log entry
+            const logEntry = foundObjectLogs.find(e => e.data.Key === testObj.Key);
+            expect(logEntry.data.bucketName).toBe(accountSource.bucketName);
```
tcarmet, line 527: nice, thank you.

tcarmet (approved):

PR #369, opened 2026-01-26
optimize count items on get buckets query 

SylvainSenechal:
ISSUE: S3UTILS-219

### DIFF
--- a/CountItems/CountManager.js
+++ b/CountItems/CountManager.js
@@ -44,7 +44,11 @@ class CountManager {
                 this.log.info('processing a bucket', {
                     method: 'CountManager::_setupQueue',
                     workInQueue: this.q.length(),
-                    bucketInfo,
+                    bucketInfo: {
+                        name: bucketInfo.getName(),
+                        uid: bucketInfo.getUid(),
+                        creationDate: bucketInfo.getCreationDate(),
+                    }
                 });
                 if (err) {
                     return done(err);
--- a/utils/S3UtilsMongoClient.js
+++ b/utils/S3UtilsMongoClient.js
@@ -2,6 +2,7 @@ const { MongoClientInterface } = require('arsenal').storage.metadata.mongoclient
 const { Long } = require('mongodb');
 const { errors, constants } = require('arsenal');
 const async = require('async');
+const { promisify } = require('util');
 const { validStorageMetricLevels } = require('../CountItems/utils/constants');
 const getLocationConfig = require('./locationConfig');
 const monitoring = require('./monitoring');
@@ -52,35 +53,6 @@ const baseMetricsObject = {
 };
 
 class S3UtilsMongoClient extends MongoClientInterface {
-    /**
-     * Get the list of buckets and their location dates
-     * @param {object} log - Werelogs logger
-     * @returns {object} - Object with bucket names as keys
-     * and their creation dates as values
-     */
-    async _getUsersBucketCreationDates(log) {
-        let cursorUsersBucketCreationDates;
-        try {
-            cursorUsersBucketCreationDates = await this.getCollection(USERSBUCKET).find({}, {
-                projection: {
-                    'value.creationDate': 1,
-                },
-            });
-            const usersBucketCreationDatesArray = await cursorUsersBucketCreationDates.toArray();
-            return usersBucketCreationDatesArray
-                .reduce((map, obj) => ({ ...map, [obj._id]: obj.value.creationDate }), {});
-        } catch (err) {
-            log.error('Failed to read __usersbucket collection', {
-                method: 'getUsersBucketCreationDates',
-                errDetails: { ...err },
-                errorString: err.toString(),
-            });
-            return null;
-        } finally {
-            await cursorUsersBucketCreationDates.close();
-        }
-    }
-
     async updateInflightDeltas(allMetrics, log) {
         let cursor;
         try {
@@ -100,13 +72,11 @@ class S3UtilsMongoClient extends MongoClientInterface {
 
             const inflights = await cursor.toArray();
             // convert inflights to a map with _id: usedCapacity._inflight
-            const inflightsMap = inflights.reduce((map, obj) => {
-                const inflightLong = obj.usedCapacity?._inflight || 0n;
-                return {
-                    ...map,
-                    [obj._id]: inflightLong,
-                };
-            }, {});
+            const inflightsMap = {};
+            for (const inflight of inflights) {
+                const inflightValue = inflight.usedCapacity?._inflight || 0n;
+                inflightsMap[inflight._id] = inflightValue;
+            }
 
             const accountInflights = {};
             allMetrics.forEach(entry => {
@@ -201,22 +171,21 @@ class S3UtilsMongoClient extends MongoClientInterface {
 
             const locationConfig = getLocationConfig(log);
 
-            const usersBucketCreationDatesMap = await this._getUsersBucketCreationDates(log);
+            let bucketCreationDate;
+            const getUsersBucketCreationDateAsync = promisify(this.getUsersBucketCreationDate).bind(this);
+            try {
+                bucketCreationDate = await getUsersBucketCreationDateAsync(bucketInfo.getOwner(), bucketName, log);
+            } catch (err) {
+                return callback(errors.InternalError);
+            }
 
             const bucketStatus = bucketInfo.getVersioningConfiguration();
             const isVer = (bucketStatus && (bucketStatus.Status === 'Enabled'
                 || bucketStatus.Status === 'Suspended'));
 
-            if (!usersBucketCreationDatesMap) {
-                return callback(errors.InternalError);
-            }
-
-            const bucketDate = usersBucketCreationDatesMap[`${bucketInfo.getOwner()}${constants.splitter}${bucketName}`];
-            if (bucketDate) {
-                bucketKey = `bucket_${bucketName}_${new Date(bucketDate).getTime()}`;
-                if (bucketKey) {
-                    inflightsPreScan = await this.readStorageConsumptionInflights(bucketKey, log);
-                }
+            if (bucketCreationDate) {
+                bucketKey = `bucket_${bucketName}_${new Date(bucketCreationDate).getTime()}`;
+                inflightsPreScan = await this.readStorageConsumptionInflights(bucketKey, log);
             }
 
             let startCursorDate = new Date();
@@ -243,7 +212,7 @@ class S3UtilsMongoClient extends MongoClientInterface {
                     bucketName,
                     bucketInfo,
                     entry,
-                    usersBucketCreationDatesMap[`${entry.value['owner-id']}${constants.splitter}${bucketName}`],
+                    bucketCreationDate,
                     isTransient,
                     locationConfig,
                     {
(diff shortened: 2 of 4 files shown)


CountItems/CountManager.js
```diff
@@ -44,7 +44,8 @@ class CountManager {
                 this.log.info('processing a bucket', {
                     method: 'CountManager::_setupQueue',
                     workInQueue: this.q.length(),
-                    bucketInfo,
+                    bucketName: bucketInfo._name,
```
SylvainSenechal: first little fix : logging bucketInfo would try to log a massive object with quite a lot of useless stuff, and an error "2026-01-20T14:08:08.645643941Z stdout F "[unable to serialize, circular reference is too complex to analyze]"
I think it's reasonable to only log this

utils/S3UtilsMongoClient.js
```diff
@@ -67,8 +67,13 @@ class S3UtilsMongoClient extends MongoClientInterface {
                 },
             });
             const usersBucketCreationDatesArray = await cursorUsersBucketCreationDates.toArray();
-            return usersBucketCreationDatesArray
-                .reduce((map, obj) => ({ ...map, [obj._id]: obj.value.creationDate }), {});
```
SylvainSenechal, line 71: The real problem, found by Maël : It's not even that we are querying all the buckets when we only need one (the perf in our test are still ok when doing this), but it's the reduce with spread operator : making a copy of the map on each iterations. Changing this from reduce to a simple loop reduce script run time from a few minutes to 2 seconds -_-

SylvainSenechal: btw, this function isn't used anymore, because as it turns out later, we don't need to query all buckets info, but only the info of a specific bucket and we have its name, so we can use another function for that.
I still kept this function, it's associated with tests, don't wanna touch too many things, anyways we are supposed to try to replace this script with SUR

utils/S3UtilsMongoClient.js
```diff
@@ -100,13 +107,14 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
             const inflights = await cursor.toArray();
             // convert inflights to a map with _id: usedCapacity._inflight
-            const inflightsMap = inflights.reduce((map, obj) => {
-                const inflightLong = obj.usedCapacity?._inflight || 0n;
-                return {
-                    ...map,
-                    [obj._id]: inflightLong,
-                };
-            }, {});
+            const inflightsMap = {};
```
SylvainSenechal, line 75: Similar opti : I know it feels good to be a functional programer but no need here

utils/S3UtilsMongoClient.js
```diff
@@ -201,22 +209,36 @@ class S3UtilsMongoClient extends MongoClientInterface {
 
             const locationConfig = getLocationConfig(log);
 
-            const usersBucketCreationDatesMap = await this._getUsersBucketCreationDates(log);
+            let bucketCreationDate;
```
SylvainSenechal, line 174: Real change : 
Given a bucketName and an OwnerId
**Before**
- query all buckets from __usersbucket collections
- Create a map with key: _id = ownerId+spliter+bucketName and value : creationDate
- then extracting from this map a value for the current owner of the bucket.

**Now** 
Query a single bucket based on bucketName and ownerId
=> no need to create a map, no need to loop through anything, we directly get the data we want

Also keeping same behavior as before : if bucket is not found : continue

SylvainSenechal: Also, this fix feels too simple and the gain too big to have been ignored, especially considering the fact that the function to query the specific bucket of a specific owner already existed. 
It makes me wonder if there is a gotcha, but can't find anything

CountItems/CountManager.js
francoisferrand: we may want to log a few other fields to fully identify the bucket?

```suggestion
                    bucketInfo: { name: bucketInfo._name, creationDate: bucketInfo._ creationDate, uid: bucketInfo._uid },
```

@benzekrimaha what do you think?

utils/S3UtilsMongoClient.js
francoisferrand: can be done functionally (but efficiently) as well:


```
            const inflightsMap = {};
            inflights.forEach(obj =>
                const inflightValue = inflight.usedCapacity?._inflight || 0n;
                inflightsMap[inflight._id] = inflightValue;
            );
```

utils/S3UtilsMongoClient.js
```diff
@@ -100,13 +107,14 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
-            const inflightsMap = inflights.reduce((map, obj) => {
-                const inflightLong = obj.usedCapacity?._inflight || 0n;
-                return {
-                    ...map,
-                    [obj._id]: inflightLong,
-                };
-            }, {});
+            const inflightsMap = {};
+            for (const inflight of inflights) {
+                if (!inflight?._id) {
```
francoisferrand: is this a real case?
I don't think it can happen with mongodb, having an id is mandatory?

utils/S3UtilsMongoClient.js
francoisferrand: There may be corner cases where we have multiple bucket (entries in DB) with the same name (but with different ID and creationDate obviously) : e.g. if we delete and quickly re-create a bucket maybe? The `creationDate` is used to uniquely build metrics per bucket, and ensure we don't mixup measures from the "old" vs "new" bucket.

Another thing it reminds me is that there is (or was) an issue on the `creationDate` : the bucket is created in 2 collections (`users bucket` & `metastore`), and at some point we could have (if unlucky) 2 different dates in each collection... So maybe the approach was added to handle such side effect?

Finally, another explanation could that code was initially optimized to make a single call at the beginning (so building a map was not a bad idea) ; but then there were issues with buckets missing from the "cache", and the call was moved (but unfortunately keeping the same mongo request and building a map)

Looking at the code, a few remarks:
- if we make this change, we should remove `_getUsersBucketCreationDates()` which is not used anymore?
- if `getUsersBucketCreationDate()` is only used here, we should make it async instead of promisifying; otherwise better to use `promisify()` instead of doing it manually
- `_getUsersBucketCreationDates()` used to filters on _id: should be useless I guess, but maybe we want to double-check?
- using `getUsersBucketCreationDate()` has the much better quality that we "compute" the ID in a single place, when making the call (i.e. in the "database layer" so to say)

If in doubt, should we split the PR in 2, to merge the map'reduce optimization quickly (not much risk I guess) and take the time to further ensure no issue for that part?

utils/S3UtilsMongoClient.js
```diff
@@ -201,22 +209,36 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+            } catch (err) {
+                if (err && err.message === 'Bucket entry not found') {
+                    log.warn('bucket entry not found in __usersbucket, continuing without inflights', {
+                        method: 'getObjectMDStats',
+                        bucketName,
+                        ownerId: bucketInfo.getOwner(),
+                    });
+                    bucketCreationDate = undefined;
+                } else {
+                    throw err;
```
```suggestion
                    return callback(errors.InternalError);
```

utils/S3UtilsMongoClient.js
```diff
@@ -201,22 +209,36 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                    });
+                });
+            } catch (err) {
+                if (err && err.message === 'Bucket entry not found') {
+                    log.warn('bucket entry not found in __usersbucket, continuing without inflights', {
+                        method: 'getObjectMDStats',
+                        bucketName,
+                        ownerId: bucketInfo.getOwner(),
+                    });
+                    bucketCreationDate = undefined;
```
francoisferrand: not needed, this is the default value (c.f. line 223: `let bucketCreationDate;`)

```suggestion
```

utils/S3UtilsMongoClient.js
```diff
@@ -201,22 +209,36 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                bucketCreationDate = await new Promise((resolve, reject) => {
+                    this.getUsersBucketCreationDate(bucketInfo.getOwner(), bucketName, log, (err, creationDate) => {
+                        if (err) {
+                            return reject(err);
+                        }
+                        return resolve(creationDate);
+                    });
+                });
+            } catch (err) {
+                if (err && err.message === 'Bucket entry not found') {
```
francoisferrand: `getUsersBucketCreationDate` is used in 2 places only : here and in collectucketMetricsAndUpdate, and it seems to me that in both case we ignore missing bucket info.
Also in that function, we explicitely detect that case and raise an exception.

It thus seems that it would greatly simplify code to directly return "undefined" creation date when the bucket is not found, instead of raising an error; and raise an exception only for an actual error.
This way we don't need extra try/catch and error filtering...

CountItems/CountManager.js
maeldonn:
```suggestions
                    bucketName: bucketInfo.getName(),
```

Prefer using getters

maeldonn: In case we are not sure which information we want to log, use this [method](https://github.com/scality/Arsenal/blob/development/8.2/lib/models/BucketInfo.ts#L385). It will surely provide more information than needed, but it avoids circular references.

SylvainSenechal: I added some more info, and used getters

utils/S3UtilsMongoClient.js
SylvainSenechal: right im too defensive here

utils/S3UtilsMongoClient.js
SylvainSenechal: Agree with the comment in general, I also saw that there was 2 db with 2 creation dates, and they are different -_-, in fact, this function takes bucketInfo as a parameter, which is already holding a creationDate, but different than the one we are getting with these db calls
I think we can keep my changes, remove the unused function, and turn getUsersBucketCreationDate into an async function

SylvainSenechal: actually i used promisify as the function is used somewhere else + it would require modifying tests, don't wanna turn this pr into 400 line changes

utils/S3UtilsMongoClient.js
SylvainSenechal: Yes I changed it to return undefined when nothing is found

utils/S3UtilsMongoClient.js
francoisferrand: > this function takes bucketInfo as a parameter, which is already holding a creationDate, but different than the one we are getting with these db calls

we should not change the "place" where we pick the `creationDate` : it is quite dirty we have 2, but today UI and/or sorbet most likely relies on the "current" source of truth for creation date....
If we pick the date from the other place, we may end up having inconsistent (or missing) output...

utils/S3UtilsMongoClient.js
```diff
@@ -1033,15 +1002,17 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
             });
             if (!res || !res.value || !res.value.creationDate) {
-                log.error('bucket entry not found in __usersbucket', {
+                log.warn('bucket entry not found in __usersbucket', {
+                    method: 'getUsersBucketCreationDate',
                     bucketName,
                     ownerId,
                 });
-                return cb(new Error('Bucket entry not found'));
+                return cb(null, undefined);
```
francoisferrand, line 1023: why are we ignoring error now?

francoisferrand: maybe need a comment explaining the behavior of this function on "not found" (in function's doc header?)

francoisferrand (approved):

utils/S3UtilsMongoClient.js
SylvainSenechal: yeah I changed it following your recommandations, as it is easier to handle the "no data case" like this

maeldonn (approved):

PR #372, opened 2026-02-10
Properly use sub clients of cloudserver client

SylvainSenechal:
ISSUE: S3UTILS-221

### DIFF
--- a/Clients/CloudserverClient.js
+++ b/Clients/CloudserverClient.js
@@ -55,14 +55,14 @@ class Client {
 
     getMetadata(params, callback) {
         const command = new GetMetadataCommand(params);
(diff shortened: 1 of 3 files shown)


Clients/CloudserverClient.js
SylvainSenechal, line 1: Should've seen it, not sure what happened but probably bumped Cloudserver client to latest version after doing the tests, needs to use sub clients like this (or could also directly import the sub client but this is fine)

SylvainSenechal: Also, not affecting dev/1.16

francoisferrand (approved):

maeldonn (approved):

PR #378, opened 2026-03-03
Seed zero-value metrics for empty buckets in count-items

delthas:
count-items only produced metric documents for buckets that hold objects. An empty bucket yielded no document in `__infostore`, so Scuba's deep health check (`getAnyMetric`) found nothing and returned 500, causing CloudServer to disable the quota service cluster-wide (ARTESCA-17063).

This seeds zero-value entries for the empty **bucket**, its **account** and its **location** in `getObjectMDStats()` when a bucket scan yields no objects, so every known resource always has a document in `__infostore`.

The bucket key is derived from the `__usersbucket` creation date, exactly like populated buckets (S3UTILS-131), so the seeded document matches the key quota lookups use. If that date is missing, the bucket is skipped with a warning rather than written under a key that lookups could never match.

Note: this addresses only the empty-bucket case. The new-bucket window (before the first count-items run) and the Scuba deep-health-check / alert behavior are tracked separately.

Issue: S3UTILS-224

### DIFF
--- a/tests/unit/utils/S3UtilsMongoClient.js
+++ b/tests/unit/utils/S3UtilsMongoClient.js
@@ -1059,6 +1059,50 @@ function uploadObjects(client, bucketName, objectList, callback) {
     }, callback);
 }
 
+describe('S3UtilsMongoClient::_seedEmptyBucketMetrics', () => {
+    const bucketInfo = BucketInfo.fromObj({ ...testBucketMD });
+    const locationConfig = { 'us-east-1': { objectId: 'us-east-1' } };
+    const bucketResource = `test-bucket_${testBucketCreationDate}`;
+
+    it('should seed zero-value bucket, account and location entries and return true', () => {
+        const collRes = { bucket: {}, location: {}, account: {} };
+        const log = { warn: sinon.spy() };
+        const seeded = mongoTestClient._seedEmptyBucketMetrics(collRes, bucketResource, bucketInfo, locationConfig, false, log);
+        assert.strictEqual(seeded, true);
+        assert.deepStrictEqual(Object.keys(collRes.bucket), [bucketResource]);
+        assert.deepStrictEqual(Object.keys(collRes.account), [testAccountCanonicalId]);
+        assert.deepStrictEqual(Object.keys(collRes.location), ['us-east-1']);
+        assert.strictEqual(log.warn.called, false);
+    });
+
+    it('should warn and seed nothing when no __usersbucket creation date is available', () => {
+        const collRes = { bucket: {}, location: {}, account: {} };
+        const log = { warn: sinon.spy() };
+        const seeded = mongoTestClient._seedEmptyBucketMetrics(collRes, undefined, bucketInfo, locationConfig, false, log);
+        assert.strictEqual(seeded, false);
+        assert.deepStrictEqual(collRes, { bucket: {}, location: {}, account: {} });
+        assert.strictEqual(log.warn.calledOnce, true);
+    });
+
+    it('should warn and seed nothing when the scan had processing errors', () => {
+        const collRes = { bucket: {}, location: {}, account: {} };
+        const log = { warn: sinon.spy() };
+        const seeded = mongoTestClient._seedEmptyBucketMetrics(collRes, bucketResource, bucketInfo, locationConfig, true, log);
+        assert.strictEqual(seeded, false);
+        assert.deepStrictEqual(collRes, { bucket: {}, location: {}, account: {} });
+        assert.strictEqual(log.warn.calledOnce, true);
+    });
+
+    it('should seed nothing when the bucket already produced metrics', () => {
+        const collRes = { bucket: { existing: {} }, location: {}, account: {} };
+        const log = { warn: sinon.spy() };
+        const seeded = mongoTestClient._seedEmptyBucketMetrics(collRes, bucketResource, bucketInfo, locationConfig, false, log);
+        assert.strictEqual(seeded, false);
+        assert.deepStrictEqual(collRes, { bucket: { existing: {} }, location: {}, account: {} });
+        assert.strictEqual(log.warn.called, false);
+    });
+});
+
 describe('S3UtilsMongoClient, tests', () => {
     const hr = 1000 * 60 * 60;
     let client;
@@ -1150,13 +1194,75 @@ describe('S3UtilsMongoClient, tests', () => {
         dataStore: 'us-east-1',
     };
 
+    const zeroUsedCapacity = {
+        current: 0n,
+        nonCurrent: 0n,
+        _currentCold: 0n,
+        _nonCurrentCold: 0n,
+        _currentRestored: 0n,
+        _currentRestoring: 0n,
+        _nonCurrentRestored: 0n,
+        _nonCurrentRestoring: 0n,
+        _incompleteMPUParts: 0n,
+    };
+    const zeroObjectCount = {
+        current: 0n,
+        nonCurrent: 0n,
+        _currentCold: 0n,
+        _nonCurrentCold: 0n,
+        _currentRestored: 0n,
+        _currentRestoring: 0n,
+        _nonCurrentRestored: 0n,
+        _nonCurrentRestoring: 0n,
+        _incompleteMPUUploads: 0n,
+        deleteMarker: 0n,
+    };
+    const zeroMetricsEntry = {
+        usedCapacity: zeroUsedCapacity,
+        objectCount: zeroObjectCount,
+    };
     const tests = [
         [
-            'getObjectMDStats() should return zero-result when no objects in the bucket',
+            'getObjectMDStats() should seed zero-value metrics for an empty bucket, account and location',
             {
                 bucketName: 'test-bucket',
                 isVersioned: false,
                 objectList: [],
+            },
+            {
+                dataManaged: {
+                    locations: { 'us-east-1': { curr: 0, prev: 0 } },
+                    total: { curr: 0, prev: 0 },
+                },
+                objects: 0,
+                stalled: 0,
+                versions: 0,
+                dataMetrics: {
+                    bucket: {
+                        [`test-bucket_${testBucketCreationDate}`]: {
+                            ...zeroMetricsEntry,
+                            locations: { 'us-east-1': { ...zeroMetricsEntry } },
+                        },
+                    },
+                    location: {
+                        'us-east-1': { ...zeroMetricsEntry },
+                    },
+                    account: {
+                        [testAccountCanonicalId]: {
+                            ...zeroMetricsEntry,
+                            locations: { 'us-east-1': { ...zeroMetricsEntry } },
+                        },
+                    },
+                },
+            },
+        ],
+        [
+            'getObjectMDStats() should not seed metrics when the bucket has no __usersbucket creation date',
+            {
+                bucketName: 'test-bucket-no-usersbucket',
+                isVersioned: false,
+                objectList: [],
+                skipUsersBucket: true,
             },
             {
                 dataManaged: {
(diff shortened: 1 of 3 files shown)


utils/S3UtilsMongoClient.js
```diff
@@ -403,6 +436,39 @@ class S3UtilsMongoClient extends MongoClientInterface {
             const retResult = this._handleResults(collRes, isVer);
             retResult.stalled = stalledCount;
 
+            // Empty buckets produce no metric document, which disables quotas
```
DarkIsDude: A 400 lines function are we are adding more logic, can we found something better ? I'm not sure the comment is really useful to understand the code ? Just explain the code. We can have context with git history if needed ?

benzekrimaha: this is re-implementing what `_handleResults` already does: every field is destructured with a 0n default, so an empty entry in collRes produces exactly the zero-value document (including the locations maps for bucket and account), there's no need to hand-build the shape we can just seed collRes before calling _handleResults

utils/S3UtilsMongoClient.js
```diff
@@ -403,6 +436,39 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                    const bucketLocation = bucketInfo.getLocationConstraint();
+                    if (bucketLocation && locationConfig && locationConfig[bucketLocation]) {
+                        const locationResource = this._getLocName(locationConfig[bucketLocation].objectId);
+                        if (!retResult.dataMetrics.location[locationResource]) {
+                            retResult.dataMetrics.location[locationResource] = zeroDataMetrics(false);
+                        }
+                    }
+                }
+            } else {
+                log.warn('unable to seed zero metrics for empty bucket: no creation date in __usersbucket', {
```
benzekrimaha: this warn is not conditioned on the bucket being empty: any bucket missing its `__usersbucket` creation date will log "unable to seed zero metrics for empty bucket" on every count-items round, even when it holds millions of objects (for which seeding is _irrelevant_ they already produce a document, just keyed `_NaN`).We should either gate it on the bucket actually being empty (`Object.keys(collRes.bucket).length === 0`), or reword; as written it will flood logs and mislead support

utils/S3UtilsMongoClient.js
```diff
@@ -403,6 +436,39 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+                    const owner = bucketInfo.getOwner();
+                    if (owner) {
+                        if (!retResult.dataMetrics.account[owner]) {
+                            retResult.dataMetrics.account[owner] = zeroDataMetrics(true);
+                        }
+                        // so the inflight block below tags it with the owner
+                        if (!accountBucket) {
+                            accountBucket = owner;
+                        }
+                    }
```
benzekrimaha: can accountBucket ever be set here? Any processed entry creates the `${bucketName}_${date}` bucket resource, so if we reach this block the cursor yielded nothing and accountBucket is necessarily undefined , the `if (!accountBucket)` check is dead. Same for` if (!retResult.dataMetrics.account[owner])` => account entries only come from object entries, which would have created the bucket entry and skipped the block. And` if (owner)` ,can a BucketInfo really have no owner? If not I think that we can drop the guard; defensive checks on impossible states can hide the actual invariants

utils/S3UtilsMongoClient.js
```diff
@@ -403,6 +436,39 @@ class S3UtilsMongoClient extends MongoClientInterface {
             const retResult = this._handleResults(collRes, isVer);
             retResult.stalled = stalledCount;
 
+            // Empty buckets produce no metric document, which disables quotas
+            // cluster-wide (ARTESCA-17063); seed zeros so every resource has one.
+            // Key off the __usersbucket creation date, like populated buckets (S3UTILS-131).
+            if (bucketCreationDate) {
+                const bucketResource = `${bucketName}_${new Date(bucketCreationDate).getTime()}`;
```
benzekrimaha: `bucketKey` => https://github.com/scality/s3utils/pull/378/changes#diff-3a778b361ddfba6106347ff0cb053bcbb1a84374fa0d7e6e6413e7c6fb6ca5cdR220 is already computed  for inflights lookup; computing${bucketName}${date} a second time inline invites the two drifting apart, we should compute the unprefixed resource name once next to `bucketKey` and derive both.

tests/unit/utils/S3UtilsMongoClient.js
```diff
@@ -1152,7 +1152,7 @@ describe('S3UtilsMongoClient, tests', () => {
 
     const tests = [
```
benzekrimaha, line 1224: can we add a case for it (empty bucket, getUsersBucketCreationDate returning nothing → no seeded entry, warn emitted) ?

utils/S3UtilsMongoClient.js
```diff
@@ -403,6 +436,39 @@ class S3UtilsMongoClient extends MongoClientInterface {
             const retResult = this._handleResults(collRes, isVer);
             retResult.stalled = stalledCount;
 
+            // Empty buckets produce no metric document, which disables quotas
+            // cluster-wide (ARTESCA-17063); seed zeros so every resource has one.
+            // Key off the __usersbucket creation date, like populated buckets (S3UTILS-131).
+            if (bucketCreationDate) {
```
benzekrimaha: `getObjectMDStats` is also called with isTransient buckets (being deleted). Do we want to seed a zero document for those too, or should seeding be skipped so a bucket in deletion doesn't get a fresh __infostore doc? Probably harmless since the next run drops it, but worth stating in the PR description.

utils/S3UtilsMongoClient.js
delthas: Extracted the seeding into a small `_seedEmptyBucketMetrics` helper so `getObjectMDStats` doesn't grow further, and trimmed the comment down to a single WHY line (the ARTESCA-17063 rationale isn't obvious from the code alone).

delthas: Done — dropped the hand-built shape and now seed empty entries into `collRes` before `_handleResults`. The existing `0n` destructuring defaults produce the exact zero-value document, including the `locations` maps for bucket and account.

utils/S3UtilsMongoClient.js
delthas: Gated both the seeding and the warning on the bucket actually being empty (`Object.keys(collRes.bucket).length === 0`), so a populated bucket missing its `__usersbucket` date no longer logs on every round.

utils/S3UtilsMongoClient.js
delthas: Dropped these guards — reaching the seed path already implies the scan yielded nothing (so `accountBucket` and the account entry are necessarily unset), and a bucket always has an owner.

utils/S3UtilsMongoClient.js
delthas: Good call — the unprefixed resource name is now computed once next to the inflight lookup and `bucketKey` is derived from it (`bucketKey = bucket_${bucketResource}`), and the seeding reuses that same value so they can't drift.

tests/unit/utils/S3UtilsMongoClient.js
delthas: Added — there's now an integration case for an empty bucket whose `__usersbucket` creation date is missing (nothing seeded, warning emitted), plus focused unit tests on the extracted `_seedEmptyBucketMetrics` covering the warn path and the no-op-when-already-populated path.

utils/S3UtilsMongoClient.js
delthas: `isTransient` here is the location constraint's transient-source flag, not the bucket's lifecycle state: `CountWorker.getIsTransient` derives it from `locationConfig[loc].isTransient` (via `isLocationTransient` / the Pensieve overlay), keyed off `bucketInfo.getLocationConstraint()`. It only affects data accounting in `_processEntryData` (source bytes are skipped once replication is COMPLETED). So the seeding isn't influenced by it — an empty bucket is seeded regardless of location type — and a bucket that is genuinely being deleted isn't reached through this flag.

utils/S3UtilsMongoClient.js
```diff
@@ -184,7 +185,9 @@ class S3UtilsMongoClient extends MongoClientInterface {
                 || bucketStatus.Status === 'Suspended'));
 
             if (bucketCreationDate) {
-                bucketKey = `bucket_${bucketName}_${new Date(bucketCreationDate).getTime()}`;
+                // creationDate comes from __usersbucket, like the per-object resource key (S3UTILS-131)
```
DarkIsDude: Why this comment ?

utils/S3UtilsMongoClient.js
```diff
@@ -432,6 +439,40 @@ class S3UtilsMongoClient extends MongoClientInterface {
         }
     }
 
+    /**
+     * Empty buckets otherwise produce no metric document, which disables quotas
+     * cluster-wide (ARTESCA-17063); seed zeros so _handleResults emits one.
```
DarkIsDude:
```suggestion
     * Seed zeros metrics for empty buckets
```

No need to describe the why, the use case, ...

utils/S3UtilsMongoClient.js
```diff
@@ -432,6 +439,40 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+     * @param{object} locationConfig - locationConfig.json
+     * @param{object} log - werelogs logger
+     * @returns{boolean} true when zero metrics were seeded, false otherwise
+     */
+    _seedEmptyBucketMetrics(collRes, bucketResource, bucketInfo, locationConfig, log) {
+        if (Object.keys(collRes.bucket).length > 0) {
+            return false;
+        }
+        if (!bucketResource) {
+            log.warn('unable to seed zero metrics for empty bucket: no creation date in __usersbucket', {
```
DarkIsDude: `no creation date in __usersbucket` when reading the code, the error message is not clear. You mention `__usersbucket` but it's in the previous context.

DarkIsDude (approved):

utils/S3UtilsMongoClient.js
delthas: Dropped it — the `getUsersBucketCreationDate` call just above already shows the source, and S3UTILS-131 / git history cover the nuance.

utils/S3UtilsMongoClient.js
delthas: Done — trimmed to a one-line description (`Seed zero-value metrics for empty buckets`), dropped the why/use-case.

utils/S3UtilsMongoClient.js
delthas: Reworded so it reads on its own while keeping the store pointer for support: `cannot seed empty-bucket metrics: bucket has no __usersbucket creation date entry`.

utils/S3UtilsMongoClient.js
```diff
@@ -432,6 +438,39 @@ class S3UtilsMongoClient extends MongoClientInterface {
...
+     * Seed zero-value metrics for empty buckets.
+     * @param{object} collRes - per-resource metrics accumulator, mutated in place
+     * @param{string} [bucketResource] - `${bucketName}_${creationDate}` key, absent when no __usersbucket date
+     * @param{object} bucketInfo - bucket attributes
+     * @param{object} locationConfig - locationConfig.json
+     * @param{object} log - werelogs logger
+     * @returns{boolean} true when zero metrics were seeded, false otherwise
+     */
+    _seedEmptyBucketMetrics(collRes, bucketResource, bucketInfo, locationConfig, log) {
+        if (Object.keys(collRes.bucket).length > 0) {
```
benzekrimaha, line 454: `object.keys(collRes.bucket).length === 0` does not necessarily mean the bucket is empty. It also happens when every entry failed processing, for example because of invalid content length or missing location configuration. We would then publish zero capacity for a non-empty bucket and potentially allow quota violations. can we track processing errors and only seed when the scan completed successfully without producing entries? we should also add a test for this case...

delthas: Good catch — you're right that an empty `collRes` can also mean every entry *failed* processing, not that the bucket is empty. Fixed: `getObjectMDStats` now tracks a `scanHadErrors` flag (set on any `_processEntryData` error), and `_seedEmptyBucketMetrics` refuses to seed (warns instead) when the scan had errors — so a bucket whose objects we couldn't account for is never seeded to zero; the next clean scan handles it. Added a focused unit test (errors → no seed) and an integration test (stub `_processEntryData` to error on a bucket that has an object, assert nothing is seeded). PHD-only / skipped-but-no-error buckets still seed, since those genuinely hold no committed data.

benzekrimaha (approved):
LGTM

PR #379, opened 2026-03-09
S3UTILS-225 [impr] list bucketd versions with full metadata

jonathan-gramain:
Create a new helper function that returns an async iterator on a bucket versions listing via bucketd, fetching the full locations array when it is pruned from the listing results.

It ensures to fetch the full metadata (including all locations) from the correct version, trying the different possible URLs for the specific case of null versions.

Note: unfortunately we don't have a clean high-level API in bucketd to do this, Cloudserver is normally doing this preprocessing to build the correct bucketd URL to fetch.

### DIFF
--- a/utils/async/httpRequest.js
+++ b/utils/async/httpRequest.js
@@ -1,0 +1,67 @@
+
+
+const http = require('http');
+const async = require('async');
+const { http: httpArsn } = require('httpagent');
+const werelogs = require('werelogs');
+
+const log = new werelogs.Logger('s3utils:httpRequest');
+
+const httpAgent = new httpArsn.Agent({
+    keepAlive: true,
+});
+
+/**
+ * Makes an HTTP request and returns the response with the full body
+ * buffered in res.body.
+ *
+ * @param {string} method - HTTP method (e.g. 'GET', 'DELETE')
+ * @param {string} url - full URL to request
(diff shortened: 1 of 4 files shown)


tmacro (approved):

utils/async/httpRequest.js
```diff
@@ -0,0 +1,67 @@
...
+ * buffered in res.body.
+ *
+ * @param {string} method - HTTP method (e.g. 'GET', 'DELETE')
+ * @param {string} url - full URL to request
+ * @param {object} [retryParams] - if provided, failed requests are retried
+ *   via async.retry with these params (e.g. { times: 100, interval: 5000 });
+ *   both network errors and 5xx responses are treated as retryable failures
+ * @returns {Promise<http.IncomingMessage>} response object with body in res.body
+ */
+function httpRequest(method, url, retryParams) {
```
tcarmet, line 25: This function doesn't seem to be tested, it is mocked on the other tests if I got it right, so a small simple unit test for this one would be nice. Nothing fancy, a simple GET is enough I think.

tcarmet (approved):
Nit comment, but LGTM

utils/async/httpRequest.js
jonathan-gramain: added unit tests

tcarmet (approved):

PR #380, opened 2026-03-13
S3UTILS-226 MPU orphan cleanup script

jonathan-gramain:
Implement cleanupMpuOrphans.js to delete orphaned parts and sproxyd keys.

Orphaned parts are part metadata which do not correspond to an overview key with the same upload ID.

Sproxyd keys are only removed if no completed MPU object has a matching upload ID in the same bucket.

### DIFF
--- a/cleanupMpuOrphans.js
+++ b/cleanupMpuOrphans.js
@@ -1,0 +1,577 @@
+/* eslint-disable no-console */
+const async = require('async');
+
+const werelogs = require('werelogs');
+
+const httpRequest = require('./utils/async/httpRequest');
+const listVersions = require('./utils/async/bucketd/listVersions');
+
+const DEFAULT_LISTING_PAGE_SIZE = 1000;
+const DEFAULT_LOG_PROGRESS_INTERVAL = 10;
+const RETRY_PARAMS = { times: 100, interval: 5000 };
+
+const {
+    BUCKETD_HOSTPORT, SPROXYD_HOSTPORT,
+    BUCKETS, RAFT_SESSIONS,
+} = process.env;
+
+const VERBOSE = process.env.VERBOSE === '1';
+const TRACE = process.env.TRACE === '1';
+
+let logLevel;
+if (TRACE) {
+    logLevel = 'trace';
+} else if (VERBOSE) {
+    logLevel = 'debug';
+} else {
+    logLevel = 'info';
+}
+werelogs.configure({ level: logLevel, dump: 'error' });
+
+const LISTING_PAGE_SIZE = (
+    process.env.LISTING_PAGE_SIZE
+        && Number.parseInt(process.env.LISTING_PAGE_SIZE, 10))
+      || DEFAULT_LISTING_PAGE_SIZE;
+
+const LOG_PROGRESS_INTERVAL = (
+    process.env.LOG_PROGRESS_INTERVAL
+        && Number.parseInt(process.env.LOG_PROGRESS_INTERVAL, 10))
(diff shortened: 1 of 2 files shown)


jonathan-gramain:
ping

arnaudlangloisscality (approved):
lgtm

cleanupMpuOrphans.js
```diff
@@ -0,0 +1,575 @@
...
+    );
+    const orphanCount = Object.keys(orphanMap).length;
+    status.orphanedUploadIds = orphanCount;
+    logProgress('discovery phase complete');
+    if (orphanCount === 0) {
+        return;
+    }
+
+    for (const [uploadId, info] of Object.entries(orphanMap)) {
+        log.info('orphaned MPU found', {
```
scality-fno, line 497: Don't know if the Usage section is meant to describe the exact _sequence_ of the log messages, but if it is, the code contradicts it: the "orphaned MPU found" message comes last. Is this intentional?

cleanupMpuOrphans.js
```diff
@@ -0,0 +1,575 @@
...
+const TRACE = process.env.TRACE === '1';
+
+let logLevel;
+if (TRACE) {
+    logLevel = 'trace';
+} else if (VERBOSE) {
+    logLevel = 'debug';
+} else {
+    logLevel = 'info';
+}
```
scality-fno, line 28: nitpick: (Claude) Those three entries (deleted orphaned sproxyd key, not deleting sproxyd key used by completed MPU, deleted orphaned part metadata) are logged at debug level. They show up when either VERBOSE=1 or TRACE=1. Labelling them [VERBOSE ONLY] could confuse someone who set TRACE=1 and wonders why they're seeing those lines. [VERBOSE or TRACE] or just [requires VERBOSE=1 or TRACE=1] would be more accurate.

cleanupMpuOrphans.js
```diff
@@ -0,0 +1,575 @@
...
+    This script takes care of cleaning up those orphaned parts data and
+    metadata keys with a two-phase process for each target bucket:
+
+    - Discovery phase: builds a map of orphaned MPU upload IDs present
+      in the bucket
+
+    - Cleanup phase: scans the bucket with a versioned listing,
+      matches any completed MPU with their orphaned counterpart to
+      detect used sproxyd keys, and only deletes the unused ones along
+      with the orphaned metadata
```
scality-fno: Doc clarity: Claude: That's accurate for the case where the orphaned upload ID matches a completed object. But the code also handles the complementary case: orphaned entries that don't match any completed version in the bucket are deleted entirely (all their sproxyd keys, unconditionally). 

```suggestion
    - Cleanup phase: scans the bucket with a versioned listing,
      matches any completed MPU with their orphaned counterpart to
      detect used sproxyd keys, and only deletes the unused ones along
      with the orphaned metadata. Any orphaned upload IDs that were not
      matched to any completed object version are also deleted 
      unconditionally.
```

scality-fno (approved):
Strictly from a doc perspective. 

I'm ok with the documentation. Some clarification could be beneficial. Not a blocker, though.

cleanupMpuOrphans.js
jonathan-gramain: I put the messages in roughly the order in which they should come indeed. I think "orphaned MPU found" is correctly placed since it occurs during the first discovery phase, before the cleanup phase.

cleanupMpuOrphans.js
jonathan-gramain: Indeed, I can do that change.

jonathan-gramain:
ping

PR #381, opened 2026-03-31
S3UTILS-227: replace util.parseArgs with manual argv parsing

nicolas2bert:
**Design note:**
`fix-missing-replication-permissions.js` is a standalone script meant to be copied into the vault container of older S3C versions. It does not run inside the s3utils container. It has to be compatible with the Node and npm packages available in the target vault container.

This "copy a self-contained script into an older container" approach is not easy to implement, test, or maintain, and is not a good practice in general.

We accepted this trade-off after discussing with Product and CS to allow fast integration on customer environments.
 
This tool should not be needed once we update replication permissions for all clients using CRR.

**Problem:**
`fix-missing-replication-permissions.js` uses `util.parseArgs` which requires Node 16.17+. The vault container ships Node 16.13.2 on all S3C versions up to 9.5.0.x, causing "parseArgs is not a function" at runtime.

**Fix:**
Replace `util.parseArgs` with manual `process.argv` parsing. No other dependency changes. There is no risk with other dependencies: vaultclient, aws-sdk v2, and process.argv all work on Node 16.13.2+.

**Testing:**
Existing functional tests in `fixMissingReplicationPermissions.js` run the script as a child process through its CLI interface (positionals, --iam-port, --https, --dry-run). Since the CLI contract is unchanged, passing tests confirm the new parsing logic is equivalent.

### DIFF
--- a/replicationAudit/fix-missing-replication-permissions.js
+++ b/replicationAudit/fix-missing-replication-permissions.js
@@ -60,6 +60,9 @@ function getConfig() {
 
     for (let i = 0; i < args.length; i++) {
         if (args[i] === '--iam-port') {
+            if (i + 1 >= args.length) {
+                throw new Error('Missing value for --iam-port');
+            }
             iamPort = parseInt(args[++i], 10);
         } else if (args[i] === '--https') {
             useHttps = true;
--- a/replicationAudit/fix-missing-replication-permissions.js
+++ b/replicationAudit/fix-missing-replication-permissions.js
@@ -30,7 +30,6 @@
 const fs = require('fs');
 const http = require('http');
 const https = require('https');
-const { parseArgs } = require('util');
 const { Client: VaultClient } = require('vaultclient');
 const AWS = require('aws-sdk');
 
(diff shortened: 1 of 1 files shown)


dvasilas (approved):

replicationAudit/fix-missing-replication-permissions.js
```diff
@@ -49,17 +48,30 @@ function log(message) {
...
 function getConfig() {
-    const { values, positionals } = parseArgs({
-        allowPositionals: true,
-        options: {
-            'iam-port': { type: 'string', default: '8600' },
-            'https': { type: 'boolean', default: false },
-            'dry-run': { type: 'boolean', default: false },
-        },
-    });
+    const args = process.argv.slice(2);
```
anurag4DSB, line 54: My thoughts but consolidated using claude LLM:
The current implementation works, but the parsing logic is tightly coupled, order-dependent, and will become harder to maintain as flags grow.

A few specific concerns:

- We’re manually mutating the loop index (i++) to consume values, which makes the control flow harder to reason about and easier to break when adding new flags.
- Validation is inconsistent, parseInt is used without checking for NaN, which can lead to silent misconfiguration.
- The if/else chain does not scale; every new flag increases cognitive load and duplication.
Parsing, validation, and state mutation are all interleaved, making this difficult to test in isolation.


If we want to solve an already solved problem(see last paragraph), I’d recommend moving toward a table-driven approach where each flag maps to a handler. That separates concerns and makes the system extensible without modifying the core loop. At a minimum, a cas-switch-based structure with explicit validation would improve readability and safety.

More broadly, this is reimplementing CLI parsing logic that libraries like minimist or commander already handle robustly (including defaults, type coercion, aliases, and help text). If this script is expected to evolve beyond a couple of flags, introducing a lightweight dependency would reduce long-term complexity and edge cases.

nicolas2bert: This ephemeral script (will be deleted o nce we update replication permissions for all clients using CRR cf PR description) is meant to be copied into older vault containers where only a minimal set of dependencies exist. 

Using `minimist` would introduce an additional dependency, and we would need to ensure it behaves consistently across different Vault versions. 

for 3 flags, the manual approach is arguably fine. The commit message on HEAD (ec54481) even says "replace util.parseArgs with manual argv parsing" -> so this was a deliberate choice for compatibility.

anurag4DSB (approved):

nicolas2bert:
@bert-e approve

PR #388, opened 2026-05-11
S3UTILS-236: surface non-404 sproxyd errors in completed scan summary

anurag4DSB:
## Intent: why does this change exist?

`verifyBucketSproxydKeys.js` has had a silent reporting bug since its first commit: the `logProgress` emission read `status.objectErrors` (typo, missing the `s`) while the counter is named `status.objectsErrors`. `JSON.stringify` drops `undefined` values, so the `errors` field vanished from the `completed scan` summary line whenever non-404 sproxyd HTTP errors occurred. Exit code `103` was correct, but consumers reading the summary log saw a false-clean output. Surfaced by a customer scan with ~27,000 HTTP 422 responses (RD-758).

## System impact: what's affected, including downstream?

The `completed scan` and periodic-progress JSON log lines will now carry an `errors` field whenever non-404 non-2xx sproxyd responses occur. Log consumers that parse the summary line (operators, log aggregators) will see a counter that was previously absent. Exit codes, per-event log lines, and all other counter values are unchanged.

## Preserved behavior: what explicitly stays the same?

- Process exit codes (101 / 102 / 103 / 104 / 105) — same logic, same triggers.
- Per-object `level:"error"` log lines — unchanged.
- The four existing summary counters (haveMissingKeys, haveDupKeys, haveEmptyMetadata, haveDupVersionIds) — same accounting.
- `NO_MISSING_KEY_CHECK`, `VERBOSE`, and other env-gated behaviours.
- CLI invocation path: `main()` runs identically when the file is executed directly.

## Intended change: what's different after this PR?

- `status.objectErrors` → `status.objectsErrors` on line 199 of `verifyBucketSproxydKeys.js`.
- `main()` is gated behind `if (require.main === module)` so the file can be required from a unit test without triggering a real scan.
- `module.exports` extended from `{ httpRequest }` to `{ httpRequest, status, log, logProgress }` for the same test.
- New behavioural unit test in `tests/unit/verifyBucketSproxydKeys.js` that sets the counter directly, calls `logProgress`, and asserts the emitted summary contains the expected `errors` value. Two cases: customer's 27,152 number and a clean-scan zero case.

## Verification: how do we know this worked, or how would we know if it didn't?

`yarn test:unit tests/unit/verifyBucketSproxydKeys.js` passes (447 / 447). Negative-verified: temporarily reverting the typo fix makes the new tests fail with `Received: {"errors": undefined, ...}` — the exact symptom the customer reported.

Refs: S3C-11181, RD-758

### DIFF
--- a/tests/unit/verifyBucketSproxydKeys.js
+++ b/tests/unit/verifyBucketSproxydKeys.js
@@ -168,3 +168,76 @@ describe('verifyBucketSproxydKeys', () => {
         });
     });
 });
+
+// Regression test for the bug where `logProgress` read `status.objectErrors`
+// (typo — missing the 's') instead of `status.objectsErrors`. Because
+// JSON.stringify drops undefined values, the `errors` field was silently
+// missing from the summary log line even when non-404 sproxyd errors had
+// been counted. A customer scan with ~27,000 HTTP 422s looked clean as a
+// result. The test below sets the counter, calls logProgress, and checks
+// the emitted summary contains the expected `errors` value.
+
+// The script reads these env vars at startup and exits if any are missing,
+// so we set them before the require. The hosts are fake — no real requests
+// are made because main() doesn't run when the file is required from a test.
+process.env.SPROXYD_HOSTPORT = process.env.SPROXYD_HOSTPORT || 'fake-sproxyd:9999';
+process.env.BUCKETD_HOSTPORT = process.env.BUCKETD_HOSTPORT || 'fake-bucketd:9998';
+process.env.BUCKETS = process.env.BUCKETS || 'test-bucket';
+
+// The script registers a periodic-progress timer at startup. If we leave it
+// alone the timer keeps the Jest worker alive after the tests finish and
+// Jest crashes with "child process exceptions". Swap setInterval for a
+// no-op while we require the file, then put it back so Jest's own internal
+// timers continue to work.
+const origSetInterval = global.setInterval;
+global.setInterval = () => undefined;
+const vbsk = require('../../verifyBucketSproxydKeys');
+global.setInterval = origSetInterval;
+
+describe('verifyBucketSproxydKeys — summary line emits errors count (S3UTILS-236)', () => {
+    let logSpy;
+
+    beforeEach(() => {
+        // Reset the counters so each test starts from zero.
+        Object.keys(vbsk.status).forEach(k => {
+            if (typeof vbsk.status[k] === 'number') {
+                vbsk.status[k] = 0;
+            }
+        });
+        // Spy on the logger so we can inspect what logProgress emits.
+        logSpy = jest.spyOn(vbsk.log, 'info').mockImplementation();
+    });
+
+    afterEach(() => {
+        logSpy.mockRestore();
+    });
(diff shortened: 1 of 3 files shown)


anurag4DSB:
ping

tests/unit/verifyBucketSproxydKeys.js
```diff
@@ -168,3 +168,76 @@ describe('verifyBucketSproxydKeys', () => {
...
+        });
+        // Spy on the logger so we can inspect what logProgress emits.
+        logSpy = jest.spyOn(vbsk.log, 'info').mockImplementation();
+    });
+
+    afterEach(() => {
+        logSpy.mockRestore();
+    });
+
+    test('emits errors count from status.objectsErrors when non-404 sproxyd errors have been counted', () => {
```
BourgoisMickael: Test names should start with `should`


```suggestion
    test('should emit errors count from status.objectsErrors when non-404 sproxyd errors have been counted', () => {
```

BourgoisMickael (approved):

tests/unit/verifyBucketSproxydKeys.js
```diff
@@ -168,3 +168,76 @@ describe('verifyBucketSproxydKeys', () => {
...
+        // would vanish from the output entirely — so we assert both the
+        // value and the presence of the key.
+        vbsk.status.objectsScanned = 100;
+        vbsk.status.objectsErrors = 0;
+
+        vbsk.logProgress('completed scan');
+
+        const emitted = logSpy.mock.calls[0][1];
+        expect(emitted.errors).toBe(0);
+        expect('errors' in emitted).toBe(true);
```
jonathan-gramain: I would reverse the two tests as the first check on `emitted.errors` may only be useful if the second expect succeeds.

jonathan-gramain (approved):

tmacro (approved):

tests/unit/verifyBucketSproxydKeys.js
anurag4DSB: Done

tests/unit/verifyBucketSproxydKeys.js
anurag4DSB: Done, sounds good to me.

tests/unit/verifyBucketSproxydKeys.js
```diff
@@ -168,3 +168,76 @@ describe('verifyBucketSproxydKeys', () => {
...
+process.env.BUCKETS = process.env.BUCKETS || 'test-bucket';
+
+// The script registers a periodic-progress timer at startup. If we leave it
+// alone the timer keeps the Jest worker alive after the tests finish and
+// Jest crashes with "child process exceptions". Swap setInterval for a
+// no-op while we require the file, then put it back so Jest's own internal
+// timers continue to work.
+const origSetInterval = global.setInterval;
+global.setInterval = () => undefined;
+const vbsk = require('../../verifyBucketSproxydKeys');
```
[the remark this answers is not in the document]
anurag4DSB: Its tests and intentional for context.

PR #392, opened 2026-05-20
S3UTILS-235: Include logrotate in s3utils

BourgoisMickael:
As object-repair will run through raft-dispatcher,
Log rotation will be handled by logrotate instead of supervisord

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3utils",
-  "version": "1.17.10",
+  "version": "1.17.11",
   "engines": {
     "node": ">= 22"
   },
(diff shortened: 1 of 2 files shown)


tcarmet (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "s3utils",
-  "version": "1.17.9",
+  "version": "1.17.10",
```
jonathan-gramain: I already bumped to 1.17.10, sorry didn't know you would also need a bump

jonathan-gramain (approved):

BourgoisMickael:
@bert-e reset

PR #394, opened 2026-05-22
Bugfix/s3 utils 239 fix count items bucket logging

benzekrimaha:
In the PR we are nullifying _bucketLoggingStatus which is safe for count-items specifically: that this metadata is unrelated to object count / storage usage computation and  only exists on the BucketInfo object because it is part of the full bucket metadata model.
The implementation we now have is cleaner and makes this explicit by naming the fields as unsupported serialized BucketInfo fields for count-items. 
The rationale being that : class-backed fields that count-items does not consume should not be rehydrated in the worker path just to count object metadata.

Issue: S3UTILS-239

### DIFF
--- a/CountItems/CountWorker.js
+++ b/CountItems/CountWorker.js
@@ -5,6 +5,7 @@ const monitoring = require('../utils/monitoring');
 const { deserializeBigInts, serializeBigInts } = require('./utils/utils');
 
 const PENSIEVE = 'PENSIEVE';
+
 class CountWorker {
     constructor(params) {
         this.log = params.log;
@@ -66,12 +67,16 @@ class CountWorker {
         if (!this.client.client) {
             return callback(new Error('NotConnected'));
         }
-        // 'fromObj' expects that the website configuration is an instance of
-        // WebsiteConfiguration as it is not used in CountItems, we nullify it.
-        if (bucketInfoObj._websiteConfiguration) {
-            Object.assign(bucketInfoObj, { _websiteConfiguration: null });
+        const bucketInfoData = bucketInfoObj.bucketInfo || bucketInfoObj;
+        // 'fromObj' expects class-backed fields to already be instances.
+        // They are not used in CountItems, so we nullify unsupported serialized fields.
+        if (bucketInfoData._websiteConfiguration) {
+            bucketInfoData._websiteConfiguration = null;
         }
-        const bucketInfo = BucketInfo.fromObj(bucketInfoObj.bucketInfo || bucketInfoObj);
+        if (bucketInfoData._bucketLoggingStatus) {
+            bucketInfoData._bucketLoggingStatus = null;
+        }
+        const bucketInfo = BucketInfo.fromObj(bucketInfoData);
         const bucketName = bucketInfo.getName();
         this.log.info(`${process.pid} handling ${bucketName}`);
         return async.waterfall([
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3utils",
-  "version": "1.16.7",
+  "version": "1.16.8",
   "engines": {
     "node": ">= 22"
   },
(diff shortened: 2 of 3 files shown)


SylvainSenechal (approved):

CountItems/CountWorker.js
```diff
@@ -5,6 +5,20 @@ const monitoring = require('../utils/monitoring');
 const { deserializeBigInts, serializeBigInts } = require('./utils/utils');
 
 const PENSIEVE = 'PENSIEVE';
+const unsupportedSerializedBucketInfoFields = [
```
SylvainSenechal: could have kept the comment about why we do this 🤔

maeldonn: nit:
```suggestion
const UNSUPPORTED_SERIALIZED_BUCKET_INFO_FIELDS = [
```

CountItems/CountWorker.js
```diff
@@ -5,6 +5,20 @@ const monitoring = require('../utils/monitoring');
...
+    '_bucketLoggingStatus',
+    '_websiteConfiguration',
+];
+
+function normalizeBucketInfoForCountItems(bucketInfoObj) {
```
maeldonn: Using object spread inside a reduce loop creates a brand new copy of the object on every iteration, leading to big time complexity and heavy memory churn. We can optimize this by cloning the object just once and using a for...of loop to safely update the properties.


```suggestion
function normalizeBucketInfoForCountItems(bucketInfoObj) {
    const normalized = { ...bucketInfoObj }; 
    
    for (const field of unsupportedSerializedBucketInfoFields) {
        if (normalized[field]) {
            normalized[field] = null;
        }
    }
    
    return normalized;
}
```

maeldonn (approved):
Once `normalizeBucketInfoForCountItems` updated, it will be good for me !

CountItems/CountWorker.js
```diff
@@ -66,12 +80,8 @@ class CountWorker {
...
             return callback(new Error('NotConnected'));
         }
-        // 'fromObj' expects that the website configuration is an instance of
-        // WebsiteConfiguration as it is not used in CountItems, we nullify it.
-        if (bucketInfoObj._websiteConfiguration) {
-            Object.assign(bucketInfoObj, { _websiteConfiguration: null });
-        }
-        const bucketInfo = BucketInfo.fromObj(bucketInfoObj.bucketInfo || bucketInfoObj);
+        const normalizedBucketInfo = normalizeBucketInfoForCountItems(bucketInfoObj.bucketInfo || bucketInfoObj);
+        const bucketInfo = BucketInfo.fromObj(normalizedBucketInfo);
```
francoisferrand: does it really make sense to do this?

sticking with the original (or just erase the field there) is much simpler, and more efficient:

```suggestion
        if (bucketInfoObj._websiteConfiguration) {
        	bucketInfoObj._websiteConfiguration = null;
        }
        if (bucketInfoObj._bucketLoggingStatus) {
        	bucketInfoObj._bucketLoggingStatus = null;
        }
        const bucketInfo = BucketInfo.fromObj(normalizedBucketInfo);
```
or maybe (less conservative, maybe share off a bit...but not sure if it may create more risk):
```suggestion
        if (bucketInfoObj._websiteConfiguration) {
        	bucketInfoObj.delete('_websiteConfiguration');
        }
        if (bucketInfoObj._bucketLoggingStatus) {
        	bucketInfoObj.delete('_bucketLoggingStatus');
        }
        const bucketInfo = BucketInfo.fromObj(normalizedBucketInfo);
```

francoisferrand (changes requested):
Current change introduces a performance impact, which is not worth as already pointed out (esp. considering count items is end of life!)

francoisferrand (approved):
LGTM.

Do we need the fix on 1.16? I see zenko 2.14 and 2.15 both use s3utils 1.17, and not sure if the problem can be reproduced with zenko 2.13 (which uses cloudserver 9.1.14)
- if not needed → rebase on 1.17
- if needed → add a bump commit on waterflow branch & release both 1.16 and 1.17

PR #396, opened 2026-05-28
✨ S3UTILS-230 implement CRR multi-site V2 replication format

DarkIsDude:
## Summary

- Add V2 replication format support to \`ReplicationStatusUpdater\`, implementing the [CRR to Multiple Sites design](../citadel.git/docs/design/replication/crr-multiple-sites.md)
- V2 configs (rules with a \`Filter\` element) route through a new \`_markObjectPendingV2\` path that writes per-backend \`destination\`/\`role\` and removes legacy top-level \`storageClass\`/\`storageType\`/\`destination\`
- V1 configs (no \`Filter\` on any rule) retain existing behavior unchanged
- Arsenal static methods (\`resolveBackends\`, \`resolveSourceRole\`, \`resolveDestinationRole\`) are reused instead of maintaining parallel implementations; arsenal dependency pinned to \`improvement/ARSN-571/crr-multi\`

## Changes

**\`CRR/ReplicationStatusUpdater.js\`**
- \`_isV1Format()\` — V2 detection aligned with arsenal: a rule is V2 if it carries a \`Filter\` element
- \`_getMatchingRules()\` — prefix-matches object key against enabled rules, deduplicates by site (highest priority wins); feeds the adapter for \`resolveBackends\`
- \`_computeTopLevelStatus()\` — aggregates backend statuses: FAILED > PROCESSING > COMPLETED
- \`_removeV1Fields()\` — strips legacy top-level \`storageClass\`/\`storageType\`/\`destination\` when migrating to V2
- \`_markObjectPendingV2()\` — V2 metadata writer: adapts AWS SDK rule shape to arsenal's \`ReplicationConfigurationMetadata\`, calls \`ReplicationConfiguration.resolveBackends\` for prefix matching / priority dedup / per-backend destination+role stamping, then calls \`resolveSourceRole\` for the top-level role

**\`package.json\`** — arsenal pinned to \`improvement/ARSN-571/crr-multi\` for the new static methods

**\`tests/utils/crr.js\`** — V2 bucket replication config fixture

**\`tests/unit/CRR/ReplicationStatusUpdater.js\`** — 7 new V2 test cases (single match, dual-rule prefix overlap, no-match skip, SITE_NAME filter, all-up-to-date skip, PROCESSING aggregate status, forceUsingConfiguration)

## Test plan

- [x] All 456 unit tests pass (\`yarn test:unit\`)
- [x] Lint clean (\`yarn lint\`)
- [x] V2 metadata written with per-backend \`destination\`/\`role\` and no legacy top-level fields
- [x] V1 configs produce identical output to before this change

Issue: S3UTILS-230

⚠️ The procedure at this page seems not up to date : https://scality.atlassian.net/wiki/spaces/TS/pages/2312536220/Ingestion+Replication+XDM#Step-one---Clear-the-replication-queue. The ENDPOINT should now be http://.... with the port `:80`. Should we ping CS or update the page directly ?

### DIFF
--- a/CRR/ReplicationStatusUpdater.js
+++ b/CRR/ReplicationStatusUpdater.js
@@ -441,7 +441,7 @@ class ReplicationStatusUpdater {
                     const hasMatchingRule = Rules.some(r =>
                         r.Status === 'Enabled' &&
                         (!this.siteName || r.Destination.StorageClass === this.siteName) &&
-                        Key.startsWith(r.Prefix || r.Filter?.Prefix || '')
+                        Key.startsWith(r.Filter?.Prefix ?? r.Prefix ?? '')
                     );
                     if (!hasMatchingRule) {
                         ++this._nSkipped;
--- a/CRR/ReplicationStatusUpdater.js
+++ b/CRR/ReplicationStatusUpdater.js
@@ -153,44 +153,6 @@ class ReplicationStatusUpdater {
     }
 
     /**
-     * Removes V1-only top-level fields from replicationInfo that have no meaning in V2 format.
-     * @private
-     * @param {Object} replicationInfo - The replication info object to mutate.
-     * @returns {void}
-     */
-    _removeV1Fields(replicationInfo) {
-        // eslint-disable-next-line no-param-reassign
-        delete replicationInfo.destination;
-        // eslint-disable-next-line no-param-reassign
-        delete replicationInfo.storageClass;
-        // eslint-disable-next-line no-param-reassign
-        delete replicationInfo.storageType;
-        // eslint-disable-next-line no-param-reassign
-        delete replicationInfo.dataStoreVersionId;
-    }
-
-    /**
-     * Initializes V2 replication info on objMD if it is missing or has no status.
-     * @private
-     * @param {ObjectMD} objMD - The object metadata.
-     * @param {string} sourceRole - The source-side IAM role ARN.
-     * @returns {void}
-     */
-    _initV2ReplicationInfo(objMD, sourceRole) {
-        const replicationInfo = objMD.getReplicationInfo();
-        if (!replicationInfo || !replicationInfo.status) {
-            const ops = objMD.getContentLength() === 0 ? ['METADATA'] : ['METADATA', 'DATA'];
-            objMD.setReplicationInfo({
-                status: 'PENDING',
-                role: sourceRole,
-                backends: [],
-                content: ops,
-                isNFS: null,
-            });
-        }
-    }
-
-    /**
      * Computes the aggregate top-level replication status from all backends per the V2 rules:
      *   any FAILED → FAILED; else any PENDING → PROCESSING; else COMPLETED.
      * @private
@@ -211,41 +173,26 @@ class ReplicationStatusUpdater {
      * @returns {boolean} True if the object should be updated.
      */
     _objectShouldBeUpdated(objMD, site) {
-        return this.replicationStatusToProcess.some(filter => {
-            if (filter === 'NEW') {
-                // Either site specific replication info is missing
-                // or are initialized with empty fields.
-                return (!objMD.getReplicationInfo()
-                    || !objMD.getReplicationSiteStatus({ site }));
-            }
-            return (objMD.getReplicationInfo()
-                && objMD.getReplicationSiteStatus({ site }) === filter);
-        });
+        const info = objMD.getReplicationInfo();
+        const siteStatus = info && objMD.getReplicationSiteStatus({ site });
+        return this.replicationStatusToProcess.some(filter =>
+            filter === 'NEW' ? (!info || !siteStatus) : (info && siteStatus === filter));
     }
 
     /**
-     * Marks an object as pending for replication.
+     * Fetches, validates, mutates, and writes back object metadata.
      * @private
-     * @param {string} bucket - The bucket name.
-     * @param {string} key - The object key.
-     * @param {string} versionId - The object version ID.
-     * @param {string} storageClass - The storage class for replication.
-     * @param {Object} repConfig - The replication configuration.
-     * @param {Function} cb - Callback function.
+     * @param {string} bucket
+     * @param {string} key
+     * @param {string} versionId
+     * @param {Function} mutate - Receives objMD; return { skip: true } to skip the write.
+     * @param {Function} cb
      * @returns {void}
      */
-    _markObjectPending(
-        bucket,
-        key,
-        versionId,
-        storageClass,
-        repConfig,
-        cb,
-    ) {
+    _updateObjectMD(bucket, key, versionId, mutate, cb) {
         let objMD;
         let skip = false;
         return waterfall([
-            // get object blob
             next => this.cloudserverclient.getMetadata({
                 Bucket: bucket,
                 Key: key,
@@ -257,77 +204,19 @@ class ReplicationStatusUpdater {
                 objMD = new ObjectMD(originalMD);
                 const newMDVersion = objMD.getModelVersion();
 
-                // Prevent schema downgrade: do not write metadata if this model version
-                // is older than the object's original version, to avoid losing newer fields.
                 if (newMDVersion < originalMDVersion) {
                     this.log.error('model version regression: newMDVersion < originalMDVersion', {
-                        bucket,
-                        key,
-                        versionId,
-                        newMDVersion,
-                        originalMDVersion,
+                        bucket, key, versionId, newMDVersion, originalMDVersion,
                     });
                     return next(new Error('model version regression: refusing to overwrite newer metadata'));
                 }
 
-                if (!this._objectShouldBeUpdated(objMD, storageClass)) {
+                const result = mutate(objMD);
+                if (result?.skip) {
                     skip = true;
                     return process.nextTick(next);
                 }
-                // Initialize replication info, if missing
-                // This is particularly important if the object was created before
-                // enabling replication on the bucket.
-                let replicationInfo = objMD.getReplicationInfo();
-                const { Rules, Role } = repConfig;
-                const destination = Rules[0].Destination.Bucket;
-                
-                if (!replicationInfo || !replicationInfo.status) {
-                    // set replication properties
-                    const ops = objMD.getContentLength() === 0 ? ['METADATA']
-                        : ['METADATA', 'DATA'];
-                    replicationInfo = {
-                        status: 'PENDING',
-                        backends: [],
-                        content: ops,
-                        destination,
-                        storageClass: '',
-                        role: Role,
-                        storageType: '',
-                        dataStoreVersionId: '',
-                    };
-                    objMD.setReplicationInfo(replicationInfo);
-                }
 
-                // Force reset object's replication configuration to match bucket's configuration
-                if (this.forceUsingConfiguration) {
-                    objMD.getReplicationInfo().destination = destination;
-                    objMD.setReplicationRoles(Role);
-                }
-                // Update replication info with site specific info
-                if (!objMD.getReplicationSiteStatus({ site: storageClass })) {
-                    // When replicating to multiple destinations,
-                    // the storageClass and storageType properties
-                    // become comma-separated lists of the storage
-                    // classes and types of the replication destinations.
-                    const ri = objMD.getReplicationInfo();
-                    ri.storageClass = ri.storageClass
-                        ? `${ri.storageClass},${storageClass}` : storageClass;
-                    if (this.storageType) {
-                        ri.storageType = ri.storageType
-                            ? `${ri.storageType},${this.storageType}` : this.storageType;
-                    }
-                    // Add site to the list of replication backends
-                    const backends = objMD.getReplicationBackends();
-                    backends.push({
-                        site: storageClass,
-                        status: 'PENDING',
-                        dataStoreVersionId: '',
-                    });
-                    objMD.setReplicationBackends(backends);
-                }
-
-                objMD.setReplicationSiteStatus({ site: storageClass }, 'PENDING');
-                objMD.setReplicationStatus('PENDING');
                 objMD.updateMicroVersionId();
                 const md = objMD.getSerialized();
                 return this.cloudserverclient.putMetadata({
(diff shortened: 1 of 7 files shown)


CRR/ReplicationStatusUpdater.js
```diff
@@ -117,6 +117,78 @@ class ReplicationStatusUpdater {
...
 
 
+    /**
+     * Returns true if the replication config uses the legacy V1 format (any rule has a
+     * comma-separated StorageClass, e.g. "dest-A,dest-B").
+     * @private
+     * @param {Object} repConfig - The replication configuration from GetBucketReplicationCommand.
+     * @returns {boolean}
+     */
+    _isLegacyFormat(repConfig) {
```
DarkIsDude: Let's use v1 / v2 format name instead of legacy. Legacy don't mean anything

CRR/ReplicationStatusUpdater.js
```diff
@@ -312,24 +502,48 @@ class ReplicationStatusUpdater {
...
-                    const errMsg = 'missing SITE_NAME environment variable, must be set to'
-                        + ' the value of "site" property in the CRR configuration';
-                    this.log.error(errMsg);
-                    return next(new Error(errMsg));
-                }
-                if (!this.siteName) {
-                    this.log.warn(`missing SITE_NAME environment variable, triggering replication to the ${storageClass} storage class`);
+
+                if (this._isLegacyFormat(repConfig)) {
+                    // V1: single storage class (possibly comma-separated, handled by _markObjectPending)
```
DarkIsDude:
```suggestion
```

CRR/ReplicationStatusUpdater.js
```diff
@@ -312,24 +502,48 @@ class ReplicationStatusUpdater {
...
+                        if (this.currentVersionOnly && !IsLatest) {
+                            ++this._nSkipped;
+                            apply();
+                            return;
+                        }
+                        this._markObjectPending(bucket, Key, VersionId, storageClass, repConfig, apply);
+                    }, next);
                 }
+
+                // V2: one rule per destination, prefix-matched per object
```
DarkIsDude:
```suggestion
```

CRR/ReplicationStatusUpdater.js
```diff
@@ -270,6 +342,124 @@ class ReplicationStatusUpdater {
...
+                    objMD.setReplicationInfo({
+                        status: 'PENDING',
+                        role: sourceRole,
+                        backends: [],
+                        content: ops,
+                        isNFS: null,
+                    });
+                    replicationInfo = objMD.getReplicationInfo();
+                }
+                // Remove legacy top-level fields that don't belong in V2 format
```
DarkIsDude: We can create a dedicated function and associated tests to avoid the comment and make the code clearer

CRR/ReplicationStatusUpdater.js
```diff
@@ -270,6 +342,124 @@ class ReplicationStatusUpdater {
...
+        let objMD;
+        let skip = false;
+        return waterfall([
+            next => this.cloudserverclient.getMetadata({
+                Bucket: bucket,
+                Key: key,
+                VersionId: versionId,
+            }, next),
+            (mdRes, next) => {
+                const originalMD = JSON.parse(mdRes.Body);
```
DarkIsDude: This function is very long and deserve several small function 🙏

CRR/ReplicationStatusUpdater.js
```diff
@@ -117,6 +117,142 @@ class ReplicationStatusUpdater {
...
+
+    /**
+     * Derives the per-rule destination role ARN by substituting the account ID.
+     * If baseRole is "src,dest" format, uses the dest part as the template.
+     * @private
+     * @param {string} baseRole - The top-level Role from the replication config.
+     * @param {string} [account] - The account ID from Destination.Account.
+     * @returns {string}
+     */
+    _deriveRuleRole(baseRole, account) {
```
maeldonn: _deriveRuleRole` duplicates `ReplicationConfiguration.resolveDestinationRole` from arsenal (`lib/models/ReplicationConfiguration.ts:228`). Same semantics, but arsenal splits the ARN on `:` instead of using a regex. Reuse it instead.

CRR/ReplicationStatusUpdater.js
```diff
@@ -270,6 +406,93 @@ class ReplicationStatusUpdater {
...
+                const rulesToUpdate = matchingRules.filter(rule =>
+                    this._objectShouldBeUpdated(objMD, rule.Destination.StorageClass)
+                );
+
+                if (rulesToUpdate.length === 0) {
+                    skip = true;
+                    return process.nextTick(next);
+                }
+
+                const sourceRole = repConfig.Role.split(',')[0].trim();
```
maeldonn: `repConfig.Role.split(',')[0].trim()` is `ReplicationConfiguration.resolveSourceRole` from arsenal (`lib/models/ReplicationConfiguration.ts:247`). Reuse it.

CRR/ReplicationStatusUpdater.js
```diff
@@ -117,6 +117,142 @@ class ReplicationStatusUpdater {
...
+
+    /**
+     * Returns the subset of rules whose prefix matches the given key, deduplicated by site
+     * (highest Priority wins when two rules target the same StorageClass).
+     * @private
+     * @param {string} key - The object key.
+     * @param {Array} rules - The replication rules from GetBucketReplicationCommand.
+     * @returns {Array}
+     */
+    _getMatchingRules(key, rules) {
```
maeldonn: `_getMatchingRules` + `_applyRulesToBackends` (line 225) together duplicate `ReplicationConfiguration.resolveBackends` from arsenal (`lib/models/ReplicationConfiguration.ts:269`), which already handles enabled-rule filtering, prefix matching, comma-separated storageClass expansion, dedup by priority, per-backend `destination`/`role` stamping, and carrying `dataStoreVersionId` forward from existing backends. A small adapter is needed to map the AWS SDK response shape (`Rules[].Destination.{Bucket,StorageClass,Account}`) to arsenal's `ReplicationConfigurationMetadata` shape, then this whole block collapses to one call. Please reuse the arsenal helper instead of maintaining a parallel implementation.

CRR/ReplicationStatusUpdater.js
```diff
@@ -117,6 +117,142 @@ class ReplicationStatusUpdater {
...
 
 
+    /**
+     * Returns true if the replication config uses the V1 format (any rule has a
+     * comma-separated StorageClass, e.g. "dest-A,dest-B").
+     * @private
+     * @param {Object} repConfig - The replication configuration from GetBucketReplicationCommand.
+     * @returns {boolean}
+     */
+    _isV1Format(repConfig) {
```
maeldonn, line 127: `_isV1Format` heuristic can be aligned with arsenal's rule (`lib/models/ReplicationConfiguration.ts:508`): v2 if the rule carries a `Filter` element. Cleaner and removes the false-positive case where a single-rule v2 config without `Priority`/`Account` is misclassified as v1.

CRR/ReplicationStatusUpdater.js
```diff
@@ -270,6 +406,93 @@ class ReplicationStatusUpdater {
...
+                );
+
+                if (rulesToUpdate.length === 0) {
+                    skip = true;
+                    return process.nextTick(next);
+                }
+
+                const sourceRole = repConfig.Role.split(',')[0].trim();
+                const replicationInfo = this._initV2ReplicationInfo(objMD, sourceRole);
+                this._removeV1Fields(replicationInfo);
```
maeldonn: You delete fields from `replicationInfo`, but you never save it back with `setReplicationInfo`. If arsenal returns a copy instead of the real object, your deletes are lost. Use `objMD.setReplicationInfo(replicationInfo)` to avoid this.

maeldonn (changes requested):
Can be simplified a lot by using arsenal logic

CRR/ReplicationStatusUpdater.js
DarkIsDude: Fixed. `_deriveRuleRole` has been removed entirely. The new V2 path uses `ReplicationConfiguration.resolveBackends`, which internally calls `resolveDestinationRole` when stamping the per-backend role.

CRR/ReplicationStatusUpdater.js
DarkIsDude: Fixed. The inline `split(',')[0]` has been replaced with `ReplicationConfiguration.resolveSourceRole(repConfig.Role)`.

CRR/ReplicationStatusUpdater.js
DarkIsDude: Fixed. `_applyRulesToBackends` and `_deriveRuleRole` are gone. The V2 path now adapts the AWS SDK rule shape into `ReplicationConfigurationMetadata` and delegates to `ReplicationConfiguration.resolveBackends`. Arsenal's `improvement/ARSN-571/crr-multi` branch was pinned in `package.json` to get the static methods.

CRR/ReplicationStatusUpdater.js
DarkIsDude: Fixed. `_isV1Format` now simply checks `repConfig.Rules.every(r => !r.Filter)`, matching arsenal's canonical V2 detection.

CRR/ReplicationStatusUpdater.js
DarkIsDude: Fixed. `objMD.setReplicationInfo(replicationInfo)` is now called immediately after `_removeV1Fields`.

tests/utils/index.js
```diff
@@ -18,19 +18,19 @@ function generateKeyId(isMaster, key, versionId) {
 function generateMD(isMaster, objectKey, lastModified, storageClasses) {
     const keyId = generateKeyId(isMaster, objectKey, testVersion);
 
+    const objMD = new ObjectMD()
+        .setKey(objectKey)
+        .setLastModified(lastModified)
+        .setVersionId(testVersion);
+    objMD.setReplicationInfo({ storageClass: storageClasses });
```
delthas: Could we make this method (`setReplicationInfo`) builder-style (return `this`)? Feels like we're breaking a useful precedent here, forcing us to extract to a var just for this setter whereas we previously had a nice inline builder pattern.

delthas (approved):

tests/utils/index.js
DarkIsDude: setReplicationInfo already returns this in arsenal. Fixed 🙏

CRR/ReplicationStatusUpdater.js
```diff
@@ -117,6 +117,92 @@ class ReplicationStatusUpdater {
...
+            if (!key.startsWith(prefix)) { continue; }
+            const site = rule.Destination.StorageClass;
+            const existing = siteMap.get(site);
+            const priority = rule.Priority || 0;
+            if (!existing || priority > (existing.Priority || 0)) {
+                siteMap.set(site, rule);
+            }
+        }
+        return [...siteMap.values()];
+    }
```
maeldonn: `_getMatchingRules` duplicates `resolveBackends` (prefix matching + priority dedup). Its only remaining role is the pre-fetch skip in `_markPending`, which can be an inline `Rules.some(...)`. Drop the helper, drop the `matchingRules` arg to `_markObjectPendingV2`, and pass `repConfig.Rules` directly so `resolveBackends` is the single source of truth.

CRR/ReplicationStatusUpdater.js
```diff
@@ -117,6 +117,92 @@ class ReplicationStatusUpdater {
...
+    /**
+     * Returns true if the replication config uses the V1 format.
+     * V2 rules carry a Filter element; V1 rules do not.
+     * @private
+     * @param {Object} repConfig - The replication configuration from GetBucketReplicationCommand.
+     * @returns {boolean}
+     */
+    _isV1Format(repConfig) {
+        return repConfig.Rules.every(r => !r.Filter);
+    }
```
maeldonn, line 129: Used once, could be inlined to simplify code reading.

CRR/ReplicationStatusUpdater.js
```diff
@@ -117,6 +117,92 @@ class ReplicationStatusUpdater {
...
+     *   any FAILED → FAILED; else any PENDING → PROCESSING; else COMPLETED.
+     * @private
+     * @param {Array} backends - The backends array from replicationInfo.
+     * @returns {string}
+     */
+    _computeTopLevelStatus(backends) {
+        if (backends.some(b => b.status === 'FAILED')) { return 'FAILED'; }
+        if (backends.some(b => b.status === 'PENDING')) { return 'PROCESSING'; }
+        return 'COMPLETED';
+    }
```
maeldonn, line 142: Since this is only used once, it could be inlined. While extracting logic to an external function usually improves readability, creating small functions for single-use cases can actually make the code harder to follow

CRR/ReplicationStatusUpdater.js
```diff
@@ -270,6 +356,141 @@ class ReplicationStatusUpdater {
...
+                }
+
+                const sourceRole = ReplicationConfiguration.resolveSourceRole(repConfig.Role);
+                this._initV2ReplicationInfo(objMD, sourceRole);
+                const replicationInfo = objMD.getReplicationInfo();
+                this._removeV1Fields(replicationInfo);
+                if (this.forceUsingConfiguration) {
+                    replicationInfo.role = sourceRole;
+                }
+                objMD.setReplicationInfo(replicationInfo);
```
maeldonn: Collapse into one `setReplicationInfo` call with the full V2 shape. V1 keys are omitted by construction, so `_initV2ReplicationInfo` and `_removeV1Fields` both go away:

```js
const prev = objMD.getReplicationInfo();
objMD.setReplicationInfo({
    status: this._computeTopLevelStatus(finalBackends),
    role: this.forceUsingConfiguration ? sourceRole : (prev?.role ?? sourceRole),
    backends: finalBackends,
    content: prev?.content ?? (objMD.getContentLength() === 0 ? ['METADATA'] : ['METADATA', 'DATA']),
    isNFS: prev?.isNFS ?? null,
});
```

Drops `_initV2ReplicationInfo`, `_removeV1Fields`, the explicit `setReplicationStatus`, and the separate `setReplicationBackends` below.

CRR/ReplicationStatusUpdater.js
```diff
@@ -270,6 +356,141 @@ class ReplicationStatusUpdater {
...
+                            return b;
+                        }
+
+                        return {
+                            ...b,
+                            status: orig.status,
+                            dataStoreVersionId: orig.dataStoreVersionId ?? b.dataStoreVersionId,
+                        };
+                    });
+                const finalBackends = [...skippedBackends, ...backendsToUpdate];
```
maeldonn: One map replaces the filter+map+spread merge if `resolveBackends` is the source of shape:

```js
const finalBackends = candidateBackends.map(c => {
    if (updatedSites.has(c.site)) return c;
    const orig = existingBackends?.find(e => e.site === c.site);
    return orig ? { ...c, status: orig.status } : c;
});
```

This also drops sites that were in `existingBackends` but no longer in the config, which is the V2-correct behavior.

CRR/ReplicationStatusUpdater.js
```diff
@@ -270,6 +356,141 @@ class ReplicationStatusUpdater {
...
+                return;
+            }
+            if (skip) {
+                ++this._nSkipped;
+            } else {
+                ++this._nUpdated;
+            }
+            cb();
+        });
+    }
```
maeldonn: `_markObjectPending` and `_markObjectPendingV2` share ~70% scaffold: `getMetadata`, parse, version-regression check, `putMetadata`, counters. Factor it into `_updateObjectMD(bucket, key, versionId, mutate, cb)` where `mutate(objMD)` returns `{ skip: true }` or void.

CRR/ReplicationStatusUpdater.js
```diff
@@ -270,6 +356,141 @@ class ReplicationStatusUpdater {
...
+                const newMDVersion = objMD.getModelVersion();
+
+                if (newMDVersion < originalMDVersion) {
+                    this.log.error('model version regression: newMDVersion < originalMDVersion', {
+                        bucket, key, versionId, newMDVersion, originalMDVersion,
+                    });
+                    return next(new Error('model version regression: refusing to overwrite newer metadata'));
+                }
+
+                const arsenalConfig = this._buildArsenalConfig(matchingRules, repConfig);
```
maeldonn: `_buildArsenalConfig` runs per-object, but its output is constant per bucket

CRR/ReplicationStatusUpdater.js
```diff
@@ -130,10 +216,10 @@ class ReplicationStatusUpdater {
...
                 // or are initialized with empty fields.
                 return (!objMD.getReplicationInfo()
-                    || !objMD.getReplicationSiteStatus(site));
+                    || !objMD.getReplicationSiteStatus({ site }));
             }
             return (objMD.getReplicationInfo()
-                && objMD.getReplicationSiteStatus(site) === filter);
+                && objMD.getReplicationSiteStatus({ site }) === filter);
         });
     }
```
maeldonn: `getReplicationInfo()` and `getReplicationSiteStatus({ site })` are called twice per filter iteration. Hoist them once before the `some()`:

```js
const info = objMD.getReplicationInfo();
const siteStatus = info && objMD.getReplicationSiteStatus({ site });
return this.replicationStatusToProcess.some(filter =>
    filter === 'NEW' ? (!info || !siteStatus) : (info && siteStatus === filter));
```

maeldonn (changes requested):
One final batch of reviews and we should be good to go 💪

CRR/ReplicationStatusUpdater.js
DarkIsDude: It was intended here. As the function is waterflow, with old style and very long, I specifically tried to split it. If it's a hard no for you, I can change 🙏

maeldonn (approved):
LGTM

package.json
```diff
@@ -33,7 +33,7 @@
     "@senx/warp10": "^2.0.3",
     "@smithy/util-retry": "^4.0.7",
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/arsenal#8.2.36",
+    "arsenal": "git+https://github.com/scality/arsenal#8.4.7",
```
maeldonn: Please update to arsenal 8.5.0

maeldonn (changes requested):

package.json
DarkIsDude: Of course boss !

maeldonn (approved):

delthas (approved):

delthas:
Sorry for the delay :) Good work

package.json
```diff
@@ -50,8 +50,10 @@
     "xml2js": "^0.6.2"
   },
   "resolutions": {
+    "@opentelemetry/core": "^2.8.0",
```
delthas: Wait, why is this needed? We try to avoid those, no? As in https://github.com/scality/cloudserver/pull/6207/changes#r3475681342

DarkIsDude: 👏 indeed. It was just before the fix on arsenal, I'll just remove it 🙏

PR #399, opened 2026-06-16
🔧 migrate code review to copilot-instructions flow

DarkIsDude:
## Summary

- Replace `.claude/skills/review-pr/SKILL.md` with `.github/copilot-instructions.md` following the new copilot review flow
- Update `review.yml` to use `scality/workflows@v2.8.3` and add `labeled`/`unlabeled` PR trigger types

## Test plan

- [ ] Verify the Code Review workflow triggers on PR open/synchronize/label events
- [ ] Confirm the criteria table in `copilot-instructions.md` is used by the automated review

Issue: S3UTILS-240


.github/workflows/review.yml
```diff
@@ -2,11 +2,11 @@ name: Code Review
 
 on:
   pull_request:
-    types: [opened, synchronize]
+    types: [opened, synchronize, labeled, unlabeled]
 
 jobs:
   review:
-    uses: scality/workflows/.github/workflows/claude-code-review.yml@v2
+    uses: scality/workflows/.github/workflows/claude-code-review.yml@v2.8.3
```
francoisferrand: best to stick with @v2 : esp. this is an _internal_ (scality) workflow, so controlling update is not really needed?

.github/workflows/review.yml
```diff
@@ -2,11 +2,11 @@ name: Code Review
 
 on:
   pull_request:
-    types: [opened, synchronize]
+    types: [opened, synchronize, labeled, unlabeled]
```
francoisferrand: is labeled/unlabeled needed? we don't have it elsewhere AFAIK, and not used by the workflow you import...

.github/workflows/review.yml
francoisferrand, line 1: should be updated to match the "default" template? (inherit secrets)

.github/workflows/review.yml
DarkIsDude: That's always a so-so in that case. I prefer to pin to a specific version, even if it's internal. In case something goes wrong, it's not everywhere... But the stay up to date is harder. I don't think we have a right answer here 😬 but you have my opinion

francoisferrand: not sure we have a "right" answer, but we have a same pattern everywhere (floating tags like `@v2`) : so I'd rather keep consistency, even if not the "best" approach.

And we can have the discussion if there is a better way: and from there do it consistently everywhere...

francoisferrand (approved):

.github/workflows/review.yml
SylvainSenechal: I'm fine with pinned version

SylvainSenechal (approved):

.github/workflows/review.yml
DarkIsDude: I checked what is done. No "same pattern everywhere" but the floating tags is more common. I don't have a strong opinion as I said, so let's go

PR #402, opened 2026-06-24
Fix count items stuck on empty bucket list

SylvainSenechal:
ISSUE: S3UTILS-243

Fix 1 bug and 1 risk :

- Count item never finishes when the artesca has 0 buckets 
- While doing local tests, I found the scripts was crashing on non existent replicationInfo metadata : added defensive code

### DIFF
--- a/CountItems/CountManager.js
+++ b/CountItems/CountManager.js
@@ -214,6 +214,10 @@ class CountManager {
             }
         });
         this.q.resume();
+        if (this.q.idle()) {
+            this.log.info('no buckets to process, skipping count');
+            process.nextTick(onceCB);
+        }
     }
 }
 
--- a/utils/S3UtilsMongoClient.js
+++ b/utils/S3UtilsMongoClient.js
@@ -280,7 +280,7 @@ class S3UtilsMongoClient extends MongoClientInterface {
                     targetCount = 'versionCount';
                     targetData = 'versionData';
 
-                    if (entry.value.replicationInfo.backends.length > 0
+                    if (entry.value.replicationInfo?.backends?.length > 0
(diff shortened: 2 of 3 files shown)


CountItems/CountManager.js
SylvainSenechal, line 1: No version bumps, we said we will release it with 4.3.0 and we have other ongoing prs opened

francoisferrand: > we have other ongoing prs opened

other PRs target 1.18, so will not help deliver this in 4.3.0 : best to make the release already for 1.16 and 1.17 ?

utils/S3UtilsMongoClient.js
```diff
@@ -523,7 +523,7 @@ class S3UtilsMongoClient extends MongoClientInterface {
                 results.location[bucketInfo.getLocationConstraint()] = size;
             }
         }
-        entry.value.replicationInfo.backends.forEach(rep => {
+        (entry.value.replicationInfo?.backends ?? []).forEach(rep => {
```
```suggestion
        entry.value.replicationInfo?.backends?.forEach(rep => {
```

francoisferrand (approved):

CountItems/CountManager.js
```diff
@@ -214,6 +214,10 @@ class CountManager {
             }
         });
         this.q.resume();
+        if (this.q.length() === 0 && this.q.idle()) {
```
maeldonn: Isn't `this.q.length() === 0` redundant with `this.q.idle()` ?

maeldonn (approved):

CountItems/CountManager.js
SylvainSenechal: yes it is

